PR #10, opened 2020-01-27
Fix up some URLs

tsibley:
Two changes:

1. Switch back to data.nextstrain.org now that it supports HTTPS.
    Not an urgent change, but loading will be faster (and cheaper) from CloudFront than S3.

2. Fix mistaken references to localhost:4000


trvrb:
Thanks Tom!

rneher:
excellent! thanks!

PR #19, opened 2020-01-29
some typo fixes and rephrasing

tolot27:
Will address parts of #20.


tolot27:
>L92: Would keep "Sterblichkeitsrate"

Fixed L44 for consistancy, too.

> L96: Difference between Swiss and German writing, i.e. not really an error. Since everything else in the narrative is using the CH writing, I would keep that consistent.

How about quotation marks and the remaining Eszett (ß)? I'm not very familar with CH writing and was not aware of all the differences. I've reverted most of them.

Maybe the file and branch should be named de_CH?

rneher:
Thank you @tolot27 and @nicfel for catching all the mistakes. My first run yesterday was very rough. I don't think we need to be too nit-picky about ß vs ss. I am certainly more used to standard German, but us the ss mostly out of convenience in typing.

PR #34, opened 2020-01-31
fix(en-us): typo: "genomic analyis"

ivan-aksamentov:
Please doublecheck. I am not a native English speaker.


emmahodcroft:
Good catch! Thank you. Will change here and copy over to the where the narratives are served from :)

PR #35, opened 2020-01-31
fix(en-us): Remove mysterious tags in "Executive summary"

ivan-aksamentov:
Curiously, tags are attached to "Answer to the Ultimate Question of Life, the Universe, and Everything". Coincidence? I don't think so. 👽 🚀


rneher:
hehe. Emma added them to facilitate search-replace. We are planning to make them more specific in the future.

ivan-aksamentov:
I see. Should translations also preserve them?

P.S. I guess this issue can be closed then.

rneher:
yes, I'd preserve them. the number of sequences changes rapidly, so ideally one would template this in some way...

PR #37, opened 2020-02-01
Remove extra period in zh narrative

IanBoyanZhang:
Cosmetic change to [nextstrain.org](nextstrain.org) only. 
Screen Shot 2020-02-01 at 12 28 33 PM


emmahodcroft:
Thank you! I will update this here and copy across to where the files are served from and the main webpage :) It'll take a few minutes for the changes to the visible on the live server.

PR #48, opened 2020-02-12
Phylogeo

trvrb:
Include phylogeographic reconstruction of `division`. This requires running augur with https://github.com/nextstrain/augur/pull/431 + https://github.com/nextstrain/augur/pull/443.

We likely need to monitor this to make sure it's behaving as intended, but I think is a big step up.


trvrb:
The blocking PRs to Augur have been merged and Augur v6.3.0 released. Merging this now.

PR #55, opened 2020-02-21
Export AA mutation frequencies

trvrb:
Export for every gene from 2020.0 to present in 1 month intervals. Not attempting to do region weighting. Parameters taken directly from seasonal-flu parameters.


rneher:
Do we still want this?

trvrb:
This never got implemented, which is fine. Closing now.

PR #56, opened 2020-02-24
Replace environment.yml with requirements.txt

kairstenfay:
Other users had some problems installing the conda environment from the
`environment.yml` I had exported and added to this repo. Change instead
to the simpler, `requirements.txt` for dependency tracking via
`pip freeze`. Upate the README accordingly.


kairstenfay:
Sorry, this is wrong. For some reason my `pip` path maps to a Pipenv instead of the appropriate conda environment. I'll push a fix.

kairstenfay:
Ok, now the `requirements.txt` file is correct.

jameshadfield:
Thanks @kairstenfay -- the requirements.txt has 50 (!) dependencies -- I would have only expected `aws` and `augur` (and maybe `nextstrain-cli`)

tsibley:
Wrote this in Slack, but for posterity here:

> I would expect nextstrain/ncov to not have a requirements.txt or environment.yml, similar to our other pathogen builds.
> 
> We do document how to get a Nextstrain environment (either via conda or via nextstrain build) in the "Building" section: https://github.com/nextstrain/ncov#building

kairstenfay:
Closing in favor of #58

PR #59, opened 2020-02-25
Add normalize_gisaid_fasta.sh for issue #53

brianpardy:
Call as:

scripts/normalize_gisaid_fasta.sh data/gisaid_cov2020_sequences.fasta data/sequences.fasta 15000

Or use params.min_length when calling from Snakefile.  Defaults to 15000 if unspecified.

I hope this is proper etiquette to submit this way.

### DIFF
--- a/Snakefile
+++ b/Snakefile
@@ -420,3 +420,19 @@ rule clean:
         "auspice"
     shell:
         "rm -rfv {params}"
+
+rule gisaid:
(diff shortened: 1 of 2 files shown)


trvrb:
Hi @brianpardy. Thanks so much for this PR. I'm sorry it took so long to review. This works exactly as it should.

PR #61, opened 2020-02-26
Fixed typo Singapore/8/2020 had extra 's' on the end

nodrogluap:


trvrb:
Thanks for the catch!

PR #65, opened 2020-03-03
Add Clade Label Functionality 

emmahodcroft:
This adds an `augur clade` rule which attaches clade labels for 3 clades - one clade including the 2 Italian sequences that link together, one clade that includes the other Italian sequence & the sequences from Brazil, Germany, Mexico, etc, and one clade with the 2 WA sequences.

Pictured here:
image

By default (modifications to the `auspice_config` files) the branch_labels are turned off, so you don't see anything. But they can be used for URL views!!

I have tried to make these things that seemed like they'd be fairly robust to re-builds. However they should be checked. If we are mainly using them in narrative view then we should link to a dated run, and so this shouldn't be a problem with the narrative.

You can easily check a run by using the menu to change "Branch labels" from "none" to "clades" and then you'll see where they are (or where they're missing). 

If this can be pushed today and the JSON files updated on Nextstrain, this will work with the narrative currently being prepared.


emmahodcroft:
Sorry about git history on this. I tried rebasing, so that's a fun thing I probably should never do again 😆

PR #87, opened 2020-03-08
added Chinese translation

fillorkill:


emmahodcroft:
@fillorkill Can you please change the merge branch back to `sit-rep_zh` so we can merge there and do a live test before merging to master? Thanks!

PR #91, opened 2020-03-08
Add Japanese translation 0305

satotake:
I do not know how to provision `data/sequences.fasta` and `data/metadata.tsv`
So I could not verify this change by building your site locally

It is not a big problem as long as this PR is valid.
After merging this, I will translate previous situation reports into Japanese

Thanks for your awesome works.


emmahodcroft:
Thank you for your hard work! Could you redirect the PR to merge into the `sit-rep_ja` branch (just created)? We can then merge there and check it works with a live preview before we merge to master! See the issue I just created and (hopefully) tagged you in :)

PR #106, opened 2020-03-09
Treat gaps as real & don't infer ambiguous mutations

jameshadfield:
Holding off pushing this for the moment as it introduces a number of `X` mutations (as expected) in low coverage areas. This is going to become very important if we have important deletions. 

image

### DIFF
--- a/Snakefile
+++ b/Snakefile
@@ -70,8 +70,7 @@ rule filter:
 rule align:
     message:
         """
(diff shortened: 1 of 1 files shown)


trvrb:
Thanks @jameshadfield! I just rebased this into bab177d26f9dad7f2a6e1dec488952f039fefadc.

PR #117, opened 2020-03-09
Add a "deploy_to_staging" rule to be used by automatic builds

kairstenfay:
Uploads files to an S3 bucket for nextstrain.org and optionally sends a
Slack alert to a configured webhook.

The S3 bucket defaults to s3://nextstrain-staging, the live staging
source, but it can be overridden for testing using snakemake's --config
or --configfile options.

The optional Slack webhook can be provided using the same config
options.  By default no message is sent.

### DIFF
--- a/Snakefile
+++ b/Snakefile
@@ -3,6 +3,8 @@ def get_todays_date():
     date = datetime.today().strftime('%Y-%m-%d')
     return date
 
+configfile: "config/Snakefile.yaml"
+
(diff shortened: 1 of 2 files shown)


kairstenfay:
[Staging builds for nextstrain/ncov](https://trello.com/c/F73KJhRi/360-staging-builds-for-nextstrain-ncov)

PR #119, opened 2020-03-09
Minor changes to tmrca_estimate.py

terrycojones:
Not tested - just some small suggestions. Feel free to cherry pick, etc.


rneher:
the Tmrca script is no longer used and was only relevant in January. while these linting changes are useful, I'd rather keep it in the state it was used and don't add to the history.

PR #124, opened 2020-03-09
Minor changes to branching_process.py

terrycojones:
Not tested - just some small suggestions. Feel free to cherry pick, etc.


rneher:
this too is no longer in use and I don't think linting is necessary at this point.

PR #161, opened 2020-03-14
Update ncov_sit-rep_nl_2020-03-13.md

rvosa:
#148 this is a draft translation where all the body text is translated into Dutch. I'm unsure whether there is anything I need to do in terms of updating navigation (i.e. internal URLs) or whether that's somehow handled automatically when you switch languages.


cassiawag:
We'll handle internal urls. You should be able to preview it now; thanks!

PR #173, opened 2020-03-14
feat: add recency field to output json

rneher:
this determines the number of days between now and submission and adds this info in different discrete categories to the output. Should help highlighting new sequences. Should be superseeded at some point by auspice interpreting submission date or `date added`


trvrb:
Thanks @rneher! I just made a couple small additions, but this is a great feature. Merging now.

PR #210, opened 2020-03-18
make GISAID fasta header line fix case-insensitive

tolot27:
Some new sequences from Netherlands were exculded by the `augur filter` because their fasta ID is prefixed in a different case (`hCov-19` instead of `hCoV-19`), e.g.: `hCov-19/Netherlands/NoordBrabant/68/2020|EPI_ISL_415524|2020-03-12`.

This PR changes the sed command to be case-insensitive.

### DIFF
--- a/scripts/normalize_gisaid_fasta.sh
+++ b/scripts/normalize_gisaid_fasta.sh
@@ -18,14 +18,14 @@ fi
 
 echo "Normalizing GISAID file $GISAID_SARSCOV2_IN to $GISAID_SARSCOV2_OUT (min length $MIN_LENGTH)"
 
-# Remove leading 'BetaCoV' and 'BetaCov' from sequence names
+# Remove leading 'hCoV-19' from sequence names (case insensitive)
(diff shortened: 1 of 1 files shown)


emmahodcroft:
@kairstenfay @trs Could you check this out, alongside any impact this may have had for our internal builds? Thanks so much!

brianpardy:
Ah I missed this one before submitting my own PR to resolve this.  TB removed the case-insensitive /i I had placed in the original version of the normalize_gisaid_fasta.sh file, this is a GNU-ism that is not supported on all versions of sed so I believe he did so intentionally.  The alternate PR I submitted uses regex [vV] to accomplish the same thing.

tsibley:
@emmahodcroft This doesn't impact our internal ingest.

emmahodcroft:
Thank you for checking @tsibley !

tolot27:
I'm closing this in favor of #212, which does not impact your internal builds as well.

PR #211, opened 2020-03-18
make sequence and metadata input files configurable

tolot27:
Introduces configurable parameters for sequences.fasta and metadata.tsv. This allows custom input file names or input files outside the git repository. Example:

```
snakemake -p --config metadata=/san/ncov/mymetadata.tsv sequences=/san/ncov/mysequences.fasta
```

Advantage: `git pull` does not return conflicts because of uncommited changes.

### DIFF
--- a/Snakefile
+++ b/Snakefile
@@ -34,8 +34,8 @@ files = rules.files.params
 rule download:
     message: "Downloading metadata and fasta files from S3"
     output:
-        sequences = "data/sequences.fasta",
-        metadata = "data/metadata.tsv"
+        sequences = config["sequences"],
+        metadata = config["metadata"]
     shell:
         """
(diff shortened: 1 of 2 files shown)


Snakefile
```diff
@@ -34,8 +34,8 @@ files = rules.files.params
 rule download:
     message: "Downloading metadata and fasta files from S3"
     output:
-        sequences = "data/sequences.fasta",
-        metadata = "data/metadata.tsv"
+        sequences = config["sequences"],
+        metadata = config["metadata"]
     shell:
         """
         aws s3 cp s3://nextstrain-ncov-private/sequences.fasta data/
```
tsibley: The shell commands run here also need to be updated to use `{output.sequences:q}` and `{output.metadata:q}`.

PR #212, opened 2020-03-18
update for hCoV/nCoV capitalization in new seqs

brianpardy:
A recent batch of Netherlands sequences included alternate capitalization in the sequence name prefix ("hCov-19" instead of "hCoV-19") and sequences from Georgia used the prefix "nCoV-19".

This pull request changes the sed commands rewriting the sequence names to support these new sequences by making the 'v' comparison case-insensitive and accepting an 'n' in place of the 'h'.

### DIFF
--- a/scripts/normalize_gisaid_fasta.sh
+++ b/scripts/normalize_gisaid_fasta.sh
@@ -25,7 +25,7 @@ echo "Normalizing GISAID file $GISAID_SARSCOV2_IN to $GISAID_SARSCOV2_OUT (min l
 # Eliminate duplicate sequences (keep only the first seen)
 
 #cat $GISAID_SARSCOV2_IN |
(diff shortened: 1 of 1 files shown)


brianpardy:
The the second and third commits on this PR add changes to update the obsolete comments left in normalize_gisaid_fasta.sh referencing 'BetaCoV', and to update the minimum sequence length from 15000 to 25000 to match the Snakefile minimum length change introduced in https://github.com/nextstrain/ncov/commit/b1ec65fcb97a22ee28a6368ba292b54a585cbe50

tolot27:
From my point of view, this PR looks good and covers more cases then #210.

tolot27:
@tsibley @emmahodcroft This PR addresses the same issue I've addressed with #210 and can be merged, IMHO, since it does not impact your internal builds.

tsibley:
Thanks for the reminder. Merged!

PR #216, opened 2020-03-18
Adds file that's been edited by translation software

cassiawag:


tolot27:
There is something wrong with this commit. Many line breaks were removed which destroys the structure of the document. 

I also recommand to use some linter to get valid Markdown/CommonMark formated files.

cassiawag:
I'm aware. This was a simply good staging place to give an example to a new GitHub user.

PR #239, opened 2020-03-21
3/20 sitrep: remove reference to color of Canadian sequence

camillemarie:
Removed reference to the color of the Canadian sequence on slide 6. 

Colorings for travel history are currently not consistent across browsers (this is probably a bug). On my laptop, the sequence looks yellow (Firefox) or blue (chrome). So long as this is true, color shouldn't be used to identify sequences when coloring by division_exposure.

I switched to identifying the sequence by date, but feel free to change my wording, or fix this in another way! Just wanted to let you know. @sidneymbell  @cassiawag 

---
Different colorings by division_exposure in the narrative:

Yellow (in Firefox)
yellow_firefox

Blue (in Chrome)
blue_chrome

### DIFF
--- a/narratives/ncov_sit-rep_2020-03-20.md
+++ b/narratives/ncov_sit-rep_2020-03-20.md
@@ -129,7 +129,7 @@ So, when reading this narrative, keep in mind that the size of each circle on th
 
 Identifying where a given case was infected is important for understanding which areas are experiencing local transmission versus primarily travel-associated cases. Sequence data can help us differentiate between these scenarios.
 <br><br>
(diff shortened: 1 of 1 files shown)


cassiawag:
Thanks so much for picking up on this @camillemarie! We really appreciate it. I've update the text accordingly.

jameshadfield:
This looks like a browser-specific bug in auspice -- notice the legend ordering is reversed, so the colour (in this case, the travel history is "Europe") is different. This shouldn't be a problem for the traits where we supply a scale, but will affect any trait where auspice generates the colour scale.

PR #255, opened 2020-03-23
Annotate haplotype status and tip frequencies

huddlej:
Adds support for annotating nodes that have identical nucleotide or amino acid sequences to a given reference node and also calculating tip frequencies using the KDE method. The combination of these two features allows us to annotate a specific nucleotide haplotype of interest (USA/WA1/2020) and then inspect the frequencies of that haplotype through time.

Enabling tip frequencies in auspice also allows us to investigate changes in frequencies through time across different node (color-by) attributes.

### DIFF
--- a/Snakefile
+++ b/Snakefile
@@ -240,6 +240,22 @@ rule ancestral:
             --infer-ambiguous
         """
 
+rule haplotype_status:
+    message: "Annotating haplotype status relative to {params.reference_node_name}"
+    input:
+        nt_muts = rules.ancestral.output.node_data
+    output:
+        node_data = "results/haplotype_status.json"
+    params:
+        reference_node_name = "USA/WA1/2020"
+    shell:
+        """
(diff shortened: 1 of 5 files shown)


trvrb:
@huddlej --- Thanks so much for this PR. Hugely appreciated. I rebased onto current `master` for testing and then made a commit that keeps calculations, but removes display. I was worried that the "Frequencies" panel has significant capacity for confusion as frequencies are completely dependent on sampling. But still good to be able to do a simple swap of `auspice_config.json` to take a look.

One idea (cc @jameshadfield), would be to add a new option to `display_defaults` that could turn on/off display of different panels. We could have a default view that is just the current 3 (tree, map, entropy) and allow frequencies by clicking on the sidebar toggle to enable this advanced feature.

jameshadfield:
Agreed. This is now an issue at https://github.com/nextstrain/auspice/issues/998

PR #262, opened 2020-03-24
add alternate name COVID-19

WeileiZeng:
COVID-19 is the most popular name right now. Added it in the first introduction line in README.


trvrb:
Thanks @WeileiZeng! I've incorporated this change here: bda415b781a3abf0d8b6202d1c3243fa08062c69. Wanted to be clear that COVID-19 is the disease and SARS-CoV-2 / hCoV-19 is the virus.

PR #268, opened 2020-03-24
Align sequences in smaller partitions to speed up builds

huddlej:

Aligning all available sequences with a single mafft process uses too much
memory. This commit introduces a Python script and Snakemake data-dependent
execution logic to partition sequences into smaller sets, align each set
individually, and aggregate the alignments into a single file.

Note that this commit also explicitly sets the number of threads to use for
augur align using Snakemake's threads keyword [1]. With the splitting of
alignments into multiple partitions, it is now possible to parallelize
alignments with multiple jobs using Snakemake's `-j` flag. It is faster to run 2
alignment jobs in parallel with 2 threads than to run 1 job at a time with 4
threads (3 min vs. 6 min on my MacBook). It is also dangerous to run multiple
jobs in parallel with Snakemake and the 'auto' threads option for augur align
because each job will try to use all available threads. Snakemake's threads
keyword is smart enough to only use the minimum of cores requested by the user
or available on the system.

[1] https://snakemake.readthedocs.io/en/stable/snakefiles/rules.html#threads

 ### Testing

I've noticed some minor stochastic variation in alignment outputs for split alignments that are concatenated compared to the corresponding full alignments. I haven't been able to track down the nature of this variation, but so far it does not seem to substantially alter the resulting trees.

### DIFF
--- a/Snakefile
+++ b/Snakefile
@@ -79,6 +79,21 @@ rule filter:
             --output {output.sequences}
         """
 
+checkpoint partition_sequences:
+    input:
+        sequences = rules.filter.output.sequences
+    output:
+        split_sequences = directory("results/split_sequences")
+    params:
+        sequences_per_group = 150
+    shell:
+        """
+        python3 scripts/partition-sequences.py \
+            --sequences {input.sequences} \
+            --sequences-per-group {params.sequences_per_group} \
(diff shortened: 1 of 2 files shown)


trvrb:
Looks great! Tested and working for most recent data. Thanks John.

PR #270, opened 2020-03-25
Build the tree with multiple threads

huddlej:

To speed up ncov builds a little more, this uses Snakemake's built-in threads keyword to specify a default number of threads to use for tree building. This commit specifically sets number of threads to 4 instead of using augur tree's default of 1 thread.

Snakemake will scale down the number of threads used to match the value of `--cores` or `--jobs`. The user can override the rule's thread setting by passing `--set-threads tree=8`, for example.

### DIFF
--- a/Snakefile
+++ b/Snakefile
@@ -165,11 +165,13 @@ rule tree:
         alignment = rules.mask.output.alignment
     output:
         tree = "results/tree_raw.nwk"
+    threads: 4
     shell:
         """
         augur tree \
             --alignment {input.alignment} \
-            --output {output.tree}
+            --output {output.tree} \
+            --nthreads {threads}
         """
 
 rule refine:


trvrb:
Thanks John! I missed this. Shouldn't this be set to `--nthreads auto`? And then it will scale to larger AWS instances if necessary?

huddlej:
That makes sense. I don't know how the AWS jobs are being executed right now, so I wrote this more to work on people's personal computers.

I've been trying to avoid `auto` locally because it will try to use all available threads in that single job even if I've run Snakemake with multiple jobs. In cases like the multiple augur align jobs, the auto option ends up overscheduling the available cores.

We can also increase the number of threads here in Snakemake to the number of available cores on AWS and Snakemake will automatically scale down that number of personal computers with fewer cores.

trvrb:
Let's see if I'm understanding. We should set `augur align` to use `nthreads 2` so that multiple alignment jobs can run simultaneously, but there should be just one `augur tree` that runs for the moment. Although I realize that should change shortly as we do region-specific trees. I'd agree with explicitly setting threads for `align` and `tree`. A larger AWS job should be able to execute with more concurrency, while a local job will execute in a more serial fashion.

I can merge this as written.

PR #271, opened 2020-03-25
Explicitly pass a string path to SeqIO.write when partitioning sequences for alignment

huddlej:
This fixes a bug on some systems (and potentially versions of BioPython) that do not accept a PosixPath as a valid input. Here we convert the PosixPath object to a string. Since this approach is currently used by augur align's write_seqs, for example, it should work here in a more cross-platform manner.


huddlej:
Merging this to allow others to run builds on non-OS X systems.

PR #274, opened 2020-03-27
harmonize python command to execute scripts

tolot27:
This PR changes all commands to execute python scripts to `python`. Before, `python` was used six times and `python3` eight times. Some environments (i.e. Anaconda in Windows with Python 3.8) only install the `python` command and do not create soft-links `to python3`, if it is not a side-by-side installation. Hence, snakemake cannot process the snakefile.

I know that this is maybe incompatible with side-by-side installations were the alternatives are configured in such a way that `python` is linked to `python2`. Nevertheless, Python 2 is EOL and this should not be a real problem, IMHO.

### DIFF
--- a/Snakefile
+++ b/Snakefile
@@ -90,7 +90,7 @@ checkpoint partition_sequences:
         sequences_per_group = 150
     shell:
         """
-        python3 scripts/partition-sequences.py \
+        python scripts/partition-sequences.py \
             --sequences {input.sequences} \
             --sequences-per-group {params.sequences_per_group} \
             --output-dir {output.split_sequences}
@@ -153,7 +153,7 @@ rule mask:
         mask_sites = "18529 29849 29851 29853"
     shell:
         """
-        python3 scripts/mask-alignment.py \
+        python scripts/mask-alignment.py \
             --alignment {input.alignment} \
             --mask-from-beginning {params.mask_from_beginning} \
             --mask-from-end {params.mask_from_end} \
@@ -252,7 +252,7 @@ rule haplotype_status:
         reference_node_name = "USA/WA1/2020"
     shell:
         """
-        python3 scripts/annotate-haplotype-status.py \
+        python scripts/annotate-haplotype-status.py \
             --ancestral-sequences {input.nt_muts} \
             --reference-node-name {params.reference_node_name:q} \
             --output {output.node_data}
@@ -328,7 +328,7 @@ rule colors:
         colors = "config/colors.tsv"
     shell:
         """
-        python3 scripts/assign-colors.py \
+        python scripts/assign-colors.py \
             --ordering {input.ordering} \
             --color-schemes {input.color_schemes} \
             --output {output.colors}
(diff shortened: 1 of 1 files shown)


tsibley:
> Some environments (i.e. Anaconda in Windows with Python 3.8) only install the python command and do not create soft-links to python3, if it is not a side-by-side installation. Hence, snakemake cannot process the snakefile.

I agree this is a problem, and we should consider how to fix it. Thanks for bringing it to our attention!

> I know that this is maybe incompatible with side-by-side installations were the alternatives are configured in such a way that `python` is linked to `python2`. Nevertheless, Python 2 is EOL and this should not be a real problem, IMHO.

But I strongly disagree that this isn't also a real problem. Most Linux environments still link `python` to Python 2. This includes our [`nextstrain/base` Docker image](https://github.com/nextstrain/docker-base).

In the Nextstrain CLI, I wrote this bit of code to figure out the correct Python executable:

https://github.com/nextstrain/cli/blob/015ab9979ef4d38b7ce17723e706d01f0b29b690/nextstrain/cli/util.py#L59-L67

but it might be a bit much to include something like that in the Snakefile.

tolot27:
I simply suggest to be consistent in calling python scripts from `Snakefile`: either `python` or `python3`. One can configure the default behaviour using `sudo update-alternatives --config python`.

tolot27:
@tsibley Would you agree on switching to `python3`? As far as I know, `python3` is available on every python 3+ installation, but `python` isn't available on never linux distributions like Ubuntu Focal, unless it is configured using alternatives.

tsibley:
Standardizing on `python3` seems good to me and is what I would have done, except for the problem you noted originally:

> Some environments (i.e. Anaconda in Windows with Python 3.8) only install the `python` command and do not create soft-links to `python3`, if it is not a side-by-side installation.

which seems to contradict:

> As far as I know, python3 is available on every python 3+ installation

Standardizing on `python3` may still be the best option, though.

tsibley:
Another option is to use a simplified version of the code I linked to earlier which [determines the currently running Python executable](https://github.com/nextstrain/cli/blob/015ab9979ef4d38b7ce17723e706d01f0b29b690/nextstrain/cli/util.py#L59-L67). For example, the top of the Snakefile could contain:

```
python = sys.executable or "python3"
```

and then rules needing to call the interpreter would use:

```
rule:
    shell:
        """
        {python:q} …
        """
```

rneher:
@tsibley the reorganization of the workflow makes this obsolete. But we still have the same inconsistency in https://github.com/nextstrain/ncov/blob/master/rules/builds.smk

Your proposal seems good to me.

tsibley:
@rneher Nod. One potential issue with my most recent proposal is if Snakemake is running locally but submitting jobs to remote machines in a cluster. Depending on how Snakemake submits those jobs—does it submit them as `snakemake` invocations or just the shell snippet?—the remote job might end up with the local `sys.executable` path, which may not be correct on the remote machines. Since you have more experience running this on a cluster, thoughts of if this will be an issue? /cc @huddlej

tsibley:
@rneher #412 is a PR implementing my proposal above. It should get some more testing before merge, which I'd appreciate your help with.

rneher:
this has been done in  different PRs by now

PR #323, opened 2020-03-31
Explain how to use GISAID data

### LINKED ISSUE #53: GISAID all-sequences fasta should be directly usable by nextstrain/ncov
Filing this as an issue as suggested by @emmahodcroft:

GISAID provides an all-sequences download button for SARS-CoV-2 sequences.  The provided file is not directly usable as a sequences.fasta file in nextstrain/ncov because of several issues in the GISAID file:

1. There is at least one duplicate sequence name (Italy/INMI1/2020) that causes errors in augur filter, and often other duplicates exist before they are renamed on GISAID
2. There are several sequences with "Hong Kong" in their names that cause errors in augur filter due to the sequence name being truncated at whitespace
3. The sequence names are appended with the EPI_ISL identifier and a datestamp, which are not stripped when loaded and cause mismatches with sequence names in metadata.tsv
4. The sequence names are prepended with 'BetaCoV' or 'BetaCov' which is not stripped when loaded and causes mismatches with metadata.tsv

I suggest a new bash script in ncov/scripts/ that would optionally normalize the GISAID all-sequences download file so that users can use it directly without a need to manually remove duplicates or edit sequence names or maintain their own automated pipeline to generate data/sequences.fasta.

I also suggest automating this in the Snakefile if possible, but I'm not sure how.  If no data/sequences.fasta file exists, but data/gisaid_cov2020_sequences.fasta exists, run scripts/normalize-gisaid-fasta.sh before the rest of the pipeline.

I can likely have a base script written to do the normalization by this evening if there is interest.

Thank you for considering this idea.

oneillkza:
Add detailed steps for how to obtain and normalize GISAID fasta, as resolved in #53 and #59 .

It took me quite a while to find #59, and I was halfway through rolling my own solution. Having it mentioned in the docs will help other users to more quickly find out how to do this.

### Related issue(s)  

Fixes #53 
Related to #59

### DIFF
--- a/DEV_DOCS.md
+++ b/DEV_DOCS.md
@@ -9,6 +9,18 @@ Please see [the main Nextstrain docs](https://nextstrain.org/docs/getting-starte
 In order to run the Nextstrain build you must provision `./data/sequences.fasta` and `./data/metadata.tsv`.
 We've included a test set of sequences that are publicly available via Genbank as `./example_data/sequences.fasta`.
 
+The public version of this NextStrain build uses data from GISAID. GISAID data is publically available but under restrictive licensing terms that prevent it from being hosted here. However, you can download the fasta yourself, and use it with a little normalization. The `./data/metadata.tsv` distributed in this repository has already been provisioned to match the GISAID data. 
+
+To get the fasta, follow these steps:
+
+#1 [Register for a GISAID account](https://www.gisaid.org/registration/register/) if you do not have one.
+#2 Log into GISAID and click the EpiCov(tm) tab.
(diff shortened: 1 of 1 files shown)


trvrb:
Thanks for this PR @oneillkza! I'm sorry it never got merged. We've since made `metadata.tsv` and `sequences.fasta` available directly from GISAID.org as described here: https://docs.nextstrain.org/en/latest/tutorials/SARS-CoV-2/steps/data-prep.html#background-contextual-sequences. Closing this now.

oneillkza:
No sweat, @trvrb . Glad you came to a good solution!

PR #324, opened 2020-04-01
Focus filter sparse

gtonkinhill:
- Adapts the `priorities.py` script to take advantage of scipy sparse matrix libraries as used in pairsnp to calculate pairwise SNP distances.

- It produces slightly different results as there is no weight for the fraction masked (`fraction_masked_in_focal`) as in the previous version

### Testing


emmahodcroft:
Thank you so much @gtonkinhill ! Giving this a go now!

PR #337, opened 2020-04-06
Include geographic build targets

trvrb:
We have lots of sequences from all over the globe now. Even with modest subsampling the "live" site is up to 3400 genomes. This is stretching the performance of the Auspice front end and makes it difficult to discern country-specific or division-specific patterns.

Our solution to this problem is generally to cut up the dataset in some fashion and run a number of similar analyses across partitions. Here, we're starting with a partitioning across "regions", ie Africa, Asia, Europe, North America, Oceania and South America.

This PR implements 7 build targets for these 6 regions plus the existing "global" build. Most of the work here went into clever subsampling by @emmahodcroft, @gtonkinhill and @rneher. However, there's additional logic surrounding coloring of tips. Only the "focal" region should have a full gamut of colors for location, division and county. Tips from non-focal regions are colored gray in this case.

### Testing
This has been testing with end-to-end builds locally. It has not yet been fully vetted on AWS.


scripts/priorities.py
```diff
@@ -0,0 +1,170 @@
...
+    #focal_seqs_array2 = np.array([list(str(focal_seqs[x].seq)) for x in focal_seqs])
+    mask_count = np.sum(np.logical_or(focal_seqs_array=='N',focal_seqs_array=='-'), axis=1)
+
+    # remove focal sequences from all sequence to provide context data set
+    keep = [(i, name) for i, name in enumerate(context_seqs_dict['names']) if name not in set(focal_seqs_dict['names'])]
+    context_seqs_dict['snps'] = context_seqs_dict['snps'].tocsr()[[i for i, name in keep],:].tocsc()
+    context_seqs_dict['names'] = [name for i, name in keep]
+
+    # for each context sequence, calculate minimal distance to focal set
+    d = calculate_distance_matrix(context_seqs_dict['snps'], focal_seqs_dict['snps'], consensus = context_seqs_dict['consensus'])
```
jameshadfield, line 141: `calculate_distance_matrix` does not use the `consensus` arg. Not being pedantic, rather want to check that the consensus wasn't intended to be used.

Snakefile_Regions
```diff
@@ -0,0 +1,184 @@
...
+
+rule subsample_focus:
+    message:
+        """
+        Subsample all sequences into a focal set
+        """
+    input:
+        sequences = rules.mask.output.alignment,
+        metadata = rules.download.output.metadata,
+        include = files.include
```
jameshadfield, line 36: I think the `include`s should be part of the context filter, not the focal filter. If they are part of the focal group they have an unwanted effect on priority calculations if the focal group is far away from the root.

scripts/priorities.py
gtonkinhill: Ah, yes this is no longer needed. It is used in the original `pairsnp` package to improve the performance of the algorithm on long bacterial alignments with many gaps. I removed it here to clean up the code but forgot to remove the argument.

PR #349, opened 2020-04-10
Add documentation describing the format of the metadata.

jameshadfield:
As we have more labs and groups running these pipelines with their own data, documentation becomes more important.
This commit explains the metadata TSV (I plan on adding more documentation and then adding links to them from the README)

### DIFF
--- a/docs/metadata.md
+++ b/docs/metadata.md
@@ -1,0 +1,161 @@
+## Main Metadata File
+
+Running the phylodynamics analysis described in this repo requires 2 main inputs: the sequence data, in FASTA format, and metadata in TSV format.
+Here we describe the format of the metadata which the pipeline expects.
+An example of this can be seen with the metadata [metadata.tsv](../data/metadata.tsv) for shared sequences which is what we use for the global analysis at [nextstrain.org/ncov/global](https://nextstrain.org/ncov/global). 
+
+#### A few points before we dive in:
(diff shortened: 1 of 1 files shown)


docs/metadata.md
```diff
@@ -0,0 +1,161 @@
...
+**Column 7: `country`**
+
+The country the sample was collected in. Our example, "NewZealand/01/2020", was collected in ....... New Zealand.
+You can run `tail +2 data/metadata.tsv | cut -f 7 | sort | uniq` to see all the countries currently present in the metadata.
+As of April 10 there were 64! 🌎
+
+**Column 8: `division`**
+
+Division currently doesn't have a precise definition and we use it differently for different regions.
+For instance, for samples in the USA we are detailing the state in which the sample was collected here.
```
emmahodcroft, line 78:
```suggestion
For instance, for samples in the USA we are detailing the state in which the sample was collected here. For other countries, it might be a county, region, or other administrative sub-division.
```

docs/metadata.md
```diff
@@ -0,0 +1,161 @@
...
+
+
+**Column 11: `country_exposure`**
+
+Analogous to `region_exposure` but for `country`.
+In our example, given the patient's travel history, this is set to "Iran".
+
+**Column 12: `division_exposure`**
+
+Analogous to `region_exposure` but for `division`.
```
emmahodcroft, line 105:
```suggestion
Analogous to `region_exposure` but for `division`. If we don't know the exposure division, we may specify the value for `country_exposure` here as well.
```

PR #352, opened 2020-04-10
handle >hCoV/Kuwait/KU001/2020 sequence in GISAID fasta

brianpardy:
The naming of the Kuwait/KU001/2020 sequence on GISAID uses a previously unseen prefix of "hCoV" that does not match the regex used in scripts/normalize_gisaid_fasta.sh, causing this sequence to be excluded from downstream analysis based on the GISAID download due to no matching sequence name in metadata.tsv. This change allows normalize_gisaid_fasta.sh to recognize and remove both the "hCoV" and "hCoV-19" prefixes.

Related to #53


brianpardy:
The sequence that prompted this PR has since been renamed in GISAID, using the same naming scheme as the other sequences. This PR is now unnecessary and I will close it.

PR #358, opened 2020-04-13
Add a check for minimum augur version

### LINKED ISSUE #336: Add snakemake check to require minimum augur (& treetime?) version
Will prevent issues such as #335. As we develop these analysis we are going to update augur in parallel and so enforcing a minimum augur version, and printing a helpful error message, will help everyone.

huddlej:
Avoid issues caused by outdated augur installations by explicitly checking for a
specified minimum augur version. If the appropriate augur version is installed,
the appropriate TreeTime should also have been installed via setup.py, so this
commit does not introduce a minimum TreeTime version.

Closes #336.

### DIFF
--- a/Snakefile
+++ b/Snakefile
@@ -1,6 +1,24 @@
 from os import environ
+from packaging import version
 from socket import getfqdn
 from getpass import getuser
+from snakemake.logging import logger
+import sys
+
+#
+# Verify that required versions of dependencies are installed.
+#
+MIN_AUGUR_VERSION = "7.0.2"
+
+try:
+    from augur.__version__ import __version__ as augur_version
+except ModuleNotFoundError:
+    logger.error("ERROR: Could not find augur. Follow installation instructions at https://nextstrain.org/docs/ and try again.")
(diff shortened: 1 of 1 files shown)


Snakefile
```diff
@@ -4,22 +4,28 @@ from socket import getfqdn
...
 #
 MIN_AUGUR_VERSION = "7.0.2"
-
 try:
     from augur.__version__ import __version__ as augur_version
 except ModuleNotFoundError:
     logger.error("ERROR: Could not find augur. Follow installation instructions at https://nextstrain.org/docs/ and try again.")
     sys.exit(1)
-
 if version.parse(augur_version) < version.parse(MIN_AUGUR_VERSION):
```
huddlej: This is a minor pedantic point, but I'd prefer to keep this whitespace for clarity. Since `try`/`except` can also end with an `else`, it's a little strange to see an `if` immediately follow in that same block.

jameshadfield: 👍 Removed my deletions

PR #362, opened 2020-04-14
Compute build title in Python instead of shell

tsibley:
Easier to understand and debug and keeps the rule's shell command
focused on the `augur export` invocation.

### Related issue(s)  
Emma's comments in Slack.

### Testing
I ran the following and made sure they produced the expected `augur export` invocations. I did not actually run the builds, though.

```
nextstrain build . -n results/ncov_with_accessions.json
nextstrain build . -n results/ncov_with_accessions_global.json
nextstrain build . -n results/ncov_with_accessions_africa.json
nextstrain build . -n results/ncov_with_accessions_north-america.json
```

### DIFF
--- a/Snakefile
+++ b/Snakefile
@@ -447,6 +447,17 @@ rule tip_frequencies:
             --output {output.tip_frequencies_json}
         """
 
+def export_title(wildcards):
+    region = wildcards.region
+
+    if not region:
+        return "Genomic epidemiology of novel coronavirus"
+    elif region == "_global":
+        return "Genomic epidemiology of novel coronavirus - Global subsampling"
+    else:
+        region_title = region.lstrip("_").replace("-", " ").title()
+        return f"Genomic epidemiology of novel coronavirus - {region_title}-focused subsampling"
+
 rule export:
     message: "Exporting data files for for auspice"
     input:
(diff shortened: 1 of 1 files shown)


Snakefile
```diff
@@ -447,6 +447,17 @@ rule tip_frequencies:
             --output {output.tip_frequencies_json}
         """
 
+def export_title(wildcards):
+    region = wildcards.region
+
+    if not region:
+        return "Genomic epidemiology of novel coronavirus"
+    elif region == "global":
```
emmahodcroft, line 455:
```suggestion
    elif region == "_global":
```
The region wildcard values come in with the attached underscores, so I think this will need to be like this - unless I'm missing somewhere where it's already been transformed so the `_` is gone?

tsibley: Ah, you are totally correct. I missed that the output of `_global` was "Global-focused subsampling" and not "Global subsampling".

tsibley: Repushed with your suggestion.

emmahodcroft:
This looks great Tom, thanks!!

tsibley:
@emmahodcroft Great! Please merge and push at your leisure then. :-)

PR #369, opened 2020-04-20
Move hardcoded settings from Snakefiles into a config file

huddlej:

Moves hardcoded parameters from Snakefile (e.g., clock rate for augur refine, etc.) into a YAML config file. Renames the existing config file from "Snakefile.yaml" to "config.yaml" to follow [standards recommended by Snakemake docs](https://snakemake.readthedocs.io/en/stable/snakefiles/deployment.html). Also adds validation of the config file with a simple config schema that can be expanded as necessary for critical parameters that would benefit from type checking (e.g., min length of sequences for augur filter).

This approach allows users to modify key parameters of the pipeline by editing a YAML file or passing them through the command line `--config` argument without needing to understand or edit Snakemake files.

## Discussion points

Moving all parameters into a single config file revealed that the regions and priorities rules both define a `subsample_focus` rule, but they use slightly different parameter values. Specifically, the regions rule uses 280 regional and 120 global sequences while the priorities rule uses 300 regional and 150 global sequences. Do these numbers need to be different or can we pick one set to use across both files?

There is a trade-off to moving parameters from the Snakefile to a separate file. On the one hand, this change increases the visibility of what can be configured in the pipeline, allows parameters to be validated, and exposes parameters to users through more standard interfaces (YAML files and command line arguments). On the other hand, this change separates the config parameters from their original context in the commands for each rule.

I've tried to address this latter change by organizing config parameters by rule name such that the config file organization generally mirrors the Snakefile organization. I have also added comments to clarify the purpose of specific parameters. Such comments could be extended to all parameters. This approach follows the example of Snakemake's creator's own [pipeline config files](https://github.com/snakemake-workflows/single-cell-rna-seq/blob/master/config.yaml).

## Testing

I've confirmed that the commands for base, region-specific, and priorities rules include the expected parameters loaded from the config file.

### DIFF
--- a/Snakefile
+++ b/Snakefile
@@ -3,8 +3,12 @@ from packaging import version
 from socket import getfqdn
 from getpass import getuser
 from snakemake.logging import logger
+from snakemake.utils import validate
 import sys
 from shutil import which
+
+configfile: "config/config.yaml"
+validate(config, schema="schemas/config.schema.yaml")
 
 #
 # Verify that required versions of dependencies are installed.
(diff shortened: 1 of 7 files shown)


huddlej:
@cassiawag I've tagged you for review here, too, if you're interested. It would be great to know how this affects your interactions with the build (or if it does at all).

cassiawag:
@huddlej -- would you prefer I look at this PR or #375?

huddlej:
@cassiawag #375 might be more helpful, actually. Thank you!

huddlej:
This PR was superseded by #375.

PR #371, opened 2020-04-21
Edits to tutorials

cassiawag:
Hi @jameshadfield, I've added a few edits for clarity and readability to  `docs/metadata.md` and `docs/running.md`. 

Overall, I thought the docs made sense and were easy to follow. I have only a few comments on `running.md`, which are not addressed in the PR:
* Under "Running the default build," the fact that only one json is outputted for ncov while two json's  are outputted for the Zika tutorial might be confusing. A statement along the lines of "Don't worry", explaining why there's only one json could be useful.

* "Debugging common issues" should include a link to another location with more fleshed out Nextstrain troubleshooting help (maybe the augur docs?). Running the pipeline seems simple enough, but a lot of errors could come up (likely related to package version control). For example, I got IQ Tree errors for awhile, and I had to solicit troubleshooting help on how to force conda to install the right version.

* At the top (maybe under "A few points before we dive in") include a link for how to install Nextstrain. For example: "This tutorial assumes Nextstrain is installed locally. If Nextstrain is not already installed, please do so at LINK before continuing on.."

### DIFF
--- a/docs/metadata.md
+++ b/docs/metadata.md
@@ -2,7 +2,7 @@
 
 Running the phylodynamics analysis described in this repo requires 2 main inputs: the sequence data, in FASTA format, and metadata in TSV format.
 Here we describe the format of the metadata which the pipeline expects.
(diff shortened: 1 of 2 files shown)


jameshadfield:
Thanks Cassia!

PR #375, opened 2020-04-24
Simplify workflows for default and regional builds

huddlej:
Demonstrates alternative approaches to constructing a Snakemake workflow with:

  - a single minimal Snakefile entry point
  - a single commented config file replacing hardcoded parameters in Snakefiles
  - conda environments per rule to enforce dependency availability and versions
  - profiles for custom Snakemake execution per group (e.g., Nextstrain team)
  - use of output references instead of explicit filenames (e.g., rules.tree.output.tree)
  - grouping of wildcard outputs by directory to clarify output organization
  - separation of common and custom rules into separate included Snakemake files (*.smk)
  - conditional rule execution using input functions instead of shell logic and placeholder rules
  - benchmarking of long-running jobs with benchmark directives

Example build execution looks like the following:

```bash
# Default build ("global" region only).
snakemake

# Region-specific builds for subset of regions.
snakemake --config regions="africa asia"

# Build tuned for Nextstrain team
snakemake --profile profiles/nextstrain

# The above command expands to the following:
snakemake \
  --configfile config/config.yaml \
               profiles/nextstrain/regions_config.yaml \
  --cores 4 \
  --printshellcmds \
  --reason \
  --stats regions_stats.json
  --use-conda
```


rules/regions.smk
```diff
@@ -0,0 +1,222 @@
...
+        """
+    input:
+        metadata = rules.download.output.metadata
+    output:
+        metadata = REGION_PATH + "metadata_adjusted.tsv"
+    conda: "../envs/nextstrain.yaml"
+    shell:
+        """
+        python3 scripts/adjust_regional_meta.py \
+            --region "$region" \
```
jameshadfield, line 163: This expansion doesn't work for me, and seems to be expecting a bash variable `$region` to exist. I changed it to the following for testing 
```
    params:
        region = "{region}"
    conda: "../envs/nextstrain.yaml"
    shell:
        """
        python3 scripts/adjust_regional_meta.py \
            --region {params.region} \
```

huddlej:
I forgot to include this in the description, but here are the before and after DAG views for both the default builds and the regional builds.

Original default build DAG:
image

New default build DAG:
image

Original regional builds DAG:
image

New regional builds DAG:
image

rules/regions.smk
huddlej: Good catch! I fixed this with the wildcard value.

huddlej:
Thank you for the detailed comments, @jameshadfield! I actually didn't fully understand how the global subsampling worked until we talked, so we both benefited from that call. :)

I have pushed commits that address the following requests:

- [x] gitignore benchmarks
- [x] fix straggling bash variable
- [x] add subsampling rule for global build
- [x] update documentation text in `running.md` to reflect new commands and outputs
- [x] add `resources` to tree and refine rules
- [x] delete Snakefile_Regions

I've verified that the subsampling works as expected for the global and regional builds. Here is a subset of the DAG for the "africa" and "global" builds. This shows how the global build is almost the same as the regional builds except that it skips the context subsampling.

image
 
The documentation in `running.md` still needs to be updated for the figure to be correct. I wonder if there is a textual representation of that same information that we could use instead (like [embedding the graphviz representation of the DAG somehow](https://github.com/TLmaK0/gravizo))?

Regarding minor questions:

> It's unfortunate that one can't overwrite the config parameters for nested dictionaries via `--config`.

I agree that this seems like a limitation of Snakemake. Although one could argue that if a configuration parameter is important enough to be used from the command line frequently, it should be in a configuration file in version control anyway. If this interface becomes an issue, we can always move commonly overridden config parameters to the top-level of the file which will expose them to the command line interface.

> I think it'd be cleaner to write the generated colours files (e.g. colors = "config/colors_{region}.tsv") into the results/region subdirectory.

I agree. I wasn't sure if people modify these colors files after the fact and track them somehow or are these files always autogenerated and never manually edited?

> The conda env is not created by default

If you already have a Nextstrain environment, you don't have to do anything differently. But if you don't have the environment already, using `--use-conda` will create it for you. I'm realizing now that we probably still need the checks at the beginning of the Snakefile for the correct augur version and binaries, since there is no way to force `--use-conda`. I have updated the tutorial in `docs/running.md` to include a default profile, though, so we could push people towards the conda environments by adding `use_conda: True` to the default profile.

Also, the `--use-conda` flag won't work in the Docker environment since the Nextstrain Docker image doesn't currently have conda installed. Tom and I have talked about how to move in that direction and have some prototypes in place, but we haven't had time to finish this up.

> One thing I'd like to parameterise is to easily say which node-data JSONs export wants (and therefore which rules run).

We do this for [the seasonal flu builds](https://github.com/nextstrain/seasonal-flu/blob/bb72f12392f96bdcc3bdaffb6dea4211c525ea19/Snakefile#L37-L67) where we only want to export specific JSONs for specific subsets of the builds. I haven't tried parameterizing this from the configuration file, but it is possible in principle.

> Over time the hardcoded stuff in _get_sampling_trait_for_wildcards etc should be made into a config setting and generalised.

Yes, that would be great!

## Next steps

I need to make sure that all of the outputs generated by the current master branch for [the staging builds](https://github.com/nextstrain/ncov-ingest/blob/master/bin/rebuild-staging) work as expected. If they do, maybe we could get some eyes on the resulting JSONs and confirmation that these look reasonable before merging this PR.

I would also love to get feedback from anyone else who regularly runs the pipeline to see how this current approach could be improved (@emmahodcroft and @cassiawag might be interested here).

config/config.yaml
```diff
@@ -0,0 +1,123 @@
+# This file contains defaults for the "config" object used in the Snakefile.
+# To temporarily override or provide a value, you can use snakemake's --config
+# or --configfile options.
+---
+s3_staging_url: s3://nextstrain-staging
+slack_token: ~
+slack_channel: "#ncov-gisaid-updates"
+sequences: "data/sequences.fasta"
```
emmahodcroft, line 8:
```suggestion
# These are the two main starting files for the run
sequences: "data/sequences.fasta"
```

config/config.yaml
```diff
@@ -0,0 +1,123 @@
+# This file contains defaults for the "config" object used in the Snakefile.
+# To temporarily override or provide a value, you can use snakemake's --config
+# or --configfile options.
+---
+s3_staging_url: s3://nextstrain-staging
```
emmahodcroft, line 5:
```suggestion

# S3 and slack options are related to Nextstrain live builds and don't need to be modified for local builds
s3_staging_url: s3://nextstrain-staging
```

I know this seems a little silly but if something's not working people zero in on this kind of stuff amazingly quickly... Anything to help convince people that it's definitely not a problem with the slack-channel name is helpful! :)

config/config.yaml
```diff
@@ -0,0 +1,123 @@
...
+# configuration files.
+files:
+  include: "config/include.txt"
+  exclude: "config/exclude.txt"
+  reference: "config/reference.gb"
+  outgroup: "config/outgroup.fasta"
+  ordering: "config/ordering.tsv"
+  color_schemes: "config/color_schemes.tsv"
+  auspice_config: "config/auspice_config.json"
+  auspice_config_gisaid: "config/auspice_config_gisaid.json"
```
emmahodcroft, line 25: Can we move files like these somewhere where they only are listed for 'our' runs? Or is that too complicated? I just worry as with the above that people tend to get really caught up in this kind of thing, and these are files that really only we have to worry about. 
However, this is a fairly minor concern.

config/config.yaml
```diff
@@ -0,0 +1,123 @@
...
+#
+# Priorities-specific settings
+# TODO: consider how to merge these with the region-specific settings to avoid redundancy.
+#
+
+regions_for_priorities: ["africa", "asia", "europe", "north-america", "oceania", "south-america"]
+
+# Subsampling settings for region of focus and context.
+# TODO: merge these with "subsample_focus" settings above
+subsample_focus_for_priorities:
```
emmahodcroft, line 118: I'm really sorry but I can't figure out where these are used, or where they come from!

emmahodcroft:
Hi John

This is fantastic! Thank you so much for the hard work! And my apologies for the delay reviewing it. I'm still wrapping my head around how it all ties together, I'll admit. I think the runs are tidied up and more flexible, but it's a little hard for me to trace what commands come from which files - but I think this is probably mostly adjustment!

I made a few small comments in the code but mostly suggestions.
I also added some code to `regions.smk` which I think was added to `Snakefile_Regions` in between this PR being born and now - it runs a python script to help identify missing colors/lat-long when doing the final clean up for runs.

This might come after this PR (I expect it will) but I think it would be nice to separate out the rules a little more - `builds.smk` still has some stuff that's only used by us (the GISAID and ZH etc - and do we want a 'default' build to do all the stuff with the colors and exposure history?) - is it possible to separate this out into a further file that's linked by the `nextstrain` profile? (Or just at the bottom of `Snakefile`, I guess?) However, this'll take a little thinking, I suppose, as the 'end' files are slightly different then, in the two cases.
As I said might be better to tackle this after this PR is set and merged. 

I remain slightly unsure how to run specific rules - see top of `regions.smk` - it seems like these are essentially the same - tacking `clean_export_regions` onto a new call - but I'll have to do a build somewhere to verify this - and with the call being different, can't quickly put this on our cluster at the moment!

I'll try to give this a trial though so I'm sure I understand it - then we can update these instructions, which is important for everyone running the builds. 

On a similar note, we'll need to update the AWS call for `ncov-ingest` when we merge this, to keep the Builds flowing, yes? (Imagine now hard though)

Thank you so much again for the good work @huddlej , and thanks for providing already great comments @jameshadfield !

I am also curious to see how it is to @cassiawag as I think it's important this makes sense to those less familiar with Snakemake and I do think moving to multiple files maybe makes this a little less intuitive - I think we can happily solve any lingering confusion with more commends & docs, but it'll be good to know exactly where the unclear things lie!

rules/regions.smk
```diff
@@ -46,18 +41,21 @@ rule clean_export_regions:
...
 # Allows 'normal' run of export to be forced to correct lat-long & ordering
 # Just runs this, not ZH & GISAID, to speed up & reduce errors.
+# Runs an additional script to give a list of locations that need colors and/or lat-longs
 rule export_all_regions:
     input:
-        colors_file = expand("config/colors{region}.tsv", region=REGIONS),
-        auspice_json = expand("results/ncov_with_accessions{region}.json", region=REGIONS),
+        colors_file = expand("config/colors_{region}.tsv", region=REGIONS),
+        auspice_json = expand(REGION_PATH + "ncov_with_accessions.json", region=REGIONS),
         lat_longs = files.lat_longs,
```
emmahodcroft, line 55:
```suggestion
        lat_longs = config["files"]["lat_longs"]
```
This is giving me an error - the above seems to work - is that right?

emmahodcroft: Ps - Not sure why this is showing on the 'wrong side' of the file.... Hope it works anyway.

cassiawag:
I'll try to get to this sometime today since I've now gotten tagged by a few people :) I haven't been running ncov builds very frequently, but hopefully I can provide the perspective of someone with less Snakemake mastery.

emmahodcroft:
Just been giving this a trial run on my personal laptop - I am finding John's 'example' run commands in the OP really helpful - maybe we should put these into `Snakefile`? For me that would be the obvious place to look for basic info on 'how do I actually start this thing'?

emmahodcroft:
Sorry for more continuous updates as I try things out... I know multiple posts rather annoying!

Running ` snakemake --config regions="asia south-america"` gives an error because the `scripts/adjust_regional_meta.py` file expects region names to come in already stripped of the `-`. (It uses a map that includes `south america`)

However, running ` snakemake --config regions="asia south america"` also errors - because now `south` and `america` are each interpreted as regions. 

Easiest fix seems to be to modify `scripts/adjust_regional_meta.py` so that it expects `south-america`. Have done this locally, but will let run continue in case of unintentional knock-on effects before I push a commit... (hopefully I'll remember)

emmahodcroft:
Did full trial of `asia` & `south-america` on my personal laptop - haven't had time to compare results, but after couple of minor bug-fixes, looks like it's working well!! :tada:

huddlej:
Thank you for testing this out, @emmahodcroft! I have a bunch of responses to your comments which I'll post later (when the small child is asleep), but the gist is that we should be able to address all the issues you mentioned especially making the separation between default and Nextstrain builds clearer. Stay tuned for more later tonight PST!

config/config.yaml
huddlej: These are not currently used. I need to test out the priorities file and move the rules into the Nextstrain-specific builds file, unless they are not used anymore.

huddlej:
@emmahodcroft, I've tried to address all major comments below. Thank you again for trying this out and taking the time to review it!

> it’s a little hard for me to trace what commands come from which files

I have this same issue! The more I work with these files, the more I think the “regions” rules can be integrated into the “builds”. It is confusing to separate the subsampling into a separate file when we want to allow others to introduce their own subsampling by region, division, etc. as needed.

I’ve updated the builds and regions files to move the Nextstrain-specific rules into the regions file and move the shared rules into the builds file. Hopefully this should make the files a little more readable. Let me know what you think though.

> it would be nice to separate out the rules a little more - builds.smk still has some stuff that’s only used by us (the GISAID and ZH etc - and do we want a ‘default’ build to do all the stuff with the colors and exposure history?) - is it possible to separate this out into a further file that’s linked by the nextstrain profile?

Yes! I didn’t fully understand how these rules are currently used, so I left them in place. I have moved these into the regions file, but maybe we should rename this to `alternate_exports.smk` or something more accurate. The domain knowledge of how these rules are used should guide the name of the file where these rules live.

> I remain slightly unsure how to run specific rules

These region-specific rules that apply to multiple outputs needed to be updated. I’ve updated these to reflect the new directory structure. Below is the list of the original and new commands to run these rules (to the best my understand of how these are used).

Original commands:

```
# To build all regions and additional auspice JSONs
# (including GISAID and ZH JSONs). 
snakemake -s Snakefile_Regions all_regions

# To update ordering/lat_longs after AWS download:
#   snakemake --touch --forceall -s Snakefile_Regions
#   snakemake -s Snakefile_Regions clean_export_regions
#   snakemake -s Snakefile_Regions export_all_regions
```

New commands:

```
# To build all regions and additional auspice JSONs
# (including GISAID and ZH JSONs). 
snakemake --profile profiles/nextstrain all_regions

snakemake --touch --profile profiles/nextstrain
snakemake --profile profiles/nextstrain clean_export_regions
snakemake --profile profiles/nextstrain export_all_regions
```

Does this seem like a correct translation of the commands?

>  ‘example’ run commands in the OP really helpful - maybe we should put these into Snakefile

I would expect to find these in the README or the running tutorial. I’ve updated the latter docs to reflect these examples, to a degree, but maybe they could be clearer.

It doesn’t hurt to include them in the Snakefiles as long as they remain up-to-date. The subsampling description in `regions.smk` is a good example of how quickly documentation that includes specific details can diverge from the truth.

> Easiest fix seems to be to modify `scripts/adjust_regional_meta.py` so that it expects `south-america`.

Awesome! Thank you for catching this and making the fix!

> we’ll need to update the AWS call for ncov-ingest when we merge this, to keep the Builds flowing, yes?

Yes, I’ll make a PR for this in that repo and we can merge it after we merge this.

I can also move configuration parameters that are specific to our team into our profile's config file to address the comments you've made inline.

emmahodcroft:
> I’ve updated the builds and regions files to move the Nextstrain-specific rules into the regions file and move the shared rules into the builds file. Hopefully this should make the files a little more readable. Let me know what you think though.

This looks great! Thanks!! 

>Yes! I didn’t fully understand how these rules are currently used, so I left them in place. I have moved these into the regions file, but maybe we should rename this to alternate_exports.smk or something more accurate. The domain knowledge of how these rules are used should guide the name of the file where these rules live.

Would it be hard or not-logical to make another file that does things like the GISAID, Chinese, etc? So they'd only run for the `nextstrain` profile. As, they shouldn't really be part of a region run either (as someone else may want to run and modify their own region build - just as eventually we'll have a country version of the same... but they probably will never want a 'GISAID' version of our output! 😆 )
However this is low priority and I think we can do this later, too. No urgency for it here.

> Does this seem like a correct translation of the commands?

I have not had a chance to test them all, but I ran a few with the `--config` option last night (when running just Oceania & S America), and had the same pattern you've used here - so it seems like this should work! For the `--config` last night I had to be careful to put the rules before `--config` though - I don't know if order matters here. Unfortunately a little hard for me to test as running the whole region build is pretty nuts on a laptop - and I'l need to modify our cluster call to get it working there... will try to do this though! However, we can also sort this out after merge - will become very apparent if they don't work! 🙃 

>It doesn’t hurt to include them in the Snakefiles as long as they remain up-to-date. The subsampling description in regions.smk is a good example of how quickly documentation that includes specific details can diverge from the truth.

Yea, for me the problem is always 'which nextstrain documentation is this in?'.... I am never fully sure where to look for our own stuff (probably not a good sign)...

I imagine the running commands will stay fairly similar (apart from major changes like this). However I take your point. How about a full link to the readme right at the top of the main snakefile, so that people like me who forget where the documentation for this is, can go straight there? 😁 

For the 'region's file though - I like having these commands there at the top. I have apparently 0 memory so I use these every time I do a run. However, I can thus then guarantee these will stay up to date!

>I can also move configuration parameters that are specific to our team into our profile's config file to address the comments you've made inline.

Again, no hurry here, these things we can also tidy after push, if it's easier. But might be good to avoid confusion about 'why I don't have a _GISAID file' etc. 

Overall I think this is looking good, thanks so much for the hard work John!

I'd still really like to see comments from someone a bit less familiar with Snakemake to ensure people can figure out how to start and trace their way through - in case we could really help this with some other simple instructions or getting-starting info. Apart from this though, I think this is looking really good :)

huddlej:
@cassiawag I've put together a quickstart for myself just to make sure I understand the whole flow of analyses from the perspective of a fresh start. If you have time, it would be amazing to hear your feedback on the process outlined below. I guess you'd probably start from checking out the `simple-build` repository and then skip the data download, but if you wanted to try the whole thing from start to end that's ok, too.

## Quickstart for simpler ncov builds
Install [Docker](https://docker.com/) and start the Docker client.
Install Nextstrain’s command line interface (CLI).

```
python3 -m pip install nextstrain-cli
```

Make sure the CLI is setup properly.

```
nextstrain check-setup
```

Alternately, see [full instructions to install the Nextstrain CLI](https://github.com/nextstrain/cli#installation).

Clone the ncov repository.

```bash
git clone https://github.com/nextstrain/ncov
cd ncov
```

Check out the simple build branch.

```
git checkout simple-build
```

Prepare a directory for the ncov data.

```
mkdir -p data
```

[Download sequences FASTA and metadata TSV from GISAID](https://www.gisaid.org/) into `data/`.
You will need a free GISAID account to download these data.

Decompress GISAID data.

```bash
bzip2 -d -c data/sequences.fasta.bz2 > data/sequences.fasta
bzip2 -d -c data/metadata.tsv.bz2 > data/metadata.tsv
```

Run builds with the Nextstrain CLI from the `ncov` directory.
The following commands include potential common scenarios.

```
# Run the default ("global") build.
nextstrain build .

# Run the default build with reasonable Snakemake defaults.
nextstrain build . --profile profiles/default

# Run region-specific builds for a subset of regions.
nextstrain build . --config regions="africa asia"

# Run the Nextstrain team build with nextstrain-cli.
nextstrain build . --profile profiles/nextstrain all_regions

# The above command expands to the following:
nextstrain build . \
  --configfile config/config.yaml \
               profiles/nextstrain/regions_config.yaml \
  --cores 4 \
  --printshellcmds \
  --reason \
  --stats stats.json \
  all_regions
```

Visualize the resulting builds.

```
nextstrain view auspice/
```

For next steps, see [the full guide to running ncov builds](https://github.com/nextstrain/ncov/blob/master/docs/running.md) including descriptions of how to:

* troubleshoot common issues
* modify build parameters in `config/config.yaml`
* modify the default profile in `profiles/default/config.yaml` to use different default Snakemake arguments

rneher:
Great work @huddlej! One quick thought. would it make sense to abstract the environment into the config file? I am thinking of development work where one might want to use different dependencies (like an experimental augur version).

cassiawag:
Hi @huddlej, amazing work on all of this! 

Unfortunately, because I don't have Nextstrain installed via Docker (I haven't tried since last summer to do so, but at the time I ran into insurmountable issues with running it all under my windows sub-system linux configuration), I can't follow your above instructions. How would I run it with the appropriate profiles and defaults using the standard `snakemake` command method?

I think the modularizing is a really good move for the ncov build. It provides more flexibility on regional builds and is much more organized. However, it needs an overview somewhere on how to actually run it.  As a Snakemake beginner who had gotten used to the previous way of running ncov builds, I'm finding it difficult to follow what exactly Snakemake is doing and where I need to be changing defaults to get the result I want. The difference between profiles, config files and setting regions is not intuitive to me at least. Mixing in this folder structure with the already existing folder structure

huddlej:
@rneher Yes, it would make sense to abstract the conda environment into a configuration variable. I'll add this today.

@cassiawag I'm sorry I forgot you were not using Docker! Here are the instructions for the snakemake version (replacing `nextstrain build .` with `snakemake`):

```
# Run the default ("global") build.
snakemake

# Run the default build with reasonable Snakemake defaults.
snakemake --profile profiles/default

# Run region-specific builds for a subset of regions.
snakemake --config regions="africa asia"

# Run the Nextstrain team build with nextstrain-cli.
snakemake --profile profiles/nextstrain all_regions

# The above command expands to the following:
snakemake \
  --configfile config/config.yaml \
               profiles/nextstrain/regions_config.yaml \
  --cores 4 \
  --printshellcmds \
  --reason \
  --stats stats.json \
  all_regions
```

I totally see what you mean about there being too many directories and files to keep track of. I'll write some documentation to clarify how these directories and files interact. Your suggestion of moving the snakemake directories into a subdirectory is a great one and actually [the recommended layout from the snakemake docs](https://snakemake.readthedocs.io/en/stable/snakefiles/deployment.html#distribution-and-reproducibility) (they use `workflow` for the subdirectory).

Let me know if the snakemake commands above work for you or not and I'll post some updated documentation here a little later today.

huddlej:
I've added some preliminary documentation of the workflow directory structure and steps executed in default and Nextstrain builds in the graph representation:

https://github.com/nextstrain/ncov/blob/simple-build/docs/workflow_structure.md

I've also moved the default conda environment into the configuration file such that users can override the default environment with their own test environments like so:

```
snakemake --use-conda --config conda_environment="../envs/experimental_features.yaml"
```

cassiawag:
The documentation looks great, John. It helps me understand more clearly what's going on. Of note, I just deleted the `background` folder on the master branch. I created it to house translations of background documents in from the nextstrain.org repo when we were doing translations in the ncov repo. 

Also, your commands worked! I was able to get it to run on my computer.

huddlej:
Great! I'm glad the docs helped, @cassiawag, and that the commands worked. I've updated the workflow structure to remove the background folder and also added these docs to the top-level README.

emmahodcroft:
Hey @jameshadfield  - regarding your first and last comment! So the Snakefile_Priorities is intended to just run priorities on a whole (un-sub-sampled) then, you can create a tanglegram of what's been sampled by Regions builds and what their priority was (it ends up as a color-by) on the big tree.

My thinking is that this should be combined with your PR for the un-sub-sampled build & adapted to work with the new snakefile format - and then set to run automatically once or twice a week (if such a thing is possible?) - giving us a big un-sub-sampled tree with the region (& whatever other countries/etc we might want to add in future) priorities on it as color-bys, for ease of comparison.

PR #376, opened 2020-04-24
Reduce inputs to augur refine with a conditional input function

huddlej:
Checks whether the current region is empty (the default build) or one of the
specific regions and only returns the full multiple sequence alignment as input
to augur refine in the default case. This should reduce the size of inputs to
augur refine for regional builds to just those sequences in the subsampled
alignment and reduce the amount of memory required for the refine step.

### DIFF
--- a/Snakefile_Regions
+++ b/Snakefile_Regions
@@ -193,7 +193,7 @@ rule subsample_regions:
         rules.subsample_focus.output.sequences,
         rules.subsample_context.output.sequences
     output:
-        "results/subsampled_alignment{region}.fasta"
+        alignment = "results/subsampled_alignment{region}.fasta"
     shell:
         """
         python3 scripts/combine-and-dedup-fastas.py \
--- a/Snakefile_Regions
+++ b/Snakefile_Regions
@@ -198,7 +198,7 @@ rule subsample_regions:
         """
         python3 scripts/combine-and-dedup-fastas.py \
             --input {input} \
-            --output {output}
+            --output {output.alignment}
         """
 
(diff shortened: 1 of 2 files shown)


Snakefile_Regions
```diff
@@ -193,7 +193,7 @@ rule subsample_regions:
         rules.subsample_focus.output.sequences,
         rules.subsample_context.output.sequences
     output:
-        "results/subsampled_alignment{region}.fasta"
+        alignment = "results/subsampled_alignment{region}.fasta"
```
emmahodcroft, line 196: If we change this here, do we need to change to `{output.alignment}` in line 201 below?

huddlej: No, since there is only one output, Snakemake will output the single path like this:

```
python3 scripts/combine-and-dedup-fastas.py \
  --input results/subsample_focus_asia.fasta results/subsample_context_asia.fasta \
  --output results/subsampled_alignment_asia.fasta
```

It is good practice to refer to the output by name though, so I should have updated that. I can make a push a little later with that change.

emmahodcroft: Not a bit deal at all! I just wasn't sure and didn't have time to investigate further when I wrote it. :)

emmahodcroft:
I haven't tested this locally (sorry) but this looks 100% sensible to me. If you've done a test John and are happy that it seems to work, then I'm happy for this to be merged 👍

huddlej:
Running a test with a full regional build for Africa on our local cluster (to avoid the sounds of my laptop taking off from the runway) and will merge after confirming that works ok.

PR #377, opened 2020-04-25
Fix typo in file from ilke to like on line 13 of ncov_trees-background.md

redrolor:

### Related issue(s)  

Fixes #  
Related to #  

### Testing


cassiawag:
Thank you for the typo fix!.
However, those are just comments for translators and are not shown on the live site. I don't think this needs to be merged since the file has already been translated.

PR #378, opened 2020-04-26
Use a curated alignment

jameshadfield:
I'm becoming increasingly convinced that we should be using a hand curated alignment and adding new sequences into this. The ~10k [global non subsampled build](https://nextstrain.org/ncov/non-subsampled) indicates a number of clades which appear to be the product of poor alignment. To explore this & promote discussion I've made this simple PR, which is a single commit (86561f3) on top of @huddlej's PR.

In this PR the `download` rule grabs a hand-curated alignment and we align all new sequences to that using `mafft`'s `--add` option. This will allow us to then subsample for different builds from the entire alignment.

---

This PR only focuses on the sequences in the following screenshot from the global non-subsampled build, where a clade is defined by a number of consecutive SNPs which are hypothesized to be alignment errors:
image

This structure is not present in this PR (screenshot below). I didn't even have to edit the alignment by hand, so I think the SNPs were erroneously introduced by the alignment approach we currently employ in master. 

image

---

@trvrb @emmahodcroft @huddlej @rneher do you think this approach is worth pursuing?

(Note that this PR only works if a fix to augur bug https://github.com/nextstrain/augur/issues/524 is present.)


rneher:
I agree that these clades are likely alignment issues and we will run into more and more of these issues as sequences accumulate. Furthermore, it seems pretty silly to realign everything everytime. Would it make sense to store batches of alignments (say in the order of submission) and only ever add new one sequences. This might have to happen outside the ncov build but instead in the ingest step.

emmahodcroft:
I also agree I think this is a good way to go - some things we see are definitely alignment artefacts and it would be great to be rid of them. (Also I think a number of sequences currently in `exclude` could go back in then, as often a lot of those I think were alignment problems. 

I do think we would need to set up so that we check our current alignment on S3 against any chances that have happened to GISIAD sequences, and if so, remove those from the aligned set and re-align them with the latest sequence - since people can change the sequence in GISAID! But this shouldn't be hard to do. 

Thanks James!

jameshadfield:
> we would need to set up so that we check our current alignment on S3 against any chances that have happened to GISIAD sequences

This is a really good point. And we can't just diff the sequences as part of manual curation may involve trimming spurious bases which flank dropout-regions. When a gisaid sequence is updated, is there some other field which changes? E.g. `covv_subm_date` or some other field that we're not seeing in our JSON?

emmahodcroft:
Hmm. @tsibley might be best placed to comment on if any field reliably changes when a sequence changes - I don't think we currently check for this at all (we've never needed to worry about it) so I've never noticed anything. 
I don't think `covv_subm_date` will work (though could be wrong) as I know that if sequences are removed entirely & then re-added, they keep that date the same (we saw this with the many Seattle sequences that went 'up and down' for a bit while metadata fixes were applied). 

What might be safest is to also save an 'unaligned' version alongside our aligned version - with 1-to-1 mapping of the sequence names - and just diff the 'original' sequences - and use those to remove any sequences from our aligned version (and the 'raw' version) that have changed.

huddlej:
This seems like a good idea. I had a couple thoughts about implementation details:

 - I can imagine public health labs and other groups may want to have their own curated alignment(s) of their local sequences (not yet submitted to GISAID) and another curated or automated alignment of GISAID sequences. At some point the pipeline will need to allow them to merge these alignments.
 - We can implement the alignment rules in this repository but as part of a separate DAG from the main pipeline. This will allow anyone to run, for example, `snakemake realign` as needed without that rule being run for the standard auspice exports. Anything we do for ourselves will likely be helpful for other groups, so keeping the rules in one repository will make it easier to share our work with those groups.
 - Anything we can do to make runs appear as deterministic as possible would be ideal. Any rules that are triggered by an apparently stochastic input (e.g., updating when a metadata field changes in the download from GISAID) could be confusing to new users and make the workflow difficult to debug for experienced users when things go wrong. If we can implement this curated alignment approach with explicit, deterministic behavior, we could minimize future confusion.

I realize I'm not providing any specific solutions to the first and third points, but I wanted to at least mention them. :)

tsibley:
I agree with all of @huddlej's points.

I also agree with @emmahodcroft's suggestion to do a direct comparison of new sequences against the raw sequences that went into the curated alignment. This is the simplest, most robust method as it relies on the relevant data itself, not any associated metadata (which may not correctly flag changes).

> What might be safest is to also save an 'unaligned' version alongside our aligned version - with 1-to-1 mapping of the sequence names - and just diff the 'original' sequences - and use those to remove any sequences from our aligned version (and the 'raw' version) that have changed.

I believe the "unaligned" version could just be the input fasta we gave to mafft when initially making the curated alignment, right? I'm assuming the curated alignment process looks like:

    raw sequences → (mafft) → initial base alignment → (manual curation) → curated base alignment

where the curated base alignment is used in each build like so:

    curated base alignment + new raw sequences → (mafft) → latest alignment

emmahodcroft:
Yes Tom - I agree! Just to ensure we save a copy of the 'pre-curation' sequences (all of them, even as new one gets added) to diff against any changes, rather than 'just' keeping the curated alignment. But I think you got that!

jameshadfield:
We pursued a different approach here (see #558)

PR #388, opened 2020-05-04
Generic subsampling

rneher:
(mostly by @huddlej )
This PR builds on PR #375 with the aim to have an explicit and generic way to run analyses at various levels of geographic resolution. There are multiple things that still need sorting out, including:

 - [x] adjust_metadata_regions needs to know what it is supposed to do for non-standard runs or be bypassed
-  [x] the specification of priorities is still to be worked out. we could imagine using a similar pattern for flu where instead of `proximity` we use `titer` priorities.  

### Directory and config structure

This PR would produce a results directory structure like so:
```
.
├── country
│   ├── switzerland
│   └── usa
├── division
│   └── washington
├── region
│   ├── global
│   └── north-america
├── split_alignments
└── split_sequences
```

The builds are configured as follows. Trait reconstruction can be made region specific:
```
traits:
  north-america:
    sampling_bias_correction: 2.5
    columns: ["country_exposure"]
```
and subsampling is defined as 
```
subsampling:
  # Default subsampling logic for regions
  region_global:
    global:
      seq_per_group: 20

  region:
    # Focal samples for region
    region:
      seq_per_group: 260
      exclude: "--exclude-where 'region!={location_name}'"
    # Contextual samples for region from the rest of the world
    global:
      group_by: "country year month"
      seq_per_group: 20
      exclude: "--exclude-where 'region={location_name}'"
      priorities:
        type: "proximity"
        focus: "region"
````
Each build can in principle be a sum or arbitrary subsamples. The defaults can be overwritten. 
the analysis to be run are specified as 
```
# Define locations for which builds should be created.
# Builds will be subsampled according to their geographic specificity
# based on the subsampling definitions below.
locations:
  region:
    - global
    - north-america
  country:
    - switzerland
    - usa
  division:
    - washington
```

### CODE REVIEW

**Comment by huddlej** [comment]
@tsibley notes:

> FWIW, in ID3C we call location_type the location's scale to avoid the overloaded term type, which might also mean hospital, residence, restaurant, etc. ID3C also supports an explicit hierarchy in its location catalog, which seems like it'd be useful for the mapping of city to country, for example. Finally, https://github.com/nextstrain/augur/issues/430 seems relevant here.

We have two other issues that need to be resolved (in addition to Richard's list above):

  - [ ] division names are not unique, so we can't use these values alone as the wildcards and confidently look up the corresponding region and country information
  - [ ] even if division names were unique, we would need a lookup table to get the region/country for a given division.

One workaround to these issues might be to require the region in a country’s output path and the region and country in a division’s path like so:

```
snakemake -np results/region/europe/tree_raw.nwk
snakemake -np results/country/europe_iceland/tree_raw.nwk
snakemake -np results/division/europe_iceland_reykjavik/tree_raw.nwk
```

Then Snakemake can parse out the different hierarchical bits as needed. This could have a nice side effect that the auspice output would look like the following and map nicely to hierarchical URLs:

```
ncov_europe.json -> /ncov/europe/
ncov_europe_iceland.json -> /ncov/europe/iceland/
ncov_europe_iceland_reykjavik.json -> /ncov/europe/iceland/reykjavik/
```

**Comment by tsibley** [comment]
Copying further Slack conversation about this to surface it so it doesn't get lost:

Thomas Sibley  22 hours ago
Encoding the hierarchy implicitly in requested outputs could work, but seems like externalizing the data in a weird way to me. Two questions that come to mind:
1. What happens if you ask for an invalid hierarchy, e.g. ncov_north-america_iceland.json?
2. How are those path/URL-safe locations labels (iceland, north-america) mapped/to from the canonical labels in the tsv file (Iceland, North America)? A strict 1:1 character substitution isn't quite sufficient when you consider locations like Braine-le-Chateau.

Thomas Sibley  22 hours ago
(2) is an issue regardless, I think. But I haven't looked at the draft of your PR, so maybe you already account for this! :-)

John Huddleston  22 hours ago
:smile: No, (2) is exactly the kind of thing that I stopped short of attempting to handle. For number (1), Snakemake would make a create a North American subsample and then an empty Iceland subsample, I think. So yeah, it’s not ideal and worth thinking/talking about more.

Richard Neher  22 hours ago
I think one aspect to think about here is whether we want to run analysis very a large number of places, in which case this hierarchy makes sense. But if this is mostly providing a structure that others use for their specific needs, a flat structure might be easier. Who know what kind of structure they might want to....

John Huddleston  16 hours ago
For the forecasting builds, I got around these types of problems by creating named builds with names of my choosing, nesting all outputs by build name within a single results directory, and storing all related configuration settings in the config file indexed by build name. One example is for a build based on simulated data here:
https://github.com/blab/flu-forecasting/blob/399fc145bb96d8f67364f7afae36e6783831ad51/config/config.json#L154-L178

**Comment by rneher** [comment]
I pushed this a little further along... not necessarily solving the issues above but I made it work for the typical case.

 * there is now a script that creates a json of the form
```
{
'region': {'north-america':{region:'North America'}, ....},
'country:{'georgia':{country:'Georgia', region:'Europe'}, ...},
'division':{'georgia':{'country':'USA', ...},
}
```
this is used in the subsample rule to population `--exclude-where=region={region}`

**Comment by huddlej** [comment]
## Handling awkward filtering interface
[On Slack, Richard says](https://bedfordlab.slack.com/archives/CSKMU6YUC/p1589141982374300):

> this custom sampling is a little annoying due to the way our --include-where statement works. This is applied after subsampling and hence works more as a force include than a pattern for subsampling (this was the intention…). The --exclude-where excludes a sequence that matches any of the criteria. To include sequences from a bunch of places, I therefore created multiple subsamples that get merged:

There is a [PR with a prototype to support more flexible boolean queries of the metadata](https://github.com/nextstrain/augur/pull/455). This PR adds a `--where` flag that takes [a pandas data frame query](https://pandas.pydata.org/pandas-docs/stable/user_guide/indexing.html#the-query-method) and filters the metadata data frame to match the query before any other filters are processed. This implementation has a couple quirks that we may want to resolve before merging (described in detail by the PR’s author), but these could be quickly addressed.

With the `--where` flag support, you can use some simple standard string comparisons like the `in` operator as follows (e.g., to get strains from New York or New Jersey):

```
augur filter \
  --sequences data/sequences.fasta \
  --metadata data/metadata.tsv \
  --where "division in ['New York', 'New Jersey']" \
  --output ny_nj.fasta
```

We could modify the subsampling rule to accept input for this argument and update the config like so:

```yaml
division_usa-north-east:
  # Focal samples for division
  division:
    seq_per_group: 100
    # The "northeast" is defined by these states.
    where: "division in ['New York', 'New Jersey', 'Connecticut', 'Massachusetts', 'Vermont', 'Maine', 'New Hampshire', 'Rhode Island']"

  # Contextual samples from division's region
  region:
    seq_per_group: 20
    where: "region == 'North America'"
    priorities:
      type: "proximity"
      focus: "division"

  # Contextual samples from the rest of the world, excluding
  # the current division to avoid resampling.
  global:
    group_by: "country year month"
    seq_per_group: 10
    where: "region != 'North America'"
    priorities:
      type: "proximity"
      focus: "division"
```

You can also use the `--where` argument to build more complex boolean queries of the metadata like this to explicitly disambiguate a division name:

```
augur filter \
  --sequences data/sequences.fasta \
  --metadata data/metadata.tsv \
  --where "(division == 'Limburg') & (country == 'Belgium')" \
  --output belgium_limburg.fasta
```

## Supporting custom geo builds
The `usa-north-east` build is implemented as a “division”, but it isn’t really a division. This is more of a custom build without a fixed geographic scale.

Could we call this something else to clearly distinguish it from the “division” builds for places like Washington State, etc.? An alternate is the “named build” approach where there is no location type field in the wildcards and there is only a build name. That name could be “division_washington” or it could be “usa-north-east” and that single name is used to lookup the wildcard information. If people want to use the default subsampling, they would have to name their builds with the correct prefix (`region_`, `country_`, or `division_`).

## Supporting keywords in subsampling config
Implementing the geo hierarchy as a config overlay is really clever! This seems like a config we’d want to hardcode into the Snakefile, so it is always loaded for people. This would avoid issues where the filter step (quietly) produces unexpected results because `{{region}}` doesn’t evaluate to anything without the config file.

huddlej:
Also, to help people visualize what is going on with each of these region, country, or division builds that have different subsampling configurations, here is a DAG view of a single build for each of these geographic scales with the corresponding subsampling rules of differing complexity. This is generated by the following command:

```bash
snakemake \
  auspice/ncov_region_north-america.json \
  auspice/ncov_country_usa.json \
  auspice/ncov_division_new-york.json \
  --configfile config/config.yaml config/geo_hierarchy.json  \
  --dag | dot -Tpdf > multi-region-dag.pdf
```


image

rneher:
I am in favor of the `--where` command. 

are we happy with the `priorities` arguments? right now there is just one type, but adding a type for future proofing seemed sensible. 

I am not sure what the adjust_meta_data should do. We could abstract this as well as in 
```
python adjust_meta_data.py --m m.tsv --where "(region!='North Amercia') --action "('division -> country')" -o out.tsv
```

rneher:
`adjust_meta_data` is still not ideal. But I am not sure what the solution is we want. I am ok with keeping full metadata for the time being.

rneher:
Thanks, John. I agree that we first need to sort out the build naming. A single name is more flexible. The reason we opted for the hierarchical approach was the more generic spec of subsampling. But we could have smth like

```
analyses:
  global:
    subsampling: "global"
    traits: "global"
  asia:
    subsampling: "region"
    traits: "region"
  iceland:
    subsampling: "country"
    traits: "country"
  my-own-crazy-run:
    subsampling: "very-special" 
```
basically link them with keys. This would allow defaults for larger categories and specific ones.

huddlej:
That's almost exactly what I was thinking. We could have a `location_scale` or `geo_scale` attribute per analysis, too, that would account for the current identification as a region, country, or division analysis. This could also allow us to continue indexing subsamping and traits settings by name of the analysis. For example:

```yaml
analyses:
  - global
      geo-scale: region
  - north-america
      geo-scale: region
  # Optional redundant encoding of country for clearer output in results directory
  # and to support auspice file names that map to a URL hierarchy like `/country/usa/`.
  - country_usa 
      geo-scale: country
  - division_washington
      geo-scale: division
  - county_king
      geo-scale: location
```

The `geo-scale` key would be used to look up default subsampling and trait information (in the same way `location_type` is being used right now) while the analysis name would be used to find analysis-specific rules. Users could define their own new geographic scale values and subsampling defaults for those scales, too. Then they could reuse those subsampling rules by setting the correct geo scale.

Maybe it is better to decouple the analyses names from the subsampling and trait rules like you've done here though. I'm not sure what will be clearer to users...

huddlej:
In chatting about this more, we've decided to implement the following changes for this PR and leave the remaining changes for later PRs:

1. Replace the current location_type and location_name wildcards with a single `build_name` wildcard and index all build-specific configurations by this name in the config file. We will try one of the two approaches outlined above for this config file structure with a slight preference for Richard's suggestion. Output files will still be stored in a subdirectory structure. For example, the European build's raw tree will be stored in `results/europe/tree_raw.nwk`. Users can choose any names they want to help organize their output in subdirectories and also for display by auspice. For example, the build name `country_usa` will have its output in `results/country_usa/` and will expand in the auspice view to a URL like `/ncov/country/usa`.

2. Implement a generic `get_params` function that can be used in the `params` block of rules and intelligently lookup default values for specific wildcard keys while also allowing for lookup of build-specific values that users have defined. An example interface for this function would look like:
  ```
  get_params(['subsampling', 'group-by'], key)
  ```
This could be implemented by generalizing the existing `_get_specific_subsampling_setting` function. We will try using this function only for the subsampling and traits initially in case we want to change the function later. However, the idea is that this function could be used in the long-term to allow users to override any build parameter we've exposed in the config file like the auspice config, colors files, etc.

3. Rename `location` to `geography` in the config and Snakemake files to distinguish the geographic scale from the actual `location` column in the metadata.

batson:
Strong agree with a single, wildcard `build_name`. We may have many builds at different scales, and do not want to force a logic going from the details of the build to the name.

batson:
In our current flow, we include:

* Group 1: All samples satisfying criterion X (eg, in CA, or from a certain DPH partner)
* Group 2: The closest samples to group 1 in terms of genetic distance
* Group 3: Contextual samples which could be specified using the proposed subsampling rules, especially with `--where` implemented.

We currently implement this by adding Group 1 and Group 2 to `include.txt`. With `--where`, Group 1 and Group 3 would be taken care of completely. Another strategy might be required to address Group 2.

Put another way, we are trying to add the genetic context of the Focal sequences, as well as the geographic context which can be specified using `division`, `location`, or the flexible `--where`.

huddlej:
Thank you! This is really helpful, @batson. It does seem like Group 2 can't be accounted for by our current plan. We'll have to discuss how to handle this best within the current augur toolkit.

In the short-term, we will plan to make the single-name build changes described here and also merge the `--where` feature for augur to at least cover your Groups 1 and 3.

huddlej:
@batson I forgot that we actually do already support subsampling for closely related sequences to a focal set thanks to @emmahodcroft's `scripts/priorities.py` script. This approach ranks sequences in the complete alignment based on their pairwise distance to a given alignment of focal sequences. I made [an example subsampling entry to represent the three groups you described above](https://github.com/nextstrain/ncov/blob/48f810cc28cdb1b52f9b9602a59ef3e6743811ea/config/config.yaml#L223-L242) which I'll re-paste here for the conversation:

```yaml
  # Default subsampling logic for locations
  location:
    # Focal samples for location
    focal:
      group_by: "year month"
      seq_per_group: 300
      exclude: "--exclude-where 'region!={{region}}' 'country!={{country}}' 'division!={{division}}' 'location!={{location}}'"

    # Samples that are genetically related to the focal samples
    related:
      group_by: "country year month"
      seq_per_group: 20
      exclude: "--exclude-where 'location={{location}}'"
      priorities:
        type: "proximity"
        focus: "focal"

    # Contextual samples from the rest of the world
    contextual:
      group_by: "country year month"
      seq_per_group: 10
      exclude: "--exclude-where 'location={{location}}'"
```

The `priorities` entry for the `related` subsampling set is the bit that links that rule to the `focal` set above it. The implementation here will take the top 20 closest sequences to the focal set per country, year, and month, but you could tune this to meet your needs.

rneher:
The offending output name is the `tip-frequencies`.  And this is admittedly a hole we dug ourselves. We use `ncov_europe_switzerland.json` to parse this into an URL `ncov/europe/switzerland` but auspice treats `ncov_europe_switzerland_tip-frequencies.json` differently. This seems very difficult to change right now...

rneher:
so, if we treat tip frequencies differently, we can use underscore wild cards that are naturally served as `/ncov/xyz/myplace/` by auspice. the way I did this by changing the tip-frequency rule to produce
`results/{build_name}/tip-frequencies.json` instead of writing straight to `auspice/`. I think this is uncontroversial and in many ways better than before. 

Then, I tied the copying to `auspice/ncov_{build_name}_tip-frequencies.json` to the generation of `auspice/ncov_{build_name}.json` without it being an explicit output file:

https://github.com/nextstrain/ncov/blob/generic-subsampling/rules/builds.smk#L668

This has of course various potential pit-falls. But having output files 

`auspice/ncov_{build_name}_tip-frequencies.json`
`auspice/ncov_{build_name}.json`

while allowing for underscores in the wildcards is ambiguous. But we want underscores for auspice....

rneher:
I also added a `swiss` profile just to explore how this could be customized:
https://github.com/nextstrain/ncov/blob/generic-subsampling/profiles/swiss/builds.yaml

rules/builds.smk
```diff
@@ -298,15 +294,15 @@ rule adjust_metadata_regions:
...
+        subsampling_scheme = lambda wildcards: _get_subsampling_scheme_by_build_name(wildcards.build_name),
         geographic_name = lambda wildcards: _get_geographic_name_by_build_name(wildcards.build_name)
     log:
         "logs/adjust_metadata_regions_{build_name}.txt"
     conda: config["conda_environment"]
     shell:
         """
         python3 scripts/adjust_regional_meta.py \
-            --{params.geographic_scale} {params.geographic_name:q} \
+            --{params.subsampling_scheme} {params.geographic_name:q} \
```
huddlej, line 305: This needs to be geographic scale here, I think. Otherwise the adjust metadata script won't know what to do with subsampling schemes like "switzerland".

huddlej:
@rneher Thank you for fixing the underscore issue! I like your solution here and agree it is worth the benefits of getting the underscore naming of auspice files that we want.

The build files are also clearer for having the separate subsampling scheme key. I noticed one place (inline commented above this) where we probably still need to use geographic scale instead of the subsampling scheme.

Do you think the example `location` subsampling scheme I proposed above would meet the BioHub team's needs as described by Josh's three major sequence groups? I could imagine tuning either priorities or filter in some way to build a set of the _n_ closest genetic samples to each focal sample.

rules/builds.smk
rneher: you are right!

rneher:
I hope this fixed it and didn't break anything. but I need to run now...

tsibley:
FWIW, if you want an alternate option for the "tip-frequencies" ambiguity problem, I believe it could also be solved by using a negative lookahead assertion in the `build_name` wildcard pattern:

```
# Define patterns we expect for wildcards.
wildcard_constraints:
    build_name = r'(?:[\w-](?!tip-frequencies))+',
```

That pattern matches a succession of word (`\w`) and `-` characters which aren't followed by the literal string `tip-frequencies`. The match stops when it encounters the first non-word or hyphen character (like the `.` in the extension) or a character (like the `_`) which immediately precedes `tip-frequencies`.

rneher:
Tom's wildcard suggestion seems to run fine. I'd suggest we leave the metadata adjustment for another PR.

huddlej:
Testing now and through tonight on Fred Hutch's cluster. Agree we should merge as soon as it is working on our cluster and also on an AWS test run.

huddlej:
I ran into two issues while testing tonight but ran out of time to implement good fixes:

1. The default build when just running `snakemake` doesn't work because all `builds` are commented out in `config/config.yaml`. My guess is that these need to be commented out or everyone will end up with a `global` build when the configs get deep-merged. My short-term solution was to uncomment the `global` build entry anyway, since the workflow should produce some meaningful output when run with `snakemake`. I wonder whether a flag for whether a build is active or not might be helpful for providing more control over which builds in the config are run by default or not.

2. The `all_regions` rule of the standard Nextstrain build failed to even build the DAG due to an error with the date wildcard in the dated auspice JSONs. For example, when I ran this command:

```bash
snakemake --profile profiles/nextstrain all_regions
```

I got this error:

```
KeyError: 'global_2020-05-17'
```

We might need more constraints on the wildcard regex for `build_name` to fix this.

rules/builds.smk
```diff
@@ -141,43 +141,81 @@ rule mask:
...
+    if "priorities" in subsampling_settings and subsampling_settings["priorities"]["type"] == "proximity":
+        return "--priority " + get_priorities(wildcards)
     else:
-        return f"--exclude-where \"region!={wildcards.region.replace('-', ' ')}\""
+        return ""
+
+
+def _get_specific_subsampling_setting(setting, optional=False):
+    def _get_setting(wildcards):
+        subsampling_scheme = _get_subsampling_scheme_by_build_name(wildcards.build_name)
```
jameshadfield, line 178: `subsampling_scheme` isn't used here (as far as I can tell)

rules/builds.smk
```diff
@@ -508,9 +508,9 @@ rule colors:
         color_schemes = config["files"]["color_schemes"],
         metadata = _get_metadata_by_wildcards
     output:
-        colors = "config/colors_{region}.tsv"
+        colors = "config/colors_{build_name}.tsv"
```
jameshadfield, line 511: I think this is the right time to change this output to `"results/{build_name}/colors.tsv"`

rules/builds.smk
```diff
@@ -141,43 +141,81 @@ rule mask:
...
+        if isinstance(value, str):
+            value = value.format(**wildcards)
+        else:
+            return value
+
+        if ("geo_hierarchy" in config
+            and geographic_scale in config["geo_hierarchy"]
+            and geographic_name in config["geo_hierarchy"][geographic_scale]):
+            return value.format(**config["geo_hierarchy"][geographic_scale][geographic_name])
+        else:
```
jameshadfield, line 196: To prevent confusing errors when the geo hierarchy doesn't exist either because it wasn't provided or because the value doesn't exist in it, we should add a check here along the lines of
```
elif re.search(r'\{.+\}', value):
    raise Exception(f"The parameters for the subsampling (sub)scheme {wildcards.subsample} require expansion from the geo hierarchy but it is not defined. Check that the build params define the correct geographic name and scale, and that these are present in the `geo_hierarchy`")
```
I don't think there's ever a valid argument to `augur filter` which contains curly braces, but perhaps?!?

jameshadfield: P.S. the entire if/else block could be replaced with
```
return value.format(**config.get("geo_hierarchy", {}).get(geographic_scale, {}).get(geographic_name, {}))
```
But I'm in favor of keeping it verbose and adding comments & checks

rules/builds.smk
```diff
@@ -141,43 +141,81 @@ rule mask:
...
     output:
-        sequences = REGION_PATH + "subsample_focus.fasta"
-    log:
-        "logs/subsample_focus_{region}.txt"
+        sequences = "results/{build_name}/sample-{subsample}.fasta"
     params:
-        group_by = _get_group_by_wildcards,
-        sequences_per_group = _get_sequences_per_group_by_wildcards,
-        exclude_argument = _get_focus_exclude_argument_by_wildcards
+        group_by = _get_specific_subsampling_setting("group_by"),
```
jameshadfield, line 214: Perhaps for another PR, but can we reformulate `_get_specific_subsampling_setting` to return "--group_by X Y Z" so that we can allow a subsampling (sub)scheme which doesn't define a "group_by" (in this case the empty string would be returned and)

scripts/priorities.py
```diff
@@ -8,14 +8,31 @@
...
+            sequence_groups[tuple(zip(ind[1], vals[0]))].append(s)
+        else:
+            sequence_groups[tuple()].append(s)
+
+    return sequence_groups
+
 INITIALISATION_LENGTH = 1000000
 
+def sequence_to_int_array(s, fill_value=110):
+    seq = np.frombuffer(s.lower().encode('utf-8'), dtype=np.int8).copy()
```
jameshadfield, line 30: This failed for me & I don't have time to debug currently:
```
  File "scripts/priorities.py", line 142, in <module>
    ref = sequence_to_int_array(SeqIO.read(args.reference, 'genbank').seq)
  File "scripts/priorities.py", line 31, in sequence_to_int_array
    seq = np.frombuffer(s.lower().encode('utf-8'), dtype=np.int8).copy()
AttributeError: 'Seq' object has no attribute 'encode'
```

jameshadfield:
I looked at this PR for the first time today, so many of my comments may be similar to those experienced by a "new" user.

* Because `config["builds"]` isn't defined for the default config, nothing works out-of-the-box which is disconcerting (John noted this in the previous comment also).
* The docs (e.g. `running.md`) need to be updated or rewritten. I completely understand that this might not have been done since this PR is still in flux, but needs to be done prior to merge.
* It took me a _long_ time to work out how the nCoV builds work in this PR -- the interplay of builds, subsampling schemes and geographic hierarchies. It was worth it, but it was a steep learning curve. Here are some notes which may be useful for the docs:

| name | description |
| ------ | ---------- |
| `build_name` | the main wildcard. Either the keys of `config["builds"]` or inferred from the build target supplied on the command line. Might be a mix of the two. Output is in `./results/${build_name}` |
| `subsampling_scheme` | the string value of  `config->builds->build_name->subsampling_scheme`, or the `build_name` if this doesn't exist |
| `subsampling_settings` | A number of dataset subsampling approaches, which together will make up the build. A dict of dicts at `config->"subsampling"->subsampling_scheme`. Each value is a dict defining a different sub-sampled alignment and so has `group_by`, `seq_per_group`, `exclude` args etc which are passed to `augur filter`.  |
| ??? | What's the name we should use for each dict in `subsampling_settings`? |
| a specific subsampling setting | The value is The names of these settings are the keys of the dicts in previous row (`group_by` etc) _BUT_ their values are evaluated using python's string formatting using `config["geo_hierarchy"][geographic_scale][geographic_name]` if it exists |
| `geographic_scale` | A build-specific setting. Default: `build_name`. Used to lookup values in the geo_heirachy. |
| `geographic_name` | A build-specific setting. Default: `build_name`. Used to lookup values in the geo_heirachy. |
| `sampling_trait` | `config->"exposure"->build_name->"trait"`. Defaults to "country". Used to modify the metadata. |
| `exposure_trait` | `config->"exposure"->build_name->"exposure"`. Defaults to "country_exposure". Used to modify the metadata. |

So each build (`build_name`) specifies a number of parameters, including which subsampling scheme to use. A single subsampling scheme can consist of multiple subsampled sets of data -- e.g. a focal set and a proximal set. The nature of how we get each of these subsampled sets of data is evaluated using build-specific settings. 

---

All in all this PR looks like a good way to run a huge number of variable geographically-subsampled runs. It's a steep learning curve, but having a number of examples (e.g. the swiss ones) really helps.

I think the next big steps (post this PR) will be to generate / modify the auspice config via the parameters of the build (`config->builds->build_name`) and to make the `incorporate_travel_history` rule dependent on `_get_trait_columns_by_wildcards`.

huddlej:
Thank you so much for digging through this @jameshadfield and writing up those variable docs! I've felt this getting more complicated as we've progressed. After we work through the issues you've raised here, we need to think about how to present these changes more clearly (as documentation, reorganization of the config file, etc.).

jameshadfield:
One thing to consider -- currently the `geo_hierarchies` serves only to simplify each build definition, but adds an extra layer of complexity. You could consider ditching it completely and adding a `config->"builds"->build_name->"subsampling_specifiers"` which has the information that would be currently found in `config["geo_hierarchy"][geographic_scale][geographic_name]`. More verbose, but also places the configuration specifics more obviously on each build definition.

rules/builds.smk
rneher: you are right. I deleted this.

rules/builds.smk
rneher: done

scripts/priorities.py
rneher: this might be due to Biopython version. It works for me. I added an explicit string cast.

rneher: yes, this was added recently: https://github.com/biopython/biopython/blame/master/Bio/Seq.py#L900

rneher:
I addressed a number of issues James spotted. My proposal for the default builds is to define them in the snakefile if and only if there are none defined in `config`. Being stuck with a huge global build just because it is in the default config is annoying. 

I agree with the comments about the geo_hierarchy. this was a bit of short-cut when we first implemented it, but it works pretty ok. 

we are facing the issue that we are trying to accommodate all possible cases which blows up the config or snakefile and makes it intransparent. 

Since there are concurrent plans to reorganize the subsampling/filtering, I don't think we should aim for perfection here. that said, the docs clearly need to be updated.

rneher:
I added a few things to `running.md` and moved more build specific things into the profiles.

rneher:
I added some more documentation. but most of the remaining issues are not issues of this PR. E.g. running example data doesn't work anymore bc we deleted the metadata.

rneher:
I pushed the output of the current config to staging with prefix gs:

https://nextstrain.org/staging/ncov/gs-global
https://nextstrain.org/staging/ncov/gs-africa

the config for ancestral reconstruction is different, but I don't think country level transmission are meaningful

rules/builds.smk
huddlej: This is a great suggestion. Several builds from the `swiss` profile were silently failing for me because of this issue. Just fixed it.

huddlej:
I really like @jameshadfield's suggestion of replacing the geo hierarchy config with more explicit fields in the config for each build. Once this information is defined in the config, it can also be reused elsewhere in the pipeline which could be helpful in the future.

I mocked up an alternative to the geo hierarchy file that is an intermediate solution to what James recommends. The idea is to require the `geographic_scale` and `geographic_name` fields in each build and then use the existing metadata file to query the other information we have available about the first record that matches the query defined by scale and name.

In the Snakefile this looks like:

```python
# Load metadata to get geographic information.
geographic_metadata_fields = ["region", "country", "division", "location"]
metadata = pd.read_table(config["metadata"])
geographic_metadata = metadata[geographic_metadata_fields].drop_duplicates()
```

Then in the subsampling scheme parsing function, this looks like:

```python
geographic_data = geographic_metadata.query(f"{geographic_scale} == '{geographic_name}'")

# If there are any records in the metadata matching the query, convert the first record to
# a dictionary and use this to format the subsampling string.
if geographic_data.shape[0] > 0:
    value = value.format(**geographic_data.iloc[0].to_dict())
```

What I liked about this approach is that it reuses existing geographic information from metadata without adding a new config file that has to be maintained separately. This approach also allows users to define a two config fields that can query the remaining geo information from metadata.

What I don't like about this approach now is that there will inevitably be redundant names at finer geographic scales and the approach of using the first record in the query here will fail for these cases. We could implement logic similar to the geo hierarchy script to use majority rule to resolve conflicts, but this also seems like an unsatisfying solution that limits the user's control over their own data.

I'm going to mockup James's recommendation and push it here as a single commit for y'all to review. If it ends up working, great. If not, we can revert it and delete it later on a rebase. :)

huddlej:
@rneher and @jameshadfield, check out the changes in this recent commit 27fc582, especially with respect to how the `builds.yaml` files now look. I personally think this approach makes the config easier to read and the subsampling rules simpler and more powerful.

rneher:
Thank you, John. I agree that not having a geo-hierarchy to maintain is desirable and the explicit definitions are not too bad. (but as predicted, the config files are picking up complexity). 

This works for me and I'd be happy to merge it. There are endless other things that we could do and discuss. but I think it is very hard to predict what exactly we need in the future. 

The main bit that I see as outstanding are the actual parameters. I'd like to keep builds at about 3k sequences. Geo reconstruction it the other set of parameters that might need tweaking.

huddlej:
Totally agree. We could tweak this forever. I'll resolve the conflicts with master and test the full Nextstrain builds on AWS to make sure we haven't broken anything for our own team.

huddlej:
@emmahodcroft Thank you for catching and fixing that issue with the metadata adjustment!

rneher:
I fixed a few things in the swiss build and added lat-longs to for Asia as region which were missing. I also wonder whether we should have the global build reconstruct region and region exposure instead of country. Regional builds should reconstruct country and region (which the exception of North America and Oceania that reconstruct division for lack of countries). But we could revert and keep this change for later.

rneher:
the default global view would then be
image

emmahodcroft:
I've done a full 'region build' on our cluster to test, and then try to 'finish' as if we were doing a 'real' push to live. After the changes above, the regional build finished without issue! 🎉  However, I've run into some problems when executing the 'finishing' commands that let one update the orderings & lat-longs. I'll work on this some more and hopefully fix - but wanted to flag this as a remaining issue that's important for keeping the builds flowing ;)

emmahodcroft:
There were a few issue with some color file paths but apart from this all seems to be good. If we make any more significant changes I'll do another test before we merge, but for the moment: region build runs as expected, and can be 'finished' with the commands in `nextstrain_exports.smk` as would be done for a 'real' push! 😄

huddlej:
@emmahodcroft I had those same colors path issues with my AWS build. Thank you for fixing them!

Also, the AWS build took 11.7 hours. We'll need to merge in Trevor's recent changes from master to subsampling and also lat_longs changes before we can merge this with master.

emmahodcroft:
11.7 hours!?! Oh dear. I did the full region build on our cluster and while I didn't time it, it certainly didn't take that long. @rneher is our sampling different from what would go to AWS? (Aka is this difference explained or is this something we should look into?)

rneher:
I updated the sampling and aligned the reconstruction with master. this finished on scicore in about 2h.

rneher:
the long run time on aws might be due to trees blocking paralell execution.

rneher:
the output of what I just merged is on staging.

huddlej:
Thank you for pushing this through, @rneher!

tsibley:
> Also, the AWS build took 11.7 hours.

Oof. This implies something is misconfigured there. No good reason it should take that long.

PR #391, opened 2020-05-05
Add log directives to most rules for better debugging

huddlej:
Updates rules to store standard output and errors from commands into log files
using a standard filename format [1]. Storing logs to file instead of writing them
out to the terminal should aid with debugging when errors occur on AWS Batch or
cluster jobs.

Note this does not log output from the final deploy to staging command or the Slack message commands. I wasn't sure whether it's nicer to have these messages in the AWS Batch output on attach or in a log file.

[1] https://snakemake.readthedocs.io/en/stable/snakefiles/rules.html#log-files

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 # Files created by the pipeline, which we want to keep out of git
 # (or at least out of _this_ git repo).
 benchmarks/
+logs/
 results/
 build/
 auspice/
--- a/docs/workflow_structure.md
+++ b/docs/workflow_structure.md
@@ -28,6 +28,7 @@ The following outlines the directory structure of the ncov workflow and describe
       - `experimental_features.yaml`: an putative example environment with experimental features that are not ready for production yet
   - `example_data/`: minimal data required to test out the default ncov build
   - `figures/`: images used in documentation and narratives
+  - `logs/`: log files containing standard output and error messages from commands run by most rules
   - `narratives/`: dated situation reports based on dated build JSONs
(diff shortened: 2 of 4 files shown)


emmahodcroft:
I won't have time to look at this in detail tonight - But having it in a file would definitely have been handy today (I ended up copying and pasting the entire terminal output 🙃 )! Thanks for working on this @huddlej !
And Tom's idea seems cool too - have it both ways!

tsibley:
@emmahodcroft Instead of copying and pasting from the terminal, you can also do this:

    nextstrain build --aws-batch --attach … &| tee build.log

emmahodcroft:
I will try that next time! :) (If this doesn't get implemented) I'm just not very good at remembering these things though... 😆

huddlej:
I like to eat cake! I'll update with `tee`.

huddlej:
I ended up using `2>&1 | tee {log}` since I have an old version of bash that doesn't support `|&` and I'm guessing not everyone else has upgrade their Mac to zsh...

tsibley:
:raised_hands: Thanks John!

I'd forgotten this was a newer Bash feature. Even on the latest macOS, the version of Bash is ancient, I believe because of licensing concerns. The default shell on the latest macOS switched to zsh, but I believe our shell snippets will still run in Bash because that's what Snakemake specifies. (And if that switched it would mean some of our rules' syntax is probably subtly wrong.)

huddlej:
This worked on a test run on AWS Batch, so merging now.

PR #393, opened 2020-05-06
Prevent workflow from trying to adjust global metadata

huddlej:
Previously, the `export_all_regions` rule tried to expand the adjusted metadata
path for all regions including the global build. The global build does not have
an adjusted metadata file, though; it uses the original metadata file instead.

This commit fixes the problem by introducing a more generic function to get the
correct metadata file conditional on the region and using this function as a
base for an input function (the original `_get_metadata_by_wildcards`) and in a
list comprehension in the `export_all_regions` rule inputs. These functions live
in a new Snakemake file for shared functions, `common.smk`, following the
guidelines recommended by the Snakemake authors [1].

[1] https://github.com/snakemake-workflows/dna-seq-gatk-variant-calling/blob/master/rules/common.smk

### Testing

I tested this code with the following steps:

1. attach a recent AWS Build
2. try to run the `export_all_regions` rule as is (without the changes in this PR) and confirm that the global build produces the original error when trying to adjust metadata.
3. apply changes in this PR to the attached build, re-run `clean_export_regions`, re-run `export_all_regions`, and confirm that the `check_missing_locations.py` script runs as expected and no errors occur

### DIFF
--- a/Snakefile
+++ b/Snakefile
@@ -41,6 +41,9 @@ rule clean:
     shell:
         "rm -rfv {params}"
 
+# Include small, shared functions that help build inputs and parameters.
+include: "rules/common.smk"
+
 # Include rules to handle primary build logic from multiple sequence alignment
 # to output of auspice JSONs for a default build.
 include: "rules/builds.smk"
--- a/rules/builds.smk
+++ b/rules/builds.smk
@@ -326,12 +326,6 @@ rule tree:
             --nthreads {threads} 2>&1 | tee {log}
         """
 
-def _get_metadata_by_wildcards(wildcards):
-    if wildcards.region == "global":
-        return rules.download.output.metadata
-    else:
-        return rules.adjust_metadata_regions.output.metadata
-
 rule refine:
     message:
         """
--- a/rules/common.smk
+++ b/rules/common.smk
@@ -1,0 +1,19 @@
+"""Small, shared functions used to generate inputs and parameters.
+"""
+def _get_metadata_by_region(region):
+    """Returns a path associated with the metadata for the given region.
+
+    The path can include wildcards that must be provided by the caller through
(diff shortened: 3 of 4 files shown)


huddlej:
Yes, I'm sorry I didn't include these to begin with. I think this covers everything I did to test:

```bash
# Clone repo.
git clone https://github.com/nextstrain/ncov.git
cd ncov

# Attach AWS Batch job with nextstrain CLI
# where <JOB_ID> matches your job you submitted earlier
# or a job reported in #ncov-gisaid-updates
nextstrain build . --aws-batch --attach <JOB_ID> .

# Run the normal commands for finalizing builds
# (https://github.com/nextstrain/ncov/blob/master/DEV_DOCS.md#finalizing-automated-builds)

# Touch all outputs to prevent Snakemake from trying to run
# the whole workflow again (since git-managed files will
# often have newer timestamps than the build outputs).
snakemake --touch --forceall --profile profiles/nextstrain

# Delete auspice files.
snakemake --profile profiles/nextstrain clean_export_regions

# Reexport auspice files.
# This command failed originally because the global metadata
# file didn't need to be adjusted.
snakemake --profile profiles/nextstrain export_all_regions

# Check out the new branch with the fix.
# You may need to run this first to undo any changes downloaded from AWS:
# git reset --hard master
git checkout dynamic-metadata

# Delete auspice files again.
snakemake --profile profiles/nextstrain clean_export_regions

# Try to export auspice files again. This time should work
# if the new branch actually fixes the issue.
snakemake --profile profiles/nextstrain export_all_regions
```

jameshadfield:
Thanks! (I thought you were attaching to a still-running instance, but "attach" in this case means downloading a completed run & then running things locally. Still really useful.)

PR #395, opened 2020-05-06
Keep gaps

rneher:
SARS-CoV-2 has meaningful deletions. We currently fill these with `N`s which makes them invisible in auspice.  This was motivated initially by spurious gaps and uncertainty about their veracity. but multiple deletions have been confirmed and observed repeatedly. This PR reinstantiates gaps. 

 * remove `--fill-gaps` flag from the alignment step
 * explicitly fill terminal gaps with `N`s in the mask step. 

### Testing
I'll post a link to staging once its done....

### DIFF
--- a/rules/builds.smk
+++ b/rules/builds.smk
@@ -89,8 +89,7 @@ rule align:
             --reference-sequence {input.reference} \
             --output {output.alignment} \
             --nthreads {threads} \
-            --remove-reference \
-            --fill-gaps 2>&1 | tee {log}
+            --remove-reference 2>&1 | tee {log}
         """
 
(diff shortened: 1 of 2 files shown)


rneher:
The staging build is now up.

This is a position with real gaps:
https://nextstrain.org/staging/ncov/global-w-gaps?c=gt-ORF8_44

This is a position near the end where the formerly abundant gaps are now gone:
https://nextstrain.org/staging/ncov/global-w-gaps?c=gt-nuc_29846

trvrb:
This looks great to me Richard. Thanks for implementing.

PR #402, opened 2020-05-12
Reconstruct division trait for North America and Oceania

trvrb:
North America and Oceania are not very interesting at the country scale. This switches trait reconstruction to be at the division level. At this point given how `modify-tree-according-to-exposure` works, it's non-trivial to reconstruct multiple traits (this should probably get attention at some point), but thought it better to reconstruct the more useful / interesting trait.

I've tested this with a condensed local build and things seemed to work alright.

This required two fixes to make it work:
1. Guarantee that `division_exposure` is actually present in metadata
2. Make sure that `division_exposure` is exported into the auspice JSON by including it as a coloring

(2) is non-ideal, but this is a persistent issue with how we handle metadata. There's no way for something to be present as an attr and not be a coloring.

### DIFF
--- a/config/auspice_config.json
+++ b/config/auspice_config.json
@@ -70,7 +70,12 @@
     },
     {
       "key": "country_exposure",
-      "title": "Exposure History",
+      "title": "Country of exposure",
+      "type": "categorical"
+    },
+    {
+      "key": "division_exposure",
+      "title": "Division of exposure",
       "type": "categorical"
     }
   ],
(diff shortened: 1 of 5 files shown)


trvrb:
Everything looks as it should:

Screen Shot 2020-05-12 at 9 21 28 AM

I'll merge this. Getting transmission line toggle in auspice (https://github.com/nextstrain/auspice/pull/1103) and default it to off would help with the overly noise (and often wrong) map transmission lines. But I think this is an improvement over existing behavior.

PR #404, opened 2020-05-12
Update mask sites

trvrb:
The original list of masked sites were chosen to mitigate against specific sequencing issues in early Wuhan samples. This is now longer the case. This PR removes these sites. Additionally, I walked through sites listed here http://virological.org/t/issues-with-sars-cov-2-sequencing-data/473, and could reproduce the noted homoplasies. However, I didn't see any that I was concerned were actually harming phylogenetic reconstruction (the homoplasies are scattered across the tree).

### DIFF
--- a/config/config.yaml
+++ b/config/config.yaml
@@ -59,12 +59,13 @@ partition_sequences:
 mask:
   # Number of bases to mask from the beginning and end of the alignment. These regions of the genome
   # are difficult to sequence accurately.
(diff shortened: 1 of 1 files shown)


rneher:
This is a much better list. We can keep 13408 on the radar for future addition.

emmahodcroft:
Thank you Trevor for doing this!

PR #409, opened 2020-05-18
produce an unsampled, non-temporal build

jameshadfield:
@huddlej @rneher and others -- here is the branch I'm using to produce non-subsampled builds. As you can see there are a few modifications i'm making to the snakemake file. Are there ways we can streamline this? I could think of a couple but am unsure the best practices here:
* A `config["refine"]["infer_temporal_structure"]` (bool) could be defined, and the snakemake rule `refine` calls `augur refine` with the appropriate parameters. 
* A `config["refine"]["traits"]["run"]` (bool) could be set and the `export` rule updated to dynamically use the appropriate node-data JSONs
* The change to `scripts/modify-tree-according-to-exposure.py` should probably be in master

Are there cleverer ways to accomplish this? (I'm also happy keeping a parallel branch, but it gets tedious...)

### DIFF
--- a/rules/builds.smk
+++ b/rules/builds.smk
@@ -316,13 +316,16 @@ rule tree:
         # Multiple sequence alignments can use up to 40 times their disk size in
         # memory, especially for larger alignments.
         # Note that Snakemake >5.10.0 supports input.size_mb to avoid converting from bytes to MB.
-        mem_mb=lambda wildcards, input: 40 * int(input.size / 1024 / 1024)
+        mem_mb=lambda wildcards, input: 40 * int(input.size / 1024 / 1024),
+    params:
+        mem_gb=lambda wildcards, resources: int(resources.mem_mb / 1024)-1,
     conda: config["conda_environment"]
     shell:
         """
         augur tree \
             --alignment {input.alignment} \
             --output {output.tree} \
+            --tree-builder-args="-mem {params.mem_gb}G" \
             --nthreads {threads} 2>&1 | tee {log}
         """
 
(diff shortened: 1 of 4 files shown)


rneher:
we could implement this in the context of the `generic-subsampling` branch. as a flag to the `config['builds']`

jameshadfield:
Will revisit this PR after the merge of #388

rules/builds.smk
```diff
@@ -311,18 +311,21 @@ rule tree:
...
+        mem_mb=lambda wildcards, input: 40 * int(input.size / 1024 / 1024),
+    params:
+        mem_gb=lambda wildcards, resources: int(resources.mem_mb / 1024)-1,
     conda: config["conda_environment"]
     shell:
         """
         augur tree \
             --alignment {input.alignment} \
             --output {output.tree} \
+            --tree-builder-args="-mem {params.mem_gb}G" \
```
jameshadfield, line 328: This doesn't work as expected (at least, not on AWS) where we get 
```
Job FAILED after 290.4 minutes (OutOfMemoryError: Container killed due to memory usage, exited 1)
```
Despite with iq-tree being run with `-mem 30G` and snakemake & AWS given 40Gb.

tsibley: Seems worth trying a node size with a much larger memory allocation, making sure the rule is benchmarked, and then seeing what is actually used to guide node size choice.

jameshadfield: Will do -- running now on a node with 64Gb (confusingly, `params.mem_gb` is 32, not sure why).

rneher:
impossible by now.

PR #418, opened 2020-05-28
Diagnostics

rneher:
This PR adds a script that scans aligned sequences for excess divergence, number of `N`s, and clusters of mutations. It is meant to facilitate spotting of outliers before the main run. The output are two files, one of which contains stats on sequence divergence, the other a list of strains flagged for exclusion along with a reason why:
```
Bangladesh/NIB_01/2020  too high divergence 252.81>20;snps clusters;
India/S1/2020   too high divergence 165.44>20;snps clusters;too many Ns (4822>3000)
USA/OH-OSUP0019/2020    too high divergence 111.62>20;snps clusters;
Australia/VIC1181/2020  too high divergence 105.86>20;snps clusters;
Wales/PHWC-3232A/2020   too high divergence 45.00>20;snps clusters;
Wales/PHWC-2BE6F/2020   too high divergence 29.93>20;snps clusters;
SaudiArabia/KAUST-Makkah155/2020        too high divergence 29.00>20;snps clusters;
England/20129172402/2020        too high divergence 28.30>20;snps clusters;
Australia/VIC877/2020   too high divergence 27.27>20;snps clusters;
SaudiArabia/KAUST-Makkah165/2020        too high divergence 26.00>20;snps clusters;
Australia/VIC909/2020   too high divergence 27.16>20;too many Ns (3737>3000)
Bangladesh/BARJ_CVASU_CTG_506/2020      too high divergence 23.01>20;snps clusters;
Georgia/Tb-6572/2020    too high divergence 23.63>20;snps clusters;
[...]
```

There are additional tests that could be added and the current ones could be refined.


emmahodcroft:
This is a great idea Richard! A couple of questions though that I can't quite puzzle out from the code. 

First, this doesn't automatically exclude sequences, right? This file would have to be looked at manually and the run stopped before it got too far to add any to exclude, is that correct?

Second, if the SNPs aren't clustered, but divergence is still high, what happens? How are the two balanced? In many cases the diversity is so high it doesn't really matter how it's happened. In other cases though, the diversity could be high but SNPs not clustered. Just curious if there's a clear cutoff there or something we need to feel our way along?

rneher:
yes, there is still some tweaking to be done. No automatic exclusion happens right now. Just flagging with reason. this could be either too high divergence OR clustered SNPs (5 in 50 bases) OR too many `N` (currently at 3000, more stringent than before).

one could either run this as a two step process

`snakemake results/flagged-sequences.tsv`

(should only take a few minutes) and then update `exclude.txt`

lmoncla:
This seems like a great feature that would be very helpful. Thanks for working on this.

rneher:
This now in addition produces a file called `results/to-exclude.txt` with content like this:
```
# clustered mutations
Spain/COV001096/2020
Switzerland/100807/2020
Sichuan/SC-NJ-091/2020
Sichuan/SC-NJ-093/2020
USA/MI-MDHHS-SC20378/2020
Chile/Santiago_53/2020
England/CAMB-1AD64E/2020
England/CAMB-1AD7C3/2020
England/CAMB-1AD927/2020
England/NORW-E7852/2020
England/CAMB-1AD9AF/2020
[...]

# divergence&clustered mutations
Spain/COV001096/2020
Switzerland/100807/2020
Sichuan/SC-NJ-091/2020
Sichuan/SC-NJ-093/2020
USA/MI-MDHHS-SC20378/2020
Chile/Santiago_53/2020
England/CAMB-1AD64E/2020
England/CAMB-1AD7C3/2020
England/CAMB-1AD927/2020
England/NORW-E7852/2020
England/CAMB-1AD9AF/2020
England/CAMB-1AD6F3/2020
England/CAMB-1AD259/2020
England/CAMB-1B0077/2020
[...]

# divergence
Spain/COV001096/2020
Switzerland/100807/2020
Sichuan/SC-NJ-091/2020
Sichuan/SC-NJ-093/2020
USA/MI-MDHHS-SC20378/2020
Chile/Santiago_53/2020
England/CAMB-1AD64E/2020
England/CAMB-1AD7C3/2020
England/CAMB-1AD927/2020
England/NORW-E7852/2020
[...]
```

rneher:
they are again sorted by submission date but grouped into categories by exclusion region.  This is meant for direct addition to the `exclude.txt`. The other files contain sequence specific reasons for exclusions.

rneher:
and the `results/flagged-sequences.tsv` is also sorted by reverse submission date.

rneher:
I adapted the criteria a bit and mask know clusters of mutations. I think this is now pretty sensitive and specific (but will probably need adjustment in a few months). 

I added the sequences flagged now to the exclude file.

rules/builds.smk
```diff
@@ -46,6 +46,85 @@ rule filter:
...
+    log:
+        "logs/excluded.txt"
+    conda: config["conda_environment"]
+    shell:
+        """
+        augur filter \
+            --sequences {input.sequences} \
+            --metadata {input.metadata} \
+            --include {input.include} \
+            --min-length 50000 \
```
huddlej, line 69: Can we delete this line? If the value is set so high as to disable the filter, we should be able to remove it.

config/config.yaml
```diff
@@ -40,15 +40,15 @@ files:
...
+  min_length: 27000
 
   # Omit sequences with incomplete date annotations.
   exclude_where: "date='2020' date='2020-01-XX' date='2020-02-XX' date='2020-03-XX' date='2020-04-XX' date='2020-05-XX' date='2020-06-XX' date='2020-01' date='2020-02' date='2020-03' date='2020-04' date='2020-05' date='2020-06'"
 
   # Select a subset of sequences per unique geographic and temporal group.
   # This limits the number of sequences in the base build to prevent long runtimes.
   group_by: "division year month"
-  sequences_per_group: 2000
+  sequences_per_group: 200000
```
huddlej, line 51: When setting the value this high, does this effectively disable the grouping part of the filter rule? Can we make the `group_by` and `sequences_per_group` config entries optional and then delete them here?

rules/builds.smk
huddlej: Sorry, I meant to highlight the line is `--min-length 50000 \`.

rneher: yes, just did. needed to check first what are required arguments...

config/config.yaml
rneher: deleted those as well

rneher:
Playing around with this a bit today, I realized how stochastic the exclusions are due to variation in how things get aligned. If many of the apparent exclusions are not due to true SNP clusters by the alignment issues, we might want to remove them from a particular run but not forever. 

So I just implemented an additional filter step where sequences that get flagged are excluded from downstream analysis for *this* run. Whether they should be added to the `exclude.txt` is another matter that given the vagaries of the alignment step might not be easy to determine automatically. The sequences with these alignment issues tend to be nanopore/primal scheme ones with lots of drop-outs. 

But this should be a good compromise between getting runs up that aren't contaminated by messed up alignments or bad sequences and not excluding too many sequences automatically.

rneher:
I added some docs: 
https://github.com/nextstrain/ncov/pull/418/files#diff-0815f7591ad0b36655edd4c94469bc0d

feel free to change the terminology of file names if anybody has strong feelings.

rules/builds.smk
rneher: this actually happens to be necessary. exclude everything and add the exclude file content.

PR #419, opened 2020-05-28
Exclude division=USA in `filter` for state-less USA Seqs

emmahodcroft:
Very simple PR - auto-filter sequences with `division=USA`.Since we fill all missing divisions with country, this will mean those with missing divisions in the USA will have `division=USA`.  

Currently, we exclude sequences with missing division (state) data manually by adding them to `exclude.txt`. 

However, there are a few shortcomings with this approach:
1. We must identify all those missing 'state' information, ideally as soon as they appear in the dataset (this is pretty good, but imperfect)
2. `exclude.txt` is by strain name - if strain name changes, the sequences are included again. We know this happens with GISAID data.
3. Later changes are usually never reflected. If a state is added later, the sequence stays excluded as it's somewhat unlikely we'll notice, check, and remove from `exclude.txt`. Similarly, if a state is removed later, we'd be similarly unlikely to notice and add it to exclude (less likely to happen).

With this method ,the sequences are automatically excluded, even if strain name changes, and will be automatically re-included if a state is ever added.

### Testing
Ran the `filter` step locally, all seemed to work without error. 45 sequences were excluded due to rule `division=USA`.


emmahodcroft:
I've run a few small builds with this and it's gotten rid of USA sequences which had division set as 'USA', which were not previously part of `exclude.txt` (I believe because strain names changed). I'm going to merge now.

PR #422, opened 2020-05-29
Allow build specific colours & config

emmahodcroft:
Currently one can specify a specific 'colors' files and 'auspice_config' file in your custom `builds.yaml` file, but only in a way that they apply to *all* the builds you're running.

This minor modification to `rules/builds.smk` means you can specify a 'colors' and 'auspice_config' file within a specific build, for example:

```
builds:
  texas:
  louisiana:
    subsampling_scheme: state
    geographic_scale: division
    region: North America
    country: USA
    division: Louisiana
    colors: "profiles/south-central/colors-Louisiana.tsv"
    auspice_config: "profiles/south-central/auspice_config_south-central_Louisiana.json"
```

### Important note
One *notable* about this PR is that it kinda changes the expected override behaviour in `augur`. By our own rules, command-line arguments override what's provided in the `config` file. That still happens here - but if you want a custom title, the easiest way I can see to provide this is within your `config` file, since otherwise you'd have to modify the `export_title` rule - and I don't think we want to encourage this.

So, I've set it so that if there's a supplied `auspice_config` file (either build specific or for all builds in the `builds.yaml` file), then the title is pulled from this and used.

An alternative route would be to allow users to set the 'title' as a field in `builds.yaml`. However, it was weird to me that setting it in the `auspice_config` file I was supplying was not changing what was exporting... But I recognise this is a little at odds with 'general' behaviour...

### Testing
Have tested on our cluster using both settings and seems to work. If you don't have a `title` field in the JSON you provide for `auspice_config` it'll revert to the defaults specified in `export_title()` in `rules/builds.smk`


emmahodcroft:
Thanks Trevor! I'm happy to surface commands like `title` instead. I took this approach without really thinking about it until I was writing up the PR, but didn't have time to adjust then. However, since then I've been thinking this is better.

However, I do think having the ability to pass in individual `auspice_config` files is good.  You don't have to - in the run I'm doing, I've gone for 'all' the runs, and then one specifically for Louisiana. Any that don't have a build-specific `auspice_config` default to whatever's in the overall `builds.yaml`. But, I think this is good to have this option - it means one can easily specify anything - filters, what's on display, etc, - without having to make a separate profile for each build you want. 

I'd be happy to also surface more 'command line' options as well, in future PRs, but this is a great way to allow people to specify everything/anything - if they want! But Don't have to!

emmahodcroft:
Can definitely work on this as well John, thanks!

emmahodcroft:
@huddlej Just as a note though - I actually think the `export_title` works pretty well!! It's automatically adjusts to be focused on whatever you've named your build, and I think it's a great flexible stand-in if you don't want to set your own title. 

I just see setting titles as something people probably _will_ want to do - but after doing all this, for my own Texas, Louisiana & South Central USA builds, I probably won't set the title for now!

(I am however, very interested in custom colours, custom display defaults, & custom maintainers!)

emmahodcroft:
I'm working on documentation, and it seems that `geographic_name` is effectively obsolite now? I've removed it from `running.md`, but please put back in if this is incorrect...

emmahodcroft:
Ok - I've updated the PR so that:
- `title` can now be set for all builds (if at 'base' level of the YAML) or at 'build' level (if within a build). Build-level overrides base-level. If nothing specified, defaults to `export_title` (which works pretty well!)
- `colors` can be set per build as well as existing functionality to set for all builds in a profile
- `auspice_config` can be set per build as well as existing functionality to set for all builds in a profile
- Docs are updated to reflect these changes, remove out-of-date config (`geographic_name`) options, and include the previously existing (but absent from docs) options for setting 'colors' and 'config' files for all the builds in a profile

I hope this is looking good now!

emmahodcroft:
If this is good, I'd like to merge before it becomes too behind master! I'll merge tomorrow if no more comments?

rules/builds.smk
```diff
@@ -693,6 +693,16 @@ def export_title(wildcards):
     # TODO: maybe we could replace this with a config entry for full/human-readable build name?
     location_name = wildcards.build_name
 
+    title = None
```
huddlej, line 696: If we `return` the title values from the config here instead of assigning them to a variable, we can delete the lines assigning title to `None` and checking for title downstream.

rules/builds.smk
```diff
@@ -693,6 +693,16 @@ def export_title(wildcards):
...
+    # If specified in config file generally, or in a config file build
+    if "title" in config["builds"][location_name]:
+        title = config["builds"][location_name]["title"]
+    elif "title" in config:
+        title = config["title"]
+    if title:
+            return title
```
huddlej, line 703: If we end up keeping this line, the indentation is off by one tab's worth.

docs/running.md
```diff
@@ -96,14 +96,24 @@ Instead, you can define your own analysis and tweak parameters by creating your
...
 file in this directory.
 This file specifies the analysis you want to run in a data structure called `builds`.
-Each entry has `build_name` which in the example is `global`.
+Each entry has `build_name` which in the example below are `switzerland`, `basel-stadt`, `ticino` and `lac-leman`.
 For each build, you can specify
 
  - `subsampling_scheme`: specifies how sequences are selected. Default schemes exist for `region`, `country`, and `division`. Custom schemes can be defined (see below).
- - `geographic_scale` and `geographic_name`: together, these two define the keys for subsampling. `geographic_name` defaults to `build_name` if not specified.
+ - `geographic_scale`: together, this defines the keys for subsampling.
  - `region`, `country`, and `division`: specify the location of the sample, down to the specified `geographic_scale`. If the build is a compilation of locations (like the `lac-leman` example below), you can provide your own final geographic location name.
```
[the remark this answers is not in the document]
emmahodcroft: Yes, thanks! Both good suggestions, I've integrated!

emmahodcroft:
Thank you John! I've made the changes requested, run a quick test of these last steps, and will merge now!

PR #428, opened 2020-06-08
Extra dedupes from lit search

nodrogluap:
Added a handful of exclusion based on literature review finding duplicate cases (annotations provided). Only added 13 lines at the end of the file, but git maybe doing something odd with the last blank line in master so please check merge.


emmahodcroft:
Thank you so much for your hard work on this @nodrogluap ! It's so much appreciated! To me these suggestions all look great. I've asked if someone else from the team can just double-confirm that all looks good, and we should be able to merge! Thank you again for your work!

trvrb:
Thanks so much for the help here @nodrogluap!

PR #434, opened 2020-06-14
Allow non-region 'focal' geographies

trvrb:

In the current behavior `adjust_regional_meta.py` takes a focal region, say "North America" and flattens country, division and location for all samples where region != "North America". 

This commit changes the flags in adjust_regional_meta to allow `--focal-resolution` and `--focal-label`. This allows focal resolution of "division" and focal label of "Washington". Doing this will additionally flatten county, division, location for samples from North America that don't belong to Washington. This allows for coloring of Washington locations that doesn't get swamped by locations from other regions in North America.

I believe this should be generally useful for geographically-focused builds.

### Testing

I've tested this with `snakemake --profile profiles/nextstrain` and `snakemake`. Things seem to be in order.

### DIFF
--- a/config/config.yaml
+++ b/config/config.yaml
@@ -65,6 +65,9 @@ mask:
   # 13402, 24389 and 24390 are restricted to Belgian samples
   mask_sites: "13402 24389 24390"
 
+adjust_metadata_regions:
+  focal_resolution: "region"
+
 tree:
   tree-builder-args: "'-ninit 10 -n 4'"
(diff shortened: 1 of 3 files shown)


emmahodcroft:
Thanks Trevor! Will have a look at this more closely!

trvrb:
I'm going to close this PR. I believe it was superseded by further work on `master` and is no longer relevant.

PR #437, opened 2020-06-16
WIP: Concatenate exclude txt

kairstenfay:
These commits were created during the ideation of #401. This branch will be discussed and revised at a future time, likely when addressing nextstrain/ncov-ingest/issues/45


huddlej:
I think it is safe to close this PR, since we've massively reworked our interactions with remote files in the workflow. @tsibley, you can reopen if you disagree, though.

PR #440, opened 2020-06-20
Subclades

rneher:
this adds four subclades to our list of major clades. There are named  19A/14805T etc and defined in a separate `config/subclades.tsv`. This file is concatenated with the major clade file and the resulting assignment is used as coloring. 

image


biocyberman:
Hi @rneher I am aiming similar to this, but in a more automated way.  I finished implementing the code, but still need to verify the results. It will be great if you could talk on Monday morning via Zoom, Slack or Skype so I can show you what I am doing?

rneher:
yes, happy to connect.

biocyberman:
@rneher Great! Sent you an email.

PR #443, opened 2020-06-22
Enable subsample filtering with augur filter's query argument

huddlej:
Adds support for a `query` key to individual subsampling definitions, allowing
users to filter strains by specifying the attribute values needed to retain
strains. This approach can make subsampling filters easier to read than the
corresponding exclusionary filters with `--exclude-where`.

Also adds an example of how to use the new `query` key in a county-level build
for King County in Washington State.

Note that the `--query` argument is only available in augur versions 8.0.0 and
later, so this PR also updates the required version of augur in the conda
environment file.

### DIFF
--- a/docs/customizing-analysis.md
+++ b/docs/customizing-analysis.md
@@ -113,9 +113,22 @@ lac-leman:
 
 All entries above canton level (the 'contextual' samples) specify priorities.
 Currently, we have only implemented one type of priority called `proximity`.
-It attempts to selected sequences as close as possible to the focal samples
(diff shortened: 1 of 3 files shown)


emmahodcroft:
I like this John! Thanks for adding this! We'll need to reconcile this with Sidney's changes to the docs and set-up - this should be merged first, but we'll also need to decide how to handle the `profiles` folder etc.

huddlej:
Thanks, @emmahodcroft! I'll update and merge this after #444 goes in.

PR #448, opened 2020-06-26
reuse colours if necessary

jameshadfield:
If a trait has more colours than defined in the colour schemes file (a) print a warning and (b) wrap around to reuse colours and thus allow the build to complete. While colour reuse isn't ideal we currently have >1000 locations to colour!

cc @lmoncla @trvrb @MoiraZuber @miparedes @emmahodcroft

### DIFF
--- a/scripts/assign-colors.py
+++ b/scripts/assign-colors.py
@@ -56,7 +56,19 @@ if __name__ == '__main__':
             if len(trait_array)==0:
                 print(f"No traits found for {trait_name}")
                 continue
-            color_array = schemes[len(trait_array)]
+            if len(schemes)<len(trait_array):
+              print(f"WARNING: insufficient colours available for trait {trait_name} - reusing colours!")
+              remain = len(trait_array)
+              color_array = []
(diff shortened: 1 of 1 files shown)


emmahodcroft:
Thank you James! This is a good solution, and maybe good enough after 1000 colors - nobody is ever going to be able to distinguish this by eye. 

I'm going to go ahead and merge this to keep the builds flowing, but I'll also tag @trvrb so we can discuss whether we want to keep this warning in and keep generating longer color lists, or whether we remove the warning here and just wrap after 1000 colors (I think this would be fine, myself)

lmoncla:
Thanks so much James! This seems really reasonable to me

trvrb:
Thanks @jameshadfield and @emmahodcroft! I think this is an excellent idea. We don't want things to break and if a trait has more than 1000 values, I don't think it's ordering is hugely important. I wasn't able to figure out what was breaking this. I'm counting 462 distinct values of `division` in `ncov/global/`.

PR #455, opened 2020-07-02
add script and new confilg files

MoiraZuber:
Introduction of a script that can be used to support dealing with new sequences, detect inconsistencies and clean up hierarchies stored in config/ordering.tsv. Currently, the script involves the following steps:
- Read in the current metadata in a hierarchical manner (region/country/division/location)
- Adjust special cases (e.g. Belgium, UK assigned as region, Turkey assigned to Asia instead of Europe etc.)
- Apply known misspellings (e.g. `Bucuresti -> Bucharest`)
- Correct known cases of false divisions (e.g. `division Hanoi -> location Hanoi with division Red River Delta`) and detect new cases by checking for inconsistencies in the metadata
- Correct known cases of duplicates (e.g. `Adams County -> Adams County WA`) and detect new cases
- Check for divisions and locations currently missing in the files config/ordering.tsv and config/lat_longs.tsv and suggest similarly spelled geographies (this should help detect misspellings)
- Compile a new ordering.tsv file which sorts all locations, divisions, countries and regions in a geographical meaningful manner (e.g. sorted by longitude), including comments providing a better overview and structuring.
- Provide as output all additions to the gisaid_annotations.tsv file that are required to correct all special cases covered by the script (e.g. `USA/WI-UW-373/2020	EPI_ISL_480355	location	Columbia County WI # previously Columbia County`)

### Testing
Application of this script to new instances of the metadata in order to detect exceptional cases. Potentially, adjustment of file paths to the new file structure is required.


emmahodcroft:
This should make a **big difference** in the build shepherd's prep - thanks Moira! I'll look over this in the next few days!

emmahodcroft:
I've just tried this out Moira - and it's fantastic! I've merged in master to make things correspond with the current `master` better - and corrected all the paths so this works.

Two small points that would be good to fix before we merge:
- I instinctively went to run this from the top-level `ncov` folder (by calling `python scripts/developer_scripts/parse_metadata.py` - but then the paths don't work! I'd suggest we maybe change the paths so that we assume the user is running from the top-level of `ncov` - that's what assumed for most other scripts. However, we can think about making this fancier in future :)
- When I copy/paste the annotation suggestions out of the command line, they don't have tabs, but instead spaces. This may be OS dependant - I didn't look into this much - but let's see if we can make them always be tabs!

I'll continue working on this, but I think we're already pretty close to being able to test & merge :)

Oh - one final point - we should maybe hide or 'put out of the way' the Cruise ship notifications - I'm unsure what to do about them. And I've left a couple of Belgium notifications because I'm just not familiar enough with Belgium - but perhaps I can chat with you and we can clear these up!

emmahodcroft:
Hey @MoiraZuber - I've just added to the script so that it now integrates using GeoPy to try and lat-long locations - give it a try! 

It should eventually print out a list of lat_longs into a file to be copied in, rather than just the screen - or perhaps we should just have it auto-insert them and print a new copy of the file, like what you do now with `color_orderings.tsv` ? That would save us having to alphabetise them by hand!

One thing - when I deleted some countries from the `lat_longs.tsv` to test it detecting 'new' countries for the GeoPy part, it doesn't seem to (even though this worked for locations and divisions). Am I missing how this works, or is something not working how it should?

emmahodcroft:
Script should now be run from top-level `ncov` directory - paths are set to expect this!

To run, ensure you are in `ncov` and use:
`python scripts/developer_scripts/parse_metadata.py`

emmahodcroft:
## **What the script does:**
Tries to highlight the problems with the metadata file you've just downloaded from the S3 bucket (`snakemake download -f --cores 1` after an ingest has completed on AWS). We can make it tell us about common problems by adding things to the files it draws on, which means it'll automatically correct these issues in future (or warn us about them). Also helps finding lat-long locations!

## **What this script won't do (completely):**
It won't tell you when a location/division/country shouldn't be what it is. It takes what GISAID lists as division and will work with that - just because you _can_ find a lat-long for that entry doesn't necessarily mean it _should_ be the division entry ('Dallas' for example - that's not only not a division (it's a location, division would be Texas), but it should be 'Dallas County' - more on how to automate that one below). 
Those changes you'll need to make manually via `gisaid_annotations.tsv` - push, re-ingest, and download a new `metadata.tsv`. 

## **How to run the script:**
`python scripts/developer_scripts/parse_metadata.py`
The script assumes you are running from the top-level `ncov` dir, and that you have `ncov-ingest` in a neighbouring folder ('sister' to `ncov` - to check `gisaid_annotations.tsv`). It'll assume `metadata.tsv` is in `data` and that the other files in `scripts/developer_scripts/` are present.

## **What it do in more detail:**
The script goes through and checks for a number of occurrences - the best way to detail these is probably just to run it!
It does some special stuff for Belgium which makes this much easier - but there are a few outstanding locations that need cleaning up - this still needs to be done. 

Apart from that, it'll auto-generate a list of things that should go into `gisaid_annotations.tsv` which will fix common problems - and you can modify the associated files to help it catch more! _(Note this list is given after the lat-long part below!)_

It'll also give a list of missing places that need to be in `color_orderings.tsv` and `lat-longs.tsv`. It'll then give an option to try to auto-generate lat-longs for these locations. _Ensure first that these are really what *should* be in division/location etc._ Fix using annotations first if needed!

However, let's go through a few files in `scripts/developer_scripts` to help show what they do in particular. We won't cover the Belgian stuff or the `accepted_exposure_additions.txt` & `sequences_exclude.txt` as they're still in flux at the moment.

### Files overview

- *`duplicates.txt`* & *`abbreviations.txt`* -  Some counties are in more than one state, so we need to append the state to them. An example is "Columbia County" which is in both Wisconsin and Oregon. If we add this county to `duplicates.txt` then the code will automatically generate a version that attaches the state abbreviation if it ever encounters this in future - to avoid us having an 'ambiguous' location without the state. To know what state abbreviation to add, it gets these from `abbreviations.txt`. It'll generate `annotations.tsv` lines!

- *`false_divisions.txt`* - If we frequently get sequences from somewhere that always lists the location as the division (ex: Hanoi - this is actually Red River Delta division) add it to this file so that it'll automatically be detected and a `annotations.tsv` line generated!

- *`variants.txt`* - Allows you to put in commonly used variants that we want to re-name. Then an `annotations.tsv` line will be generated automatically. Example: 'Kzn' becomes 'KwaZulu-Natal'!

- *`wrong_regions.txt`* - Places like "Turkey" and "Georgia" often get put by some into Europe and some into Asia. We can ensure they always end up in the right place by detecting where they *should* be and auto-generating `annotations.tsv` lines!

## **Output:**

At the moment most output is commad-line: 

- First, it's important to scroll up and see *all* the command-line info that's printed - it'll suggest some thing you should add to files & other warnings. Try to resolve these and then re-run the script to see if cleared/handled how you want!

- a list of `annotations.tsv` lines will be generated, which you can copy/paste into the `gisaid_annotations.tsv` file, and use to re-generate the metadata file. You may have to replace 4 spaces with tab - sorry - working on this!

- Once you do the lat-long generation, a list of the new lat-long values will be printed (or blank lines for those you need to look up manually). You can copy these into the lat-long file - and remove the lat-long to put into `color_orderings.tsv` if you like.

- `/scripts/developer_scripts/color_ordering.tsv` is generated, which is ordered in a new way - comments delimit countries & divisions & things are ordered within these by latitude. In future, we can switch to using this file entirely, so no more manual alphabetising!

**REMEMBER TO COPY ANY CHANGES TO LAT-LONGS & COLOR_ORDERINGS TO _MASTER_ OR THEY WON'T AFFECT THE BUILD!!**


## To-do:
- [ ] Make spaces into tabs for output on comand-line
- [ ] Make file print out lat-long info
- [ ] Clarify the Belgian files & how they work - & remaining seqs sticking around
- [ ] Clarify `exposure_additions` & `sequences_exclude`

emmahodcroft:
Turns out making it print tabs isn't really a thing, but the new annotation additions now print to a file as well, so they can be copied/pasted easily! Can do same for lat-long but train connection isn't good enough to use GeoPy at the moment...

scripts/developer_scripts/country_ordering/Belgium_variants.txt
```diff
@@ -0,0 +1,147 @@
...
+Graven	Grez-Doiceau
+Itter	Ittre
+Geldenaken	Jodoigne
+Terhulpen	La Hulpe
+Nijvel	Nivelles
+Perwijs	Perwez
+Tubeke	Tubize
+Waver	Wavre
+Aat	Ath
+'s Gravenbrakel	Braine-Le-Comte
```
kairstenfay, line 23: Just browsing through this PR and I think this line might have a typo? It starts with `'s`

kairstenfay: Nevermind I see this in other places, I'm just uncultured apparently :smile:

emmahodcroft: Haha I totally would have agreed with you and fixed it! Oh Belgium :)

emmahodcroft:
One downside of the script is that if you need to replace two things, it won't do both. For example, if both country and division are 'Bahrein' - it'll replace `division` to 'Bahrain' but one must manually generate the `gisaid_annotations.tsv` entries to also fix `country` - this isn't hard (literally copy paste and replace), but might be easily forgotten.

eharkins:
Notes from using the script (sorry if I missed somewhere where they are already addressed!):
- Would be good to add a list of dependencies this introduces e.g. `geopy`
- (only missing in ordering) locations are sometimes not added to `scripts/developer_scripts/color_ordering.tsv`, e.g.
```False division corrected: Keur Massar (true division: Dakar)
False division corrected: Yeumbeul (true division: Dakar)
False division corrected: Sangalkam (true division: Dakar)
False division corrected: Tuebingen (true division: Baden-Wuerttemberg)
=============================
(2 cruise ship entries ignored for duplicate locations)
(2 cruise ship entries ignored for missing state abbreviations)
=============================
Missing locations:
# Senegal #
Dakar
location        Keur Massar (only missing in ordering)
location        Sangalkam (only missing in ordering)
```
where Keur Massar and Sangalkam don't get added to `scripts/developer_scripts/color_ordering.tsv` maybe because they are being corrected? Not sure but can investigate more if necessary.
- Not sure how to handle when I'd like to add something to variants.txt like Gyeonggi -> Gyeonggi Province but Gyeonggi is a location and division and I only want to make this correction when it is set as a division?
- It was super useful that the script printed annotations for correcting unknown to blank! However when i had some with unknown division and location, it only printed annotations for locations
- It would be great to have a way to gather submitting authors after sequences have potentially been filtered out so as to not accidentally tweet thanking authors whose sequences were filtered out of the build. Not sure this belongs in this script, [more on this in slack](https://bedfordlab.slack.com/archives/CTZKJC7PZ/p1594428048299600)

emmahodcroft:
These are great notes @eharkins! And many of them similar to stuff I've noticed, too. I'm going to go ahead and merge this PR, and make an Issue that references all of these things, so we can start addressing them with PRs, and fixes.

emmahodcroft:
Oh, and just as a note on dependancies - I agree, but I wouldn't want to add this anywhere more formally in the `ncov` repo as it's not needed by the vast majority of people using it - just us. I don't know if there's a more standard way of handling this, though!

emmahodcroft:
Sorry one more note on the tweeting - I think this should perhaps be a separate script, as that would more likely be run at the end of the build, when the sequences have been filtered/quality assessed, etc - whereas this one is mostly run pre-build!

emmahodcroft:
Here's a link to the new issues: https://github.com/nextstrain/ncov/issues/467

eharkins:
> Oh, and just as a note on dependancies - I agree, but I wouldn't want to add this anywhere more formally in the ncov repo as it's not needed by the vast majority of people using it - just us. I don't know if there's a more standard way of handling this, though!

Since we are keeping things specific to the `scripts/developer_scripts` in that same directory, we could just add an environment file (pip, conda, or something like that) in there to track dependencies for those scripts. Not as big of a priority since as you said it's only for internal use.

emmahodcroft:
That sounds like a good solution to me!

PR #461, opened 2020-07-09
Revise tutorial quickstart and setup guides

### LINKED ISSUE #456: Replace the "quickstart" guide with an initial installation and quick test guide with a minimal viable output to visualize
**Current Behavior**

Starting on [the introduction page of the tutorial](https://nextstrain.github.io/ncov/index), users are unsure whether to start at the "quickstart" or at the first section of the tutorial. The "quickstart" actually includes critical steps for installation of prerequisites, so it may not appear optional at first glance.

Also, the current quickstart runs a full workflow with multiple builds and isn't really quick.

**Expected behavior**

Users should have a single clear entry point to the tutorial instead of multiple potentially conflicting entry points. If we do retain a "quickstart", it should run a very simple "toy" workflow as a sanity check before users proceed with the rest of the tutorial.

**Possible solution**

I would recommend the following steps to address this issue:

  - Remove the "quickstart" section from the index page
  - Move the "setup and installation" page to the beginning of the tutorial
  - Copy the quickstart installation instructions for conda + augur + auspice into the "setup and installation" page along with the link to alternate installation instructions. (This replaces the current installation section which currently doesn't link out to proper installation instructions.)
  - Add the quickstart's example of how to run the workflow with Snakemake to the "setup" page after the installation instructions
  - Define a new profile named "getting_started" that defines only a single global build using the original smaller set of "example data" and update the  `--profile my_profiles/example` argument from the Snakemake command in the quickstart to run this build. We will need to restore the original example sequences and metadata and also override the default parameters for these in the config file of the profile.

These steps should clarify the flow of the tutorial to users and also get them to the point of having software installed and a simple workflow tested before they embark on the more advanced bits of the tutorial.

**Additional context**

[This issue was originally raised by Stacia Wyman on the Nextstrain discussion board](https://discussion.nextstrain.org/t/feedback-on-tutorial/58).

huddlej:

This PR attempts to minimize the number of entry points for users to the tutorial content. This issue is fully described in #456. This PR moves much of the quickstart content into the setup/installation guide and removes the quickstart such that there is only a single entry point to the documentation for users.

This PR also reorders the setup/installation instructions to precede the "preparing data" section, so users have a working Nextstrain environment before they embark on the rest of the tutorial. The installation instructions are copied from the quickstart and along with the proper link to full installation instructions in the Nextstrain docs.

### Related issue(s)

Fixes #456 

### Testing

I have visually confirmed that edits appear as intended through GitHub's markdown rendered. Whether edits appear properly after Jekyll rendering has not been tested.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,9 +15,9 @@ If you were running customized builds before this, these changes may cause some 
 
 **We've written a comprehensive guide to get you up and running in <1 hr. Click on the below links to follow it. It covers:**
 
-0. [**Introduction and quickstart**](https://nextstrain.github.io/ncov/index) _(Start here)_
-1. [**Preparing your data**](https://nextstrain.github.io/ncov/data-prep)
-2. [**Setup and installation**](https://nextstrain.github.io/ncov/setup.md)
+0. [**Introduction**](https://nextstrain.github.io/ncov/index) _(Start here)_
+1. [**Setup and installation**](https://nextstrain.github.io/ncov/setup.md)
+2. [**Preparing your data**](https://nextstrain.github.io/ncov/data-prep)
 3. [**Orientation: analysis workflow**](https://nextstrain.github.io/ncov/orientation-workflow)
 4. [**Orientation: which files should I touch?**](https://nextstrain.github.io/ncov/orientation-files)
 5. [**Running & troubleshooting**](https://nextstrain.github.io/ncov/running)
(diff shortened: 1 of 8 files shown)


huddlej:
Brian O'Roak from OHSU points out that the running guide (`docs/running.md`) needs to mention that users have to modify the `config.yaml` file as well as the `builds.yaml`. Otherwise, the config will still refer to the original `example` profile directory.

The relevant sentence to modify is:

> To adapt this for your own analyses, copy my_profiles/example to my_profiles/<my-new-name> and open the builds.yaml file in this directory.

PR #465, opened 2020-07-12
Excluding Gujarati minor allele genomes, keep major allele only

nodrogluap:
In their preprint (https://www.biorxiv.org/content/10.1101/2020.07.10.197095v1) the Gujarat team manually produced a second, minor allele genome with 'b' suffix for any sample when they saw heterozygous mutations, under the assumption that co-infection was present. This assumption may overinflate representation of these individual samples in the tree, hence excluding.


emmahodcroft:
Thanks so much for spotting this and adding it @nodrogluap !!

PR #466, opened 2020-07-13
Update latest sitrep in footer (or replace entirely with /ncov-sit-reps

eharkins:
I happened to notice that the footer text on nextstrain.org (that comes from this repo `nextstrain_profiles/nextstrain/nextstrain_description{_zh}.md`) currently links to this sit rep:  https://nextstrain.org/narratives/ncov/sit-rep/2020-04-17.

I am assuming with this PR that we would be ok to replace that (in both the english and Chinese versions) with a link to https://nextstrain.org/ncov-sit-reps/ - so that we don't forget to update it again later when we do more sit reps (even though the 05-15 one is
>  our last weekly Situation Report for a while

). Otherwise, I can remove the second commit and go back to the first one on this branch which just updates each of them to https://nextstrain.org/narratives/ncov/sit-rep/2020-05-15.


emmahodcroft:
Good catch @eharkins ! I agree switching to this link is better!

eharkins:
@emmahodcroft any one in particular i should tag or request review from to merge?

emmahodcroft:
It's super quiet these days, I don't know who's around. I'd just merge it - it makes more sense than linking the old stuff, and we can easily change or update if anyone has another idea or big concerns about something!

Issue #467, opened 2020-07-14
Issues/Features for Build Shepherd `parse_metadata` Script


emmahodcroft:
These are issues that should be tackled in order to improve the use of the `parse_metadata.py` script used by the Nextstrain build shepherds. A mix of bugs and features.

## High priority 
- [ ] **1** More specific `variants.txt` corrections. 
Example 1 : 'Gyeonggi' is a location *and* a division. We want to correct `division Gyeonggi` to `division Gyeonggi Province` but _only_ when it's a division, not when it's a location.
Example 2: 'Bahrein' is both a country and a division - both need to be corrected to 'Bahrain'. Currently only the `division` annotations are automatically generated.
Example 3: Annotations are auto-generated to correct 'unknown' to ` ` (blank). However if one has both `unknown` divisions _and_ locations, only location corrections are generated. _(Emma note: I believe these should also only go to blank when the country is 'USA' as we don't do this for other countries - right? May need to be special case)_
- [ ] **2** Auto-add missing places to the auto-generated `color_ordering.tsv` file
Example: 'Sangalkam' is missing from the `defaults/color_ordering.tsv` and is displayed with a warning (`Sangalkam (only missing in ordering)`), but this is not added to the auto-generated `developer_scripts/color_ordering.tsv`). This needs to happen before we can switch to just using an auto-generated orderings file
Note: Ensure this happens both just if the place is missing from `color_ordering` *and* if it's a 'false' division that's been corrected. See example [here](https://github.com/nextstrain/ncov/pull/455#issuecomment-657661501)

## Medium priority
- [ ] **3** Automatically add new locations needing lat-longs into `defaults/lat_longs.tsv` and re-sort the file as necessary (in order by `location`, `division`, `country`, etc)
- [ ] **4** Clarify the role of `sequences_exclude.txt` - what does it do, how can we use it better?
- [ ] **5** Clarify role of `accepted_exposure_additions.txt` and figure out how to get working optimally/resolve inconsistencies 

## Low priority
- [ ] **6** Convey to people who might want to run this script that `geopy` is necessary dependency
- [ ] **7** Clear remaining `Belgium` alerts at start of script & outline best way to handle similar alerts in future
- [ ] **8** Better handle ignoring (but informing user about) cruise-related missing data (display more compactly/prettily - and just once)

emmahodcroft:
[...]

PR #468, opened 2020-07-14
Allow country names to also be corrected via `parse_metadata`

emmahodcroft:
Problem **1** in issue #467 includes a problem where changes specified in `variants.txt` were not applied to `country` but only to `location` and `division`.

The changes here should allow the script to generate `gisaid_annotations` fixes for `country` as well. Note this doesn't fix the problem where you can't specify a change at one level (say, division) but not at another (say country) if the names are identical - it'll change them both!

### Testing
You can test this branch by changing a few `Bahrain` country and division in `metadata.tsv` to `Bahrein` and then running the script. It should output `gisaid_annotations` corrections for both country and division (current `master` produces them only for `division`)

@MoiraZuber I'd appreciate if you can say if this looks ok to you!


eharkins:
haven't looked at the diff but it works for the one I needed to correct today!
```
Apply variant (country): United States -> USA                                                            │
```

emmahodcroft:
Thanks Moira! I'll try to get this merged now :)

PR #473, opened 2020-07-17
Use python3 executable instead of system executable

### LINKED ISSUE #442: Use of Python system executable prevents rules from accessing Python modules in Snakemake conda environments
## Description of the problem

[Discussion in a PR about how to reference Python executables ](https://github.com/nextstrain/ncov/pull/274) originally included a plan to standardize on `python3` instead of `python`. Continued discussion led to [an alternate solution](https://github.com/nextstrain/ncov/pull/412) where the system `python` would be used when it was defined or `python3` would be used. After additional testing, it turns out this solution prevents proper behavior of Snakemake’s conda environments with the `--use-conda`  flag.

When running ncov builds with `--use-conda` and the system executable, Snakemake activates the rule’s conda environment but the Python modules and executables installed in that environment are not available to the rule. This is because the `python` executable used in the shell command does not belong to the conda environment but to whatever environment the user executes Snakemake from.

For example, I want to build a subsampled alignment for a county-level ncov build like so:

```
snakemake \
  --use-conda \
  --profile profiles/king-county \
  results/usa_washington_king-county/subsampled_alignment.fasta
```

At the “combine and deduplicate” step of the workflow, snakemake activates the conda environment that has `nextstrain-augur` installed. The environment I’m calling Snakemake from does not have augur installed. The shell command for this rule references the system executable Python, so the command that gets executed is:

```
/Users/jlhudd/miniconda3/envs/nextstrain/bin/python scripts/combine-and-dedup-fastas.py \
  --input [...snip...] \
  --output [...snip...]
```

This execution produces the following error:

```
  File "scripts/combine-and-dedup-fastas.py", line 3, in <module>
    from augur.align import read_sequences
ModuleNotFoundError: No module named 'augur'
```

However, if I modify the shell command to use `python3` , as shown below, everything works as expected.

```
python3 scripts/combine-and-dedup-fastas.py \
  --input [...snip...] \
  --output [...snip...]
```

## Proposed solution

I propose that we revert back to running Python with the `python3` command.

One major reason originally given to not use this approach was that [Anaconda installations on Windows do not always symlink a `python3` executable](https://github.com/nextstrain/ncov/pull/274#issuecomment-622996487). Since we do not officially support augur on Windows anyway, this is does not seem to be a major issue. The inability to use `--use-conda` is a bigger issue, especially for those of us running builds in environments where we can’t use Docker (e.g., a cluster).

huddlej:
Allows Snakemake rules to use the most appropriate Python 3 executable currently available in the environment instead of preferring the system executable. This distinction is most important when users run Snakemake with the `--use-conda` flag and expect to have access to Python modules installed in the corresponding conda environment for each rule. When the system executable was preferred, it would be possible for the rule to run with a Python executable that did not have access to the same environment as the user's requested conda environment.

Minimal steps to recreate the original problem described in #442 and to test this PR:

```bash
# Create and activate a minimal Python 3 environment.
conda create -n ncov python=3.6
conda activate ncov

# Install Snakemake.
python3 -m pip install snakemake

# Run the "getting started" analysis with --use-conda.
snakemake --use-conda --profile my_profiles/getting_started/
```

This last step produces several errors when the system executable Python (from the outer `ncov` environment) does not have access to pandas, augur, etc. This issue is resolved by running all Python scripts with `python3`.

This PR also removes a dependency on TreeTime in the main workflow. This dependency causes an error in the minimal example above where the user will not necessarily have TreeTime installed in their current environment.

Resolves #442

### DIFF
--- a/Snakefile
+++ b/Snakefile
@@ -6,11 +6,6 @@ from snakemake.utils import validate
 configfile: "defaults/parameters.yaml"
 validate(config, schema="workflow/schemas/config.schema.yaml")
 
-# In our shell rules, we try to use the same Python executable running
-# Snakemake, if known, by interpolating {python:q}.  The aim is to prevent
-# issues with multiple Python versions or symlinks for python vs. python3.
-python = sys.executable or "python3"
-
 # default build if none specified in config
 if "builds" not in config:
     config["builds"] = {
--- a/workflow/snakemake_rules/export_for_nextstrain.smk
+++ b/workflow/snakemake_rules/export_for_nextstrain.smk
@@ -63,7 +63,7 @@ rule export_all_regions:
     conda: config["conda_environment"]
     shell:
         """
-        {python:q} ./scripts/check_missing_locations.py \
+        python3 ./scripts/check_missing_locations.py \
             --metadata {input.metadata} \
             --colors {input.colors} \
             --latlong {input.lat_longs}
@@ -158,7 +158,7 @@ rule incorporate_travel_history_gisaid:
     conda: config["conda_environment"]
     shell:
         """
-        {python:q} ./scripts/modify-tree-according-to-exposure.py \
+        python3 ./scripts/modify-tree-according-to-exposure.py \
             --input {input.auspice_json} \
             --colors {input.colors} \
             --lat-longs {input.lat_longs} \
@@ -183,7 +183,7 @@ rule incorporate_travel_history_zh:
     conda: config["conda_environment"]
     shell:
         """
-        {python:q} ./scripts/modify-tree-according-to-exposure.py \
+        python3 ./scripts/modify-tree-according-to-exposure.py \
             --input {input.auspice_json} \
             --colors {input.colors} \
             --lat-longs {input.lat_longs} \
@@ -203,7 +203,7 @@ rule fix_colorings_gisaid:
     conda: config["conda_environment"]
     shell:
         """
-        {python:q} scripts/fix-colorings.py \
+        python3 scripts/fix-colorings.py \
             --input {input.auspice_json} \
             --output {output.auspice_json} 2>&1 | tee {log}
         """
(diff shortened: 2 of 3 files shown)


huddlej:
@tsibley Thank you for looking this over! I am torn about the version availability issue. It's something I've thought a lot about since our ncov builds are tied to the version of augur in Docker and any new major release of augur has the potential to break things for everyone.

I would love to get to the place where our canonical mode of executing workflows is with `snakemake --use-conda`. This would allow build maintainers to explicitly opt in to a new version of augur (or any other tool, really) and test the effects of these version change(s) in separate branches of the workflow repo before merging to master. This would also allow us to safely push new features of Augur (like ISO dates) into people's workflows if they are pulling from this repo.

Part of this ideal world (in my mind) would include allowing users to specify `--use-conda` for workflows executed with the Nextstrain CLI, too (`nextstrain build . --use-conda`). This functionality would allow users to not only specify versions of Augur, but they could also easily install custom tools that are required for a given workflow but that do not need to be part of the core Docker image. An example use case here is the Zika USVI workflow that requires custom matplotlib extensions that are specific to Alli's project.

If this is a reasonable vision of the future, I think it requires revisiting the work I was doing pre-pandemic to use a different base Docker image for the CLI. This image would need to have conda installed for the `--use-conda` flag to work, so that means an increased default image size. I have some prototypes of a Dockerfile for this, but maybe we can work together to optimize these for a production-ready Dockerfile?

tsibley:
@huddlej Thanks for your thoughts on this; they're well-considered, as usual! I agree with the high-level goals of making our core builds more robust to software releases/updates/new features and allowing custom software to included. I think there are some good solutions using Conda and Docker (together or independently) that are worth exploring more. I'm particularly interested in making our software and workflow version management both a) easy to reason about for people who want/need to and b) fairly hidden from/transparent to anyone who doesn't. (Indeed, that's one of the goals of the CLI.) It'd be great to explore all this with you some more in conservation and prototyping.

huddlej:
I split out the TreeTime/ISO dates modification into a separate PR (#483), so merging this PR will change "one thing" and have less of an adverse effect on users.

PR #474, opened 2020-07-22
Reduce zero length branches into polytomies in non-timetree runs

rneher:
Enable non-time-tree trees without near zero branch length. Most treebuilders produce binary trees resulting in very many unsupported branches in SARS-CoV-2 whose length is a small value within the accuracy of the branch length optimization. Without timetree estimation in refine, these branches persist. 

This PR introduces a new rule sandwiched between `tree` and `refine` that reduces all these branches into polytomies.

### Testing
It shouldn't affect the builds, the normal refine step would do this reduction anyway.


huddlej (commented):
To make sure I understand the motivation here:

Normally, the timetree estimation of the  `refine` step would collapse zero-length branches like the code you’ve added here. But now users can opt out of this calculation when they disable `--timetree` in the config. This PR will ensure that the zero-length branches are always collapsed prior to `refine` no matter what the parameters for the `refine` step are. Is that correct?

The new code works for me when I run inside an existing conda environment, but the `run:` block in Snakemake doesn’t support conda environments with `--use-conda`. I can push a small commit that fixes this by putting the TreeTime code in its own script that gets called from a `shell:` block instead.

As an alternative to adding a new rule to the workflow, it looks like we can get very similar results by telling IQ-TREE to collapse short branches with the `-czb` flag passed through `--tree-builder-args`. This flag “[collapses zero-length branches](http://www.iqtree.org/doc/Command-Reference#miscellaneous-options)” based on a length threshold that [depends on the alignment length by default](https://github.com/Cibiv/IQ-TREE/blob/24bece92249d508af2fae103282ef274546bc41c/main/phyloanalysis.cpp#L2427-L2431).

I tried disabling `--timetree` and building auspice JSONs with the new remove-short-branches rule or with the `-czb` flag for IQ-TREE. Here are the trees side by side (using the smaller example dataset):

image

Here is a closer look at clade 20C:

image

For reference, here is the same pruned tree (left) compared to the tree I got without pruning short branches (right). This confirms that the branch pruning rule seems to work.

image

I also get that the [prune_short_branches](https://github.com/neherlab/treetime/blob/3b494be89eb898d0b0c3231c160446f219442f74/treetime/treeanc.py#L1329-L1346) function does something slightly different that the IQ-TREE flag and also works with trees from all tools (FastTree, RAxML, etc.). Separate from this PR, would it be helpful to have an `augur tree` flag like `--prune-short-branches` that would allow us to call the TreeTime code you’ve written here as a general function of tree building, independent of how the tree builder handles short branches?

rneher:
yes, this was my thinking (triggered by the need for a reference tree for clades). I wasn't aware of the `-czb` option. That certainly does the trick.

We could also stick this code into `refine`

https://github.com/nextstrain/augur/blob/master/augur/refine.py#L231

since `augur refine` already handles polytomy options.

huddlej:
Would you prefer the IQ-TREE flag over the new rule, for now, then?

I'm happy either way. With the new rule, I would just need to add the script to support conda environments.

We could discuss the `--prune-short-branches` type of option at the next meeting, but either tree or refine seem good.

PR #478, opened 2020-08-03
Regional builds should not display the map in triplicate

jameshadfield:
The map-triplicate functionality is useful for interpreting sequential transmissions which involve the entire world, however this comes at a performance cost (more elements in the DOM, and larger arrays for the map logic to loop through) as well as producing default map-bounds which nearly always show the entire world (although the default views are by no means perfect with the changes in this PR).

For regional nCoV builds, the focus is on an individual location (region) and thus it's not necessary to display the map in triplicate.

Current master (triplicate):

image

This PR (not triplicate):

image


trvrb:
I think this is a great idea. I'm all for performance improvements. I just rebased to `master` and tested with latest JSONs and everything worked as it should.

PR #479, opened 2020-08-07
produce root-sequence sidecar JSON

jameshadfield:
This modifies the default workflow to produce the sidecar root-sequence JSON
for each build. This is in preparation for https://github.com/nextstrain/auspice/pull/1197
(see there for an explanation of the advantages this JSON gives us)

@huddlej / @tsibley would you mind checking the snakemake usage in this commit?


jameshadfield:
Thanks @huddlej for that snakemake magic 😄 

@rneher this should be good to merge now

PR #483, opened 2020-08-12
Do not rely on TreeTime when augur filter supports ISO 8601 dates

huddlej:
The user's environment will not necessarily have TreeTime installed globally, leading to an error when running the basic workflow. This commit takes advantage of the fact that augur filter supports ISO 8601 dates as of v9.0.0 to avoid a call to TreeTime's numeric_date function.

This PR splits out a commit that was original part of #473. While that PR's change to the Python 3 executable should be mostly transparent to users (except for those using `--use-conda`, for whom it will make life better), this PR makes a backwards-incompatible change to the workflow that relies on augur version 9.0.0 and later.

### DIFF
--- a/Snakefile
+++ b/Snakefile
@@ -1,3 +1,4 @@
+from datetime import date
 from os import environ
 from socket import getfqdn
 from getpass import getuser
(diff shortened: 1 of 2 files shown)


rneher:
this seems very uncontroversial to me. If you fix the merge conflict, I think this can go in.

PR #484, opened 2020-08-14
Provide example FASTA sequences in a compressed format

### LINKED ISSUE #453: Repo can't be used as a template because it has files larger than GitHub allows
**Current Behavior**  
Click the "Use this template" button from the ncov main repo page. GitHub redirects to this error page:

image

**Possible solution**  
The only file in `master` that is bigger than 10 MB is `example_data/sequences.fasta` which is 12 MB. We could compress this and update our documentation to include a decompression step for folks after they've clone the repo.

huddlej:
The large file size of our example sequence FASTA (>10 MB) prevents the ncov
repository from being used as a template repository on GitHub. This commit
replaces the uncompressed data with a gzip compressed file and updates the
corresponding tutorial documentation to include a decompression step before
users start running analyses.

An alternate solution to this one would be to not host the example data in the
repository at all but instead host it in an S3 bucket, etc. One benefit of the
current approach is that all of the data users need to run an analysis come with
the repository, making a complicated workflow just a little simpler.

Resolves #453


rneher:
we could also support reading of `fasta.gz`.

emmahodcroft:
I tested this by downloading the compressed file into the `data` folder and following the new commands to unzip it. Worked for me! I think this should be merged to make it easier for people to use the repo.

PR #485, opened 2020-08-21
Allow use of `--subsample-max-sequences` in `filter`

emmahodcroft:
In `augur` 10.0.0 we have fantastic new argument for `filter` - `--subsample-max-sequences`, which means users do not have to set their own limits for the number of sequences (for example) per country, month, and year - but set an overall total and the program figures it out.

However, in the `ncov` repo we can only make use of parameters that fit in our rather extensive snakemake architecture. This PR changes some hardcoded bits to be more flexible, so users can supply either `--sequences-per-group` (the current option) or `--subsample-max-sequences` using the build parameters `seq_per_group` (same as currently) and `max_sequences` (to indicate the new option).

**There is an important (but logical) change in overriding - read below.**

### Notes on changes

Previously in `main_workflow.smk` we have the argument `--sequences-per-group` hard-coded and just look up the value to go in. Now, there's use of `_get_specific_subsampling_setting` to check for both options (which are both now optional). This means we must return the whole argument string (both argument and value) - `_get_specific_subsampling_setting` is modified to look out specifically for `seq_per_group` and `max_sequences` and return the correct argument string. 

### Override change

This results in one remaining issue - because we have a default value for `seqs_per_group` (in `parameters.yaml`) and this gets deep-merged with any users `builds.yaml` file, if you have sampling schemes with names matching the defaults in `parameters.yaml` (quite normal) and if you have specified `max_sequences` in the builds file, you actually end up with a `config` file that has both `max_sequences` *and* `seq_per_group` with the default parameter value. This is not allowed by `filter` (as it should be) and so causes an error.

We *should not be changing the values in `parameters.yaml`* at this point because other users may be depending on them. However, we already have the behaviour that whatever users specify in their `builds.yaml`, this will override the defaults in `parameters.yaml` - by putting in their builds file we assume they want whatever they've put. Thus, I've implemented the same behaviour for `max_sequences` here - if the users has supplied this, it 'overrides' the `seq_per_group` argument which comes by default, and just `max_sequences` is used. I think this is a pretty sensible solution, and will not disturb anyone depending on the defaults. 

### Testing
I have **not** changed anything in the parameters or `build.yaml` for the Nextstrain runs - this is a separate step. However, this will allow other builds to start using this parameter (the main Nextstrain builds can move whenever we want, if we want). I've used the new setting in the Neher Lab Europe builds - in just running up to `subsample` it works with both arguments. Am currently testing on 'Portugal' build, and also will test on Europe Nextstrain build, will report back when this finishes.


emmahodcroft:
@huddlej Thank you for such great and detailed feedback!

## Silent Preference

Yes, I had the same thoughts about this as well. I'm just very unsure what's the most intuitive option here - supplying an empty or 0 value certainly forces users to be more explicit, but it's also a bit arbitrary - everything else you just specify, so why to do this must you change 2 things? Particularly when from the user's perspective they will likely feel like they didn't "supply both" - their file has only one! However we could perhaps help with this via a useful error message - the one you have above about 'you've supplied both', but extending it to say, "here's what you need to include if you want to use `max_sequences`.

## User direction

For user direction, I think this is worth some discussion. I'm afraid we've backed ourselves a little into a corner here because users may already be using our defaults (`parameters.yaml`) for their own runs - if we remove these or change them we could really screw people's builds up. I don't really know the best way to handle this at this point. 

Perhaps the best way to handle this is to remove subsampling schema from `parameters.yaml` - at least then this will cause an error to anyone using them (rather than sneakily changing their build, possibly without them realising). Could we somehow make it give an error that sends them to a URL that explains how to 'fix' their builds? (Here's a list of the old subsampling commands, and here's where you need to put them in your profile.)

If we did this we'd also need to copy subsampling definitions into:
- the 'default' build (I believe this is what this is?) - `my_profiles/getting_started` - uses `region_global`
- `my_profiles/example` - uses `country`, `region`, `division`, `region_global`
( `my_profiles/example_advanced` defines all its own schemes)
- `nextstrain_profiles/nextstrain` - uses `region_global` & `region`

However this is going to make it that little bit harder to make your own run, as users will need to copy something over, rather than just trusting us to define what a good 'country' sampling is. It's really hard to weigh up what the damage from this would be - as a user I can imagine figuring out exactly what's happened and how to fix it might be _incredibly_ frustrating. 

## Nextstrain Main Builds

On that note though, I do think we should move to having our own copies of `region_global` and `region` in the `nextstrain_profiles/nextstrain` profile - at some point one of us is likely going to want to change the subsampling and I don't think we should be doing this in the defaults... right? 

## Documentation

Yes, happy to also make documentation whenever we figure exactly which way we want to go with this. 

## Wrapping augur commands

I don't really know enough about what it entails to use Snakemake's wrapper interface - how that would impact files users might be relying on, or how such a thing is implemented. So can't really comment! Except to highlight that, again, we'd need to think through how a change like this would impact users relying on our defaults, etc. 

My first thoughts on how to implement this PR were actually exactly along your last lines - to change so that the 'key' can just directly be what gets passed as an argument. It seems the kind of most 'beautiful' solution. However, it will break anyone using the current format to specify `seqs_per_group`, so this is why I didn't go this route. (Or, I guess we'd have to have some kind of special check to convert this? "If you don't find `sequences-per-group` check for `seqs_per_group`") 

It is worth noting that this doesn't overcome the main issue of how to override the default `seqs_per_group` value, so we'd still need to figure out how to handle that. 


# Bottom line

I'm having trouble weighing up which of these routes is likely to allow us to be most flexible in the future, with two big issues being:
- Should we remove things like 'subsampling' defaults from `parameters.yaml` to force users to specify their own, reducing the risk that their builds change 'secretly' if we change these parameters? (And how much destruction do we cause if we remove these/how do we mitigate that?)
- Can/should we modify existing variable names in the snakemake pipeline (changing `seqs_per_group` to `sequences-per-group`) and how much will this break things for existing users pipelines? (Even if they are defining their own subsampling, this will break stuff, so how to we mitigate that?)

I feel like maybe these issues are larger than this PR, so my inclination at the moment is to go with John's suggestion of forcing users to explicitly 'cancel' out the `seqs_per_group` variable and try to generate a useful error helping users figure this out. This will at least surface this to users, and we can perhaps tackle the larger questions raised above separately...

emmahodcroft:
(Force-push above is me removing a bunch of 'build'-related changes I accidentally committed here last night in the confusion, when they were meant to go `master`!)

huddlej:
I agree, @emmahodcroft, that most of the issues we're talking about here are outside the scope of this PR. I just got excited thinking about this stuff again and wanted to get thoughts down here. Maybe we can move this discussion to the discussion board to keep it going?

For this PR, I really like the solution of throwing an informative error when users have defined truthy values for both `seq_per_group` and `max_sequences`. To better expose this behavior, we should add an empty `max_sequences` key to our default subsampling rules and add comments to both keys in the example YAMLs to describe this mutually exclusive behavior.

In the mean time, we can copy the default subsampling schemes into the Nextstrain builds YAML and stop using the defaults. This could be a separate PR though.

workflow/snakemake_rules/main_workflow.smk
```diff
@@ -318,6 +323,12 @@ def _get_specific_subsampling_setting(setting, optional=False):
             build = config["builds"][wildcards.build_name]
             value = value.format(**build)
         else:
+            # If is 'seq_per_group' or 'max_sequences' build subsampling setting,
+            # need to return the 'argument' for augur
+            if setting == 'seq_per_group':
+                value = "--sequences-per-group " + str(value)
+            elif setting == 'max_sequences':
+                value = "--subsample-max-sequences " + str(value)
```
ttung, line 331:
```suggestion
            if setting == 'seq_per_group':
                value = f"--sequences-per-group {value}"
            elif setting == 'max_sequences':
                value = f"--subsample-max-sequences {value}"
```

emmahodcroft:
Hi @ttung Thanks for the PR and the comments here. Unfortunately due to some personal issues I haven't been able to take this further until now. I will prioritise looking at your PR tomorrow in light of the above discussion, and hopefully one way or another we can get this feature available very soon! I believe John's on holiday for the moment but I think we know the direction for this so it shouldn't need to be delayed much longer.

huddlej:
Echoing @emmahodcroft, here, thank you for testing this PR and adding your own, @ttung! I've collected the list of items we should address before we merge this based on all comments here: 

- [x] Print error and exit when both `sequences_per_group` and `subsample_max_sequences` are specified
- [x] Print error and exit when neither `sequences_per_group` nor `subsample_max_sequences` are specified
- [x] Update augur version in conda environment to include support for max sequences in subsampling
- [x] Update the message for subsample rule
- [x] Add an empty `max_sequences` key to our default subsampling rules and add comments to both keys in the example YAMLs to describe their required and mutually exclusive behavior
- [x] Use f-strings instead of str casting
- [ ] Add documentation for this new subsampling functionality to [the customizing analysis docs](https://github.com/nextstrain/ncov/blob/master/docs/customizing-analysis.md) 

@emmahodcroft Regarding the docs updates, the current docs in the tutorial refer to the `my_profiles/example_advanced_customization/builds.yaml` which is based on the Neher labs Swiss builds, right? Do you have an example you’d want to include here of how you use the max sequences approach now? Or would you favor updating the current Swiss example builds YAML to use this argument somewhere useful?

workflow/snakemake_rules/main_workflow.smk
```diff
@@ -342,7 +353,15 @@ def _get_specific_subsampling_setting(setting, optional=False):
 rule subsample:
     message:
         """
-        Subsample all sequences into a {wildcards.subsample} set for build '{wildcards.build_name}' with {params.sequences_per_group} per {params.group_by}
+        Subsample all sequences by '{wildcards.subsample}' scheme for build '{wildcards.build_name}' with the following parameters:
+
+         - group by: {params.group_by}
+         - sequences per group: {params.sequences_per_group}
+         - subsample max sequences: {params.subsample_max_sequences}
```
ttung, line 360: so this will be a bit odd because the other parameters will just be the value, but sequences_per_group and subsample_max_sequences will include both the value and the --arg.

huddlej: It's definitely not ideal. Unfortunately, we have a couple other cases where the parameter is actually an argument string for augur filter. I wanted to convey with the message what the possible parameters are and what the defined values are. The trade-off is some redundancy in the output. Here is an example I tried out while testing this branch:

```
        Subsample all sequences by 'another' scheme for build 'global' with the following parameters:

         - group by: country year month
         - sequences per group:
         - subsample max sequences: --subsample-max-sequences 100
         - exclude:
         - include:
         - query: --query "country == 'USA'"
         - priority:
```

Ideally, we'd have implemented our mapping of filter config to arguments more cleanly (maybe like I mocked up in [an earlier comment on this PR](https://github.com/nextstrain/ncov/pull/485#pullrequestreview-472843630) in the section "Wrapping augur commands in the Snakemake config interface").

Maybe a simpler alternative is to define the message like so, such that the argument params only appear when defined:

```
    message:
        """
        Subsample all sequences by '{wildcards.subsample}' scheme for build '{wildcards.build_name}' with the following parameters:

         group by: {params.group_by}
         sequences per group: {params.sequences_per_group}
         subsample max sequences: {params.subsample_max_sequences}
         {params.exclude_argument}
         {params.include_argument}
         {params.query_argument}
         {params.priority_argument}
        """
```

But this might look strange to users, too, when most arguments aren't defined and appear as newlines:

```
        Subsample all sequences by 'another' scheme for build 'global' with the following parameters:

         group by: country year month
         sequences per group:
         subsample max sequences: --subsample-max-sequences 100


         --query "country == 'USA'"

```

emmahodcroft: I think I'd prefer the clarity of the first approach even with the redundancy. Better for users to have double the idea of what's appearing rather than potentially being confused by newlines and 'random' argument values appearing below them, IMO!

emmahodcroft:
I've update both example files so that they have blank `max_sequences:` entries where nothing else changes (where we're using `seq_per_group`). I also added the same comment 'explanation' that @huddlej added to the `parameters` file yesterday.

I've then updated one part of the `example_advanced_customization/builds.yaml` file so that in the `switzerland` subsampling, the `country` level uses `max_sequences` instead of `seq_per_group`, and I've expanded the docs in `customizing-analysis.md` to use this example and explain the two arguments & their difference.

I wanted to do a quick test on a local build, but unfortunately we've run into some issues with today's sequences & running the diagnostic... I'll try to get back to this later! However, the docs should be good now (though happy to take recommendations/edits!)

emmahodcroft:
Was also thinking we could tweet about this change and maybe post in SPHERES when it's merged? As anyone pulling from `ncov` will get an error about needing to add a 'blank' entry to `max_sequences`, so we should warn about that if we can?

Here's a tweet (character count checked) if we want to go that route:
```
We've updated our `ncov` pipeline so you can specify the max # of seqs wanted in subsampled #SARSCoV2 builds!

If you define your own subsampling & want to continue using `seqs_per_group`, you'll need to add an empty `max_sequences` entry.
Read more here!

https://github.com/nextstrain/ncov/blob/master/docs/customizing-analysis.md#custom-subsampling-schemes
```

emmahodcroft:
@huddlej I adjusted the error that's thrown if you have 'both' arguments as I think if it's pulling one from default this is confusing - it says you specified both, but you didn't! It now says one could be a default and to add an 'empty' argument to solve it. (I hope that will make sense to most people...)

emmahodcroft:
@huddlej Have you run a full test through this? I'm trying at the moment and while each subsampling step (`sample-country.fasta` `sample-division.fasta` etc) is completing perfectly, when I try to run the whole thing I'm getting the error:

```
InputFunctionException in line 435 of /scicore/home/neher/hodcroft/ncov_sc/workflow/snakemake_rules/main_workflow.smk:
Exception: The subsampling scheme 'state' for build 'texas' must define `max_sequences` or `seq_per_group`.
Wildcards:
build_name=texas
```

But... every level has `seq_per_group` defined... and I can run each of these separately just fine.
It seems like the problem might be in `_get_subsampling_settings` which is called by `_get_subsampled_files` which is called by `combine_samples`. This only has a wildcard of the build name, so it returns the whole `subsampling` settings... which has no top-level key of `seq_per_group` or `max_sequences` (as these are within `country`, `region`, etc). 

Or did I just manage to somehow really screw things up?

workflow/snakemake_rules/main_workflow.smk
ttung: It seems it would be good to have a mapping of logical config names to 1) parameters for augur, and 2) textual description.

emmahodcroft:
> Could this be solved by merging the defaults after the user config has been loaded?

Sorry @ttung, not sure which bit you are referring to here!

ttung:
> Sorry @ttung, not sure which bit you are referring to here! (Could what be solved?)

Sorry! I was referring to the requirement that both `seq_per_group` and `max_sequences` are specified in order to override the defaults from `parameters.yaml`.

ttung:
This is sort of what I mean.  This is untested code, but hopefully it conveys the idea:

```
diff --git a/Snakefile b/Snakefile
index a3ecfe0..5d17ed2 100644
--- a/Snakefile
+++ b/Snakefile
@@ -6,6 +6,9 @@ from snakemake.utils import validate
 configfile: "defaults/parameters.yaml"
 validate(config, schema="workflow/schemas/config.schema.yaml")
 
+# save the subsampling scheme.  we will retroactively apply it to the config *after* the user's config is loaded.
+default_subsampling = config.pop("subsampling")
+
 # default build if none specified in config
 if "builds" not in config:
     config["builds"] = {
@@ -55,3 +58,27 @@ if "localrules" in config:
 if "custom_rules" in config:
     for rule_file in config["custom_rules"]:
         include: rule_file
+
+# apply the subsampling rules from defaults/parameters.yaml, selectively _removing_ any
+# overlapping seq_per_group and max_sequence directives.
+for subsampling_scheme_name, subsampling_scheme_directives in default_subsampling.items():
+    if subsampling_scheme_name not in config["subsampling"]:
+        # ezpz, doesn't exist over there, so just copy over.
+        config["subsampling"][subsampling_scheme_name] = subsampling_scheme_directives
+        continue
+    for region_type, region_directives in subsampling_scheme_directives.items():
+        if region_type not in config["subsampling"][subsampling_scheme_name]:
+            # again, ezpz, doesn't exist over there, so just copy over.
+            config["subsampling"][subsampling_scheme_name][region_type] = region_directives
+            continue
+
+        existing_region_directives = config["subsampling"][subsampling_scheme_name][region_type]
+        if "seq_per_group" in existing_region_directives or "max_sequences" in existing_region_directives:
+            # subsampling frequency rate is specified in the user's config, so drop it from the
+            # defaults.
+            region_directives.pop("seq_per_group", None)
+            region_directives.pop("max_sequences", None)
+
+        for region_directive_key, region_directive_value in region_directives.items():
+            if region_directive_key not in existing_region_directives:
+                existing_region_directives[region_directive_key] = region_directive_value
```

huddlej:
@emmahodcroft Thank you for catching the issue with the exception! It was definitely me and not you. :) I just pushed the change you recommended after confirming that it fixes the same issue for me locally. Separate from this PR, we might want to think about refactoring the `_get_subsampling_settings` function into separate functions that return 1) a complete subsampling scheme (a block of one or more sets of subsampling) and 2) one specific subsampling set from a scheme. I'll make an issue for this.

I also doublechecked by running the example Swiss build that we don’t really need the empty `max_sequences:` entries for custom subsampling schemes; we only need these when a scheme with the same name and subsampling sets is already defined in an upstream config (like `region_global` in `defaults/parameters.yaml`). It doesn’t hurt to include the extra empty settings, though.

@ttung Resolving the merge order issue (or how we define default subsampling) is out of scope for this PR, but we definitely need to continue this discussion. I’ll create a separate issue for this conversation and post the link here.

ttung:
> @ttung Resolving the merge order issue (or how we define default subsampling) is out of scope for this PR, but we definitely need to continue this discussion. I’ll create a separate issue for this conversation and post the link here.

My concern is that releasing this and then subsequently fixing this behavior in the future is churn for downstream users.  They'll be told that they need to define both, and then in the future don't define both.

huddlej:
> My concern is that releasing this and then subsequently fixing this behavior in the future is churn for downstream users. They’ll be told that they need to define both, and then in the future don’t define both.

I see what you’re saying. @emmahodcroft and I were opting for a little bit of user churn with the current PR over a potentially breaking change that fully resolves the issue of conflicting defaults.

The code you’ve proposed should resolve this specific issue of conflicting mutually exclusive arguments, but two different issues worry me about this solution:

1. It changes the default behavior of Snakemake’s config merge, making errors in config definitions harder to debug (for example: “Snakemake always does a deep merge of configs except for specific keys in the subsampling definitions” is harder to track than “Snakemake always does a deep merge of configs”).
2. It doesn’t protect us from similar issues that will arise in the future due to conflicting defaults in subsampling.

The issue of the conflicting defaults should only affect users who define their own subsampling config that reuses the same names as the defaults to provide their own custom parameters. Separate from the mutually exclusive keys issue, I am more concerned about folks unwittingly using this approach and picking up default settings that they don’t expect through Snakemake’s deep merge. Even if people use this approach intentionally, any change we make to the defaults has the potential to be silently breaking for them.

For example, here is a simplistic default subsampling definition that might live in `defaults/parameters.yaml`:

```yaml
# Define the default subsampling for "region_global".
subsampling:
  region_global:
    global:
      group_by: "country year month"
      seq_per_group: 20
      exclude: "--exclude-where 'division==USA'"
```

And here is an example of how users might override these defaults (intentionally or not):

```yaml
# Override the default subsampling scheme for "region_global".
subsampling:
  region_global:
    global:
      # Inherits `seq_per_group` and `exclude` from defaults.
      group_by: "country year month"
      max_sequences: 100
```

Even if we’ve addressed the issue of the mutually exclusive keys, we still have the issue of inheriting other keys (like `exclude`) from the deep merge. This may or may not be what users expect, but it would be safer if they didn’t rely on the defaults.

My preferred solutions to the merge issue would be to either:

1. Not define a default subsampling config and require users to define their own (e.g., copy and paste code we provide). Or,
2. Throw an informative error when users define their own subsampling config that conflicts with an existing default.

Both of these are potentially breaking changes. The first option will definitely break builds for everyone including people who don’t define their own subsampling but rely on the existing defaults. The second option will break builds for anyone who reuses parts of an existing subsampling scheme.

I would prefer the second option, since it breaks the fewest builds. We could even use some of the code you’ve proposed above to detect and report the conflicting subsampling schemes.

With the current state of the PR, we’re asking folks to define an empty parameter for either `seq_per_group` or `max_sequences`. As a user, I find this annoying, but I know my builds won’t break when I pull this new code. If the empty parameter isn’t required in the future, that is also annoying (or distracting from the functional parts of the config), but it shouldn’t break my builds.

huddlej:
@ttung @emmahodcroft After thinking about this some more over the long weekend, I added a deprecation warning when users define a subsampling scheme with the same name as a default scheme but different settings. Commit 4a05815 adds this warning with a detailed description of why conflicting subsampling schemes are a problem and a suggestion for how to fix the problem. This check uses the idea from @ttung's mockup code above to fetch the user's config before the default.

I think if we can move toward making these kinds of conflicts errors at run time before the workflow starts, we can avoid the confusion of defining empty subsampling attributes or generally worrying about what will happen when Snakemake deep merges configs.

What do y'all think of this as an alternative solution to defining empty attributes?

ttung:
I like this solution!  I have some comments which I'll put inline in the code.

Snakefile
```diff
@@ -1,11 +1,42 @@
...
 from getpass import getuser
+from snakemake.logging import logger
 from snakemake.utils import validate
+import time
+
+# Store the user's configuration prior to loading defaults, so we can check for
+# reused subsampling scheme names in the user's config. We need to make a deep
+# copy because Snakemake will deep merge the subsampling dictionary later,
+# modifying the values of a reference or shallow copy.
+user_subsampling = copy.deepcopy(config.get("subsampling", {}))
```
ttung, line 13: is the config even defined at this point?  I thought the default parameters are loaded here, and the user's config is loaded ~L49-57.

huddlej: Yeah, `config` is defined as an empty dictionary initially. If the user provides a `--configfile` command line argument or a `configfile` attribute in their profile config, the files referenced in that argument get loaded into the `config` dictionary.

Strangely, though, these files don't appear to get immediately deep merged if there are multiple arguments. For example, the "getting started" profile (`my_profiles/getting_started/config.yaml`) specifies the default parameters and its own build parameters. The keys from both files appear in the `config` dictionary, but the values from the custom build parameters appear to override the defaults instead of getting merged. For example, the default parameters specify these in the `files` config attribute:

```yaml
files:
  include: "defaults/include.txt"
  exclude: "defaults/exclude.txt"
  reference: "defaults/reference_seq.gb"
  outgroup: "defaults/outgroup.fasta" # is this arg still used? file doesn't exist in gh repo
  ordering: "defaults/color_ordering.tsv"
  color_schemes: "defaults/color_schemes.tsv"
  lat_longs: "defaults/lat_longs.tsv"
  clades: "defaults/clades.tsv"
  legacy_clades: "nextstrain_profiles/nextstrain/legacy_clades.tsv"
  subclades: "defaults/subclades.tsv"
```

And the getting started profile also defines a `files` attribute with:

```yaml
files:
  auspice_config: "my_profiles/example/my_auspice_config.json"
  description: "my_profiles/example/my_description.md"
```

But when I inspect the `config` dictionary first thing in the `Snakefile`, only the "getting started" `files` entries are in there. When I inspect the `config` dictionary after the `configfile` declaration in the `Snakefile`, all of the config attributes have been deep merged.

So, I guess the downside to this new approach is that it depends on some magical behavior of Snakemake that may not persist in the future.

huddlej:
I forgot to mention that I tested the new subsampling conflict warning by adding the following block to the "getting started" builds config (`my_profiles/getting_started/builds.yaml`):

```yaml
# Define custom subsampling schemes.
subsampling:
  # Custom subsampling logic for global region.
  region_global:
    global:
      group_by: "country year month"
      max_sequences: 100
```

This scheme conflicts with the default `region_global` scheme that specifies a number of sequences per group instead of max sequences. I tested this with this command:

```bash
snakemake -np --profile my_profiles/getting_started/
```

And got this output:

```
WARNING: The following subsampling scheme(s) have the same name as a default scheme in this workflow but different definitions:

  - region_global

  This means Snakemake will merge your scheme with the default scheme and may produce unexpected behavior.
  To avoid errors in your workflow, rename your schemes with unique names (e.g., 'custom_region_global')
  In future versions of this workflow, overlapping subsampling scheme names will produce an error.

Building DAG of jobs...
```

Snakefile
ttung: Ohh, I was completely mistaken.  The later loading is for snakemake rules.

My reading of [the snakemake docs](https://snakemake.readthedocs.io/en/v4.5.0/snakefiles/configuration.html#standard-configuration) is that if `--configfile` is specified, then the defaults are not loaded.  So I guess the only case there is the deep merge is if someone specifies `--config`?

huddlej: I think [the deep merge happens for all combinations of config inputs](https://snakemake.readthedocs.io/en/stable/snakefiles/configuration.html#standard-configuration) except when someone provides a `--config` that conflicts with an existing `configfile` entry, the `--config` value will overwrite the other values.

> Note that any values parsed into the config dictionary with any of above mechanisms are merged, i.e., all keys defined via a configfile statement, or the –configfile and –config command line arguments will end up in the final config dictionary, but if two methods define the same key, command line overwrites the configfile statement.

But it seems like the deep merge doesn't happen until after the `configfile` directive has been processed in the `Snakefile`. I can see how this could be a headache from the Snakemake developers' perspective as much as it from ours.

Do you know if nextflow behaves similarly?

emmahodcroft:
Sorry all, am catching up on this now! I think this idea overall seems a good one. I was a little confused with where the discussion over when the deep-merge happens ended, though - are we confident we'll catch same-name builds that modify the defaults?

If we're happy with this being handled as we wish, then I'm happy to go this direction. I'd like to keep some kind of warning/Exception if people add `max_sequences` and thus path 'both' values to `augur filter` - I think it's otherwise not very intuitive what the problem is. We could modify this to include a reference to changing the build name - but it might get a little long...?

Apart from that, we can remove the references to the blank values from the docs and from the example files, but we should go ahead and follow our own rules and ensure we've renamed our own example (and Nextstrain profile) subsampling schemes...! As part of that, we should ensure they're copied in full (not just the bits that change). Happy to help with this!

huddlej:
@emmahodcroft I dug into the Snakemake code a little more and traced the order that our inputs are getting parse. The general flow goes like this:

1. First, [configfile arguments are loaded and config is built from these](https://github.com/snakemake/snakemake/blob/a7ac40c96d6e2af47102563d0478a2220e2a2ab7/snakemake/__init__.py#L466-L471)
2. Then, [config arguments are loaded and override existing config keys](https://github.com/snakemake/snakemake/blob/a7ac40c96d6e2af47102563d0478a2220e2a2ab7/snakemake/__init__.py#L476-L477)
3. Then, the [Snakefile is parsed and configfile directive inside the Snakefile is processed](https://github.com/snakemake/snakemake/blob/a7ac40c96d6e2af47102563d0478a2220e2a2ab7/snakemake/__init__.py#L551-L553)
	* When configfile is loaded from the directive in the Snakefile, [the config dictionary is deep merged with the files](https://github.com/snakemake/snakemake/blob/a7ac40c96d6e2af47102563d0478a2220e2a2ab7/snakemake/workflow.py#L1088-L1094) from the externally provided config files.
	* This is the only place the deep merge happens using [the update_config function](https://github.com/snakemake/snakemake/blob/a7ac40c96d6e2af47102563d0478a2220e2a2ab7/snakemake/utils.py#L455-L476).

From this flow, we can be sure that the `--configfile` arguments provided by the user's `builds.yaml` will always get loaded before the default parameters associated with the `configfile:` directive in the `Snakefile`.

Interestingly, the deep merge only happens when there is a `configfile:` directive. That's not very intuitive, but at least now we know! :)

huddlej:
I've renamed the only subsampling scheme that conflicted with a default (the example profile's "county" scheme) and removed all references to the empty subsampling scheme attributes. I also updated the docs and error messages to reflect this change in approach.

I'm not sure we need to copy and rename the Nextstrain profiles here, since all of our builds use the default subsampling schemes, but we could do this to make our configs more future-proof.

emmahodcroft:
Sorry - late here so I'll read through properly tomorrow! But just wanted to add that we don't use the defaults anymore for the Nextstrain builds - just before I left on holiday we changed the sampling numbers to bring run size down. As part of that I went ahead and copied them into the Nextstrain profile `builds.yaml` so that any future changes are just to our runs and not to the defaults (which people might be using). I can try to resolve this by pulling `master` into this branch tomorrow!

But I believe we'll then want to rename all of the sampling schemes in the Nextstrain profile, as at the moment they're 100% copied from the defaults, just with different sampling numbers per country/month/yr (etc) - can do this too :)

huddlej:
Thanks, @emmahodcroft! I missed that change to the Nextstrain defaults. That would be awesome if you could migrate those then. We're so close to merging this!

emmahodcroft:
Ok, I pulled in master and I've updated the Nextstrain `builds.yaml` file so that all the subsampling regimes have new names. I checked and they are 'fully' copied over from defaults, so we shouldn't now be missing some specification (if we'd previously copied over just the variables we wanted to change, depending on the deep-merge to set rest to the defaults, this might have been an issue).

I went through John's commits for the docs & updating the error messages and all looks good!

PR #486, opened 2020-08-29
support for both --sequences-per-group and --subsample-max-sequences

ttung:
This allows ncov to use the `--subsample-max-subsequences` option introduced in https://github.com/nextstrain/augur/pull/593

### Testing
Started a ncov build that uses both `--sequences-per-group` and `--subsample-max-sequences` and it produces the expected commands.

### DIFF
--- a/workflow/snakemake_rules/main_workflow.smk
+++ b/workflow/snakemake_rules/main_workflow.smk
@@ -328,10 +328,36 @@ def _get_specific_subsampling_setting(setting, optional=False):
 
     return _get_setting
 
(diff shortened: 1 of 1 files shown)


rneher:
this looks like it has a similar aim to

https://github.com/nextstrain/ncov/pull/485

huddlej:
Thank you for posting this, @ttung, and for catching the similarity, @rneher. I'm going to close this in favor of the existing PR, but we have copied most of the changes from this PR into that one.

PR #487, opened 2020-09-09
Allow for per-build description files.

ttung:
This allows for description files to be customized per-build.

Fixes https://github.com/czbiohub/covidtracker/issues/1

### DIFF
--- a/docs/customizing-visualization.md
+++ b/docs/customizing-visualization.md
@@ -34,7 +34,20 @@ to your `builds.yaml` file.
 
 ## Changing the dataset description
 
-The dataset description, which appears below the visualizations, is specified in `my_profiles/example/my_description.md`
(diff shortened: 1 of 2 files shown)


emmahodcroft:
I haven't tested this but it seems a good idea, and uses the same patterns of logic we currently use to designate other build-specific things. Thanks @ttung !

huddlej (changes requested):
Thanks, @ttung! This works for me. As part of this PR, would you mind adding a brief code block example of how to use this new build attribute to [the documentation about customizing visualizations](https://github.com/nextstrain/ncov/blob/master/docs/customizing-visualization.md#changing-the-dataset-description)?

huddlej (approved):

huddlej:
Perfect! Thanks again, @ttung! Merging now.

PR #489, opened 2020-09-15
RFC: Align sequences with mafft's reference alignment mode instead of augur

huddlej:
Alignment with default mafft through augur align takes a very long time (in time per CPU) with our original workaround of partitioning sequences into smaller batches. Since April, [mafft has supported a reference-based alignment tuned for SARS-CoV-2-length sequences](https://mafft.cbrc.jp/alignment/software/closelyrelatedviralgenomes.html). This commit replaces the augur align command and partitioning of sequences with a single align command that uses the latest version of mafft.

The partition with augur align approach requires ~10 min on 96 CPUs. The mafft reference alignment approach requires ~13 min on 8 CPUs, making it about 10 times faster. This approach also has the benefit of being much simpler to maintain and run on a cluster or an AWS batch node.

An alternate (and maybe better long-term solution) would be to add support for this reference-based alignment mode to augur align. This solution should help most users in the short-term including the Nextstrain team as we continue to run on AWS Batch nodes with 96 CPUs (alignment should take <5 minutes).


huddlej:
I should have noted this is more of an RFC kind of PR. If folks like this general direction, I can refine the code and docs a bit more around this initial commit.

emmahodcroft:
I haven't had a chance to test this, but it sounds like a good direction to me - we are definitely burning a silly amount of core time re-aligning sequences every day, and anything we can do to be less wasteful & help others run this more easily (if they do not have 96 cores to hand) would be good. I guess we'd want to do a run to ensure this isn't introducing anything weird alignment-wise?

huddlej:
I agree, @emmahodcroft. I haven't spent as much time studying these alignments as other folks, so I will try to team up with someone to review an alignment from the current `master` approach and this reference-based mafft approach.

jameshadfield:
Thanks John - this direction seems like a good one to go in. Eventually I think this will want to be exposed by `augur align` but let's use it here for a while first.

rneher:
I am in favor of this. The only caveat being that everybody needs to update their mafft. I can run it tomorrow and look over the alignment.

emmahodcroft:
Good point Richard - we should be cautious to break this for people who need to update mafft!  Should probably have some intelligent error .... we don't want to unknowingly commit them to unefficient hours of alignment...

rneher:
I ran this today and it worked fine (merged master into it and upped the threads/mem config). The mafft version shouldn't be an issue for builds on aws. only people running it locally need to update their envs. 

increasing the number of threads to 16 would be good.

rneher:
One additional thing. we should add `--anysymbol` as a flag to mafft. otherwise this will fail on whenever a sequence has a weird character.

PR #490, opened 2020-09-17
Change SARS-CoV-2 tutorial heading level

eharkins:
This fixes a bug where this page would render as many sidebar headings [1] instead of just one in the docs.nextstrain.org read the docs build. 

[1] see https://docs.nextstrain.org/en/migrate-ncov-tutorial/tutorials/SARS-CoV-2/index.html ; note the extra headings when compared with https://nextstrain.github.io/ncov/.


jameshadfield (approved):
👍  Other pages in the tutorial use h1 headers as the main header, so this PR standardizes things. LGTM!

PR #495, opened 2020-09-29
Min max data subsampling

rneher:
Allow producing subsamples specific to date ranges. 

This PR introduces `max/min-date` fields for the subsampling rules. The envisioned use-case is tailoring sampling density to specific time slices. One might for example want a sparse sampling of the first half of the pandemic and a denser focus on the last few months. 

For Europe, this could look like this:

OLD:
image

NEW:
image

Note that this is completely optional, just exposing some additional flags.


huddlej (approved):
This is really cool! It would be a nice feature to add to seasonal flu builds some day, too.

rneher:
thanks John!

PR #499, opened 2020-10-11
Add frequencies panel to Auspice config

trvrb:

We have been exporting tip frequencies for all regional ncov builds as _tip-frequencies.json. However, these haven't been shown due to the absence of "frequencies" in the list of "panels" in the Auspice config. We now have enough temporal variation that I think it's warranted to include a frequencies view.

Further discussion on Slack here: https://bedfordlab.slack.com/archives/CSKMU6YUC/p1597861118040000

This does surface issues with geographic over-representation, especially in the global view, where Europe is over-represented:

ncov_global_frequencies

but I don't think this should block this PR. We should get regions equilibrated separately with https://github.com/nextstrain/ncov/pull/494.

### Testing
I've rerun `augur export` from latest AWS Batch build and everything looks as it should.

### DIFF
--- a/nextstrain_profiles/nextstrain/africa_auspice_config.json
+++ b/nextstrain_profiles/nextstrain/africa_auspice_config.json
@@ -130,6 +130,7 @@
   "panels": [
     "tree",
     "map",
-    "entropy"
+    "entropy",
+    "frequencies"
   ]
 }
--- a/nextstrain_profiles/nextstrain/asia_auspice_config.json
+++ b/nextstrain_profiles/nextstrain/asia_auspice_config.json
@@ -130,6 +130,7 @@
   "panels": [
     "tree",
     "map",
-    "entropy"
+    "entropy",
+    "frequencies"
   ]
 }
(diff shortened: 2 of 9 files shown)


jameshadfield:
FYI you can set `panels` inside the `display_defaults` as well ([docs](https://nextstrain.github.io/auspice/advanced-functionality/view-settings#dataset-json-configurable-defaults)) if you don't want the frequencies to be shown by default for certain builds.

emmahodcroft:
I actually wonder if this might be a good first stop? Enable it, but have it off by default, and we can manually turn on and evaluate how this is looking under different settings/builds, before turning it on for everything? Interest is rising sharply again, along with cases, but recency of sequence info is often pretty skewed - would be good to check that our 'leading edge' isn't too crazy, perhaps?

trvrb:
I think okay to hide things if preferred, but I would note that comparator tools (that are getting lots of use) are just raw looks at variant frequencies by week in GISAID. This will be more skewed than what we have. Subsampling and KDEs should help significantly to smooth out inference.

emmahodcroft:
Fair point.

rneher:
I am in favor of showing frequencies by now. the last 4 weeks are tricky though since they are heavily dependent on who uploads...

trvrb:
Can we merge this? Frequencies are becoming increasingly relevant.

emmahodcroft:
I think there's perhaps not a convincing argument not to. I'm ok with it!

PR #500, opened 2020-10-12
Write subsampling log output to a file

huddlej:
When I added support for logging output from all primary rules to files in a `logs/` directory, I forgot to add a `log` directive to the subsampling rule. This oversight causes logging output to be written to the screen only and [prevents users from easily debugging issues with subsampling](https://discussion.nextstrain.org/t/diagnosing-error-filtering-issues/149/5). This PR addresses this issue by adding a `log` directive to the subsampling rule.

I have tested this change with the following commands:

```bash
# Run "getting started" workflow up through subsampling.
snakemake --profile my_profiles/getting_started --use-conda -j 4 results/global/sample-global.fasta

# Confirm subsampling log output exists and is not empty.
cat logs/subsample_global_global.txt
```

The output of the last command should be:

```
-2 sequences were dropped during filtering
	0 of these were dropped because of subsampling criteria

	2 sequences were added back because they were in defaults/include.txt
4 sequences have been written out to results/global/sample-global.fasta
```

### DIFF
--- a/workflow/snakemake_rules/main_workflow.smk
+++ b/workflow/snakemake_rules/main_workflow.smk
@@ -374,6 +374,8 @@ rule subsample:
         priorities = get_priorities
     output:
         sequences = "results/{build_name}/sample-{subsample}.fasta"
+    log:
+        "logs/subsample_{build_name}_{subsample}.txt"
     params:
         group_by = _get_specific_subsampling_setting("group_by"),
         sequences_per_group = _get_specific_subsampling_setting("seq_per_group", optional=True),


emmahodcroft:
I haven't tested this myself but this is a great find John! I'd noticed this myself but all the logging seems a bit magical to me so I figured there was a technical reason we couldn't see it... 🙃 Thanks for catching and fixing this!

emmahodcroft (approved):

huddlej:
Thank you, @emmahodcroft! There isn't any specific magic in logging beyond the magic of the bash shell which is admittedly pretty magical :D

PR #501, opened 2020-10-13
add parse_new_sequences.py to developer_scripts

MoiraZuber:
#  parse_new_sequences.py

This python script can be used to create a concise overview of new or changed sequences provided within the `metadata-changes.txt` files that are uploaded to Slack as part of the daily ingest. The produced overview will contain a total count, a date check for every sequence and a collection of submitting lab twitter handles. 

## Input files
This script requires two files to be placed within the folder `inputs_new_sequences`:
1. The latest `metadata-changes.txt` file taken from Slack (optional: several `metadata-changes.txt` files can be placed within the folder and will be read in correctly, as long as they contain some numbering system that allows for sorting in the right order, e.g. `metadata-changes (1).txt`,  `metadata-changes (2).txt` etc. This way, the effects of following ingests can be included into the analysis. Only newly added sequences will be taken into consideration, however for these, all future changes found in subsequent `metadata-changes.txt` files will be applied for the overview).
2. The current `Who to Tag in Nextstrain Update Posts COVID-19.xlsx` excel file, downloaded from the google spreadsheet. This will be used for the automatic collection of submitting lab twitter handles.

## How to run
The script can be called with the command `python scripts/developer_scripts/parse_new_sequences.py` from the `ncov` repository.

## Console output
The script produces three types of output directly on the console:
1. **Date checks:** This check examines the sampling dates of all newly provided sequences in two fashions:
    - *Invalid sample dates:* All sequences that have invalid sampling dates will be listed. This includes incomplete dates (e.g. `2020`), future or past (before 2020-12) dates. Sequences listed here will automatically be removed from the final overview counts. Potentially, they might not be removed automatically from the run, so please double check whether the kind of invalidity detected will be detected by the system also.
    - *Early sample dates:* All sequences from February 2020 or before will be flagged, but not removed from the overall counts. Depending on the country, a double check for the first confirmed cases in this country might need to be performed, and the sequences might need to be excluded in case of inconsistencies.

2. **Overview counts:** All sequences that passed the invalid sample date check will be counted by total, country and division.

3. **Lab and twitter handles collection:** All *submitting labs*, *originating labs* and *authors* will be collected. For all submitting labs and potentially some originating labs, twitter handles will be searched for in the spreadsheet provided within the `inputs_new_sequences` folder. Lab descriptions that were not found are marked with a questionmark.

## Output files
Within the folder `outputs_new_sequences`, several files will be deposited:
1. **Date plots:** For each country, all sequence sampling dates are presented as bar plots. This can be used for a visual check of the distrbution of dates, and the detection of irregularities (such as a suspiciously large amount of sequences sampled on exactly the same day).
2. **Additional output:** It is encouraged to add additional checks and functions to the script that are useful for the daily checks and tweets. Suggestions are:
    - List of all strain names and submission dates - useful for finding the right date filters on nextstrain
    - List of all strains found within a certain region/country after a certain month - might be of personal interest
    - Additional utils for the automatic production tweets

### DIFF
--- a/scripts/developer_scripts/config_files_parse_metadata/requirements.txt
+++ b/scripts/developer_scripts/config_files_parse_metadata/requirements.txt
@@ -1,2 +1,0 @@
-# install with pip install -r scripts/developer_scripts/requirements.txt
-geopy
--- a/scripts/developer_scripts/parse_new_sequences.py
+++ b/scripts/developer_scripts/parse_new_sequences.py
@@ -1,0 +1,324 @@
+import os
+import datetime
+import matplotlib.pyplot as plt
+import pandas as pd
+from pandas.plotting import register_matplotlib_converters
+register_matplotlib_converters()
+
+
+
+def bold(s):
+    return('\033[1m' + s + '\033[0m')
+
+# Cut a string of the format "key: content" into a tuple (key, content)
+def cut(s):
+    key = s.split(":")[0]
+    content = ":".join(s.split(":")[1:])[1:]
+    return (key, content)
+
+# Read all files within given directory that start with "metadata-changes" in a sorted manner. Contents are stored in a
+# dictionary first by GISAID epi isl, then by info-type (e.g. date, division, host)
+def read_data(path):
+
+    data = {} # added and changed sequences
+
+    for file in sorted(os.listdir(path)):
+        if file == '.DS_Store':
+            continue
+
+        if file.startswith("metadata-changes"):
+            with open(path + file) as f:
+                metadata_changes = f.readlines()
+
+            added = False
+            changed = False
+            for i in range(len(metadata_changes)):
+                k = metadata_changes[i].strip()
+
+                if k.endswith("sequences added") or k.endswith("sequence added"):
+                    added = True
+                    changed = False
+                if k.endswith("sequences changed") or k.endswith("sequence changed"):
+                    added = False
+                    changed = True
+                if k.endswith("sequences removed") or k.endswith("sequence removed"):
+                    added = False
+                    changed = False
+
+                if k.startswith("gisaid_epi_isl"):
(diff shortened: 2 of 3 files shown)


rneher:
Thanks, Moira. This looks super useful! One small request: could you put the lines after line 315 into an `if __name__ == '__main__':` block (see here https://medium.com/python-features/understanding-if-name-main-in-python-a37a3d4ab0c3)

this would allow the functions you define to be imported by other scripts...

eharkins:
This has been saving me time getting twitter handles for sure! Thanks Moira. 

I did need to `pip install xlrd` to make it work with the .xlsx format of the twitter handle spreadsheet, so maybe we can add that in here somewhere in a similar way to:
https://github.com/nextstrain/ncov/blob/f95105d40e6984926f6f986197f4c462f79c1a62/scripts/developer_scripts/config_files_parse_metadata/requirements.txt

Maybe we can combine the two into an environment so that one can install dependencies needed for both scripts at once? Happy to add a commit attempting to do that next week if we think it's important enough - not really sure who it applies to other than us who use these scripts within the team.

MoiraZuber:
@rneher The main function is now introduced. @eharkins I am fine with both solutions! I agree that an environment seems like more than is necessary at the moment, but it might be a good idea planning ahead, since there might be more scripts added or other people using the pipeline in the future.

eharkins:
@MoiraZuber I created #508 which:
- adds the dependency i mentioned to a list so we can track them
- removes the .gitkeep files here since they will be ignored by 53a795525cadd5f3e8d307bbc6521b2ee12e1aa1 - I'm assuming this is ok given our discussion on slack, but i can always change it if not!

rneher:
Thanks for the fix, Moira. I'll merge this now!

PR #503, opened 2020-10-14
Early late

rneher:
This PR refocuses sampling on the second half to 2020 (June --) to resolve recent dynamics better. 

### Testing
let's see whether we like it...

https://nextstrain.org/staging/ncovRefocus/europe
https://nextstrain.org/staging/ncovRefocus/north-america

(global and asia currently missing)

### DIFF
--- a/nextstrain_profiles/nextstrain/builds.yaml
+++ b/nextstrain_profiles/nextstrain/builds.yaml
@@ -107,14 +107,31 @@ subsampling:
   # Custom subsampling logic for regions
   nextstrain_region:
     # Focal samples for region
-    region:
+    region_early:
       group_by: "division year month"
-      max_sequences: 2500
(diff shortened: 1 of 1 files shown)


trvrb:
I really like this general direction (with sparser sampling before timepoint _t_ and denser sampling after timepoint _t_). However, I'm concerned that _t_ is fixed to 2020-06-01 here. Is there a way to have _t_ always be 4 months prior to current date? Otherwise, we will have to be updating this with some regularity and we know from experience with `viruses-per-month` that this tends to creep without attention.

rneher:
yes, this is clearly a problem. But I can't think of a good way of doing this with the current way how the subsambling is set up. A monthly calendar event seems like the easier solution to me atm... but we could put some code into the function that parses the subsampling settings.

PR #511, opened 2020-10-27
Update lat_longs.tsv

acx01b:
just a sign in the longitude of Bretagne


emmahodcroft:
Thanks for bringing this to `ncov` @acx01b ! Much appreciated!

acx01b:
@emmahodcroft Sorry to disturb you again, I didn't find the corresponding piece of code but if with 20A/EU1 you have a way to define country specific clades in https://nextstrain.org/groups/neherlab/ncov/ 
then a clade for G25563T, C15324T in France (50% and 30% of the spring outbreak) and A20268G in Spain (40% of the spring outbreak) would be helpful, the goal being to see easily in the time tree in which lineage the recent sequences are falling,

PR #516, opened 2020-11-19
Download aligned

rneher:
add rules to download precomputed alignments and other files. 

the workflow is now essentially controlled by the rule-order directives.


rneher:
@huddlej I merged the reference alignment into this and added rules to download. have a look whether this works for you.

huddlej:
Thanks, @rneher! I will check this out today.

emmahodcroft:
I haven't had a look at this really closely but I guess my big question would be, is it very clear how to modify the pipeline so that if you can't download from our S3, or manually add the files from GISAID (as they're not there yet), then you can run it the 'old fashioned' way? (Or will this work automatically if it detects it can't download from S3?) We should also make sure it's clear where to put the aligned files if taken off GISAID (not in `data` folder!).

jameshadfield:
Thanks Richard!

P.S. @emmahodcroft I found that the default ruleorders here will align everything locally (i.e. no change) but that I could override the default ruleorder like so:

1. In `my_profiles/nz/builds.yaml`:
```
localrules: "my_profiles/nz/rules.smk"
```
2. In `my_profiles/nz/rules.smk`:
```
ruleorder: download_aligned>align
ruleorder: download_filtered>filter
ruleorder: download_refiltered>refilter
ruleorder: download_masked>mask
ruleorder: download_diagnostic>diagnostic
```

jameshadfield:
Are there any objections to merging this? It's opt-in functionality (via `ruleorder`s) and thus shouldn't break any existing workflows.

emmahodcroft:
Do we need/should we add a bit to the documentation so that we can refer people on how to do this? Or are we going to wait until we can negotiate with GISAID so the file is actually available for others to download?

At the moment it's essentially only useful for us (though, very useful!), so perhaps we should wait before writing documentation.

huddlej (changes requested):
This worked for me and really speeds up the build process, but I did notice a couple of issues that we might want to address before merging.

1. I can't tell how the build runners for Nextstrain would be able to restart part of the analysis from scratch (instead of downloading from S3) without manually modifying the rule orders here. For example, what if they need to update the filter rules and regenerate `filtered.fasta`?

Is this a realistic situation, @emmahodcroft? If so, we could add a config variable to the YAML that toggles the S3 download on or off.

2. Even though the masked alignment is already available from S3, the workflow has to run the `download` rule to get the metadata file. In doing so, it also spends unnecessary time and disk downloading the original sequences file. If we split the `download` rule into `download_sequences` and `download_metadata` rules, this would speed up new runs of the workflow by avoiding the sequences download. Since no one outside our group has access to the S3 bucket, this change shouldn't break anyone else's workflows.

If folks think these are reasonable changes, I can push a commit with example implementations.

huddlej:
> Do we need/should we add a bit to the documentation so that we can refer people on how to do this?

I agree that we should wait to provide any documentation about this until we have figured out what the workflow will look like for users who get an alignment from GISAID.

rneher:
splitting the download rule makes total sense. so far the PR was a "minimal intervention". 

one other thing that should be added is the exclude to the subsampling like so

https://github.com/neherlab/ncov-europe/commit/956f51ebad82559ed7031f93770a6bcec7c149c5

otherwise, you don't exclude what needs excluding (our original filter happens before aligning).

rneher:
we should also think about where/when to run the alignments... but this is a different topic.

emmahodcroft:
> one other thing that should be added is the exclude to the subsampling like so
But this is only valid if the exclude rule has changed between when `filtered.fasta` was generated and when one is running, right? Just so I'm sure I'm understanding this clearly. So likely an issue for our 'side' runs but not for the main build?

>Is this a realistic situation, @emmahodcroft? If so, we could add a config variable to the YAML that toggles the S3 download on or off.

Is this part tied to Richard's suggestion above? In that if build shepherds modify `exclude.txt` then these would just be filtered out later in `subsample`? Alternatively we could just have the main Nextstrain builds start from the `aligned.fasta` instead of `filtered.fasta` and maintain doing the filtering as part of the main run.

>Even though the masked alignment is already available from S3, the workflow has to run the download rule to get the metadata file. In doing so, it also spends unnecessary time and disk downloading the original sequences file. If we split the download rule into download_sequences and download_metadata rules, this would speed up new runs of the workflow by avoiding the sequences download. Since no one outside our group has access to the S3 bucket, this change shouldn't break anyone else's workflows.

I'm in favour of this. Might be nice if we keep the `download` rule as having the output of two rules: `download_meta` & `download_fasta` (or something) so that those of us who have `snakemake download -f` in muscle memory can continue living an easy life ;) But actually a lot of time I do just want the metadata, so having a `download_meta` would save me time pretty often.

huddlej:
> Is this part tied to Richard's suggestion above? In that if build shepherds modify exclude.txt then these would just be filtered out later in subsample?

I was thinking more generally about any changes to the filter config after the alignment step and how folks would re-run steps, if they needed to. Starting from `aligned.fasta` instead of `masked.fasta` would address this problem, I think.

> I'm in favour of this. Might be nice if we keep the download rule as having the output of two rules: download_meta & download_fasta (or something) so that those of us who have snakemake download -f in muscle memory can continue living an easy life

Cool! I will push a commit with these changes a little later. I just added `download_metadata` and `download_sequences`. I'll add a `download` rule that refers to both of these as a shortcut though.

rneher:
I removed exclude filtering from the preprocessing pipeline

rneher:
this added the exclude step to subsampling. @emmahodcroft brought up the point that we could incorporate the preprocess pipeline by adding an `upload` rule to this workflow. 

(my initial attempt to carve it out was more a proof of concept).

this would mean including this:
https://github.com/nextstrain/ncov-preprocess/blob/master/Snakefile#L4

along with a few scripts that I took from ncov ingest. 

Argument for having a separate pipeline: don't complicate ncov further
Argument against: avoid having two align workflows.

emmahodcroft:
I can see the point for not complicating `ncov` further, but I think having this 'all-in-one' might make it easier to maintain, and easier to make sense of by looking at.

emmahodcroft:
@jameshadfield has identified that another issue is that since the stripped down rule used to generate the `filtered` `masked` etc files also doesn't include any date filtering, this is also now 'missing' from the pipeline. 

I think the most straightforward way is likely to add the `preprocess` scripts to `ncov` (as we discuss above). We can do two options after that:

- Keep the 'simple' filter prior to generating the `masked` etc files and push these up to S3 - then add another filter step when it 'comes back down' to go through our pipeline. (Pro: allows anyone to modify that filter step easily to match different requirements. Con: generates another 5GB file in the process)
- Make the filter step prior to generating `masked` etc be the 'full' filter step that does dates and all - everyone gets a version of alignment with our own filtering applied (pro: you can get to work straight away with this file, no extra file generation needed. Con: no custom filtering easily possible)

rneher:
OK. I changed this such that upload is now part of the main workflow and whether this is used or not depends on rule orders. It also uses `xz` compression which can be specified in the `build.yaml`. Filtering happens following masking.

workflow/snakemake_rules/main_workflow.smk
```diff
@@ -45,6 +45,17 @@ rule filter:
             --output {output.sequences} 2>&1 | tee {log}
         """
 
+rule download_filtered:
+    message: "Downloading metadata and fasta files from S3"
```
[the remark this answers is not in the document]
rneher: should be fixed now

rneher:
I am testing this with our europe builds again. But I think this can be merged soon.

workflow/snakemake_rules/main_workflow.smk
```diff
@@ -1,58 +1,57 @@
...
-        metadata = config["metadata"]
+        sequences = config["sequences"]
     conda: config["conda_environment"]
+    params:
+        s3_bucket = config["S3_BUCKET"]
     shell:
         """
-        aws s3 cp s3://nextstrain-ncov-private/metadata.tsv.gz - | gunzip -cq >{output.metadata:q}
-        aws s3 cp s3://nextstrain-ncov-private/sequences.fasta.gz - | gunzip -cq > {output.sequences:q}
+        aws s3 cp s3://{params.s3_bucket}/sequences.fasta.gz - | gunzip -cq > {output.sequences:q}
```
jameshadfield, line 10: There's a bit of inconsistency here where the other rules use `{params.compression}` but this uses gzip. Switching would require us modifying the ingest scripts which is for another day.

workflow/snakemake_rules/main_workflow.smk
```diff
@@ -1,58 +1,57 @@
...
-            --max-date {params.date} \
-            --min-date {params.min_date} \
-            --exclude {input.exclude} \
-            --exclude-where {params.exclude_where}\
-            --min-length {params.min_length} \
-            --output {output.sequences} 2>&1 | tee {log}
+        aws s3 cp s3://{params.s3_bucket}/metadata.tsv.gz - | gunzip -cq >{output.metadata:q}
         """
 
+rule upload:
```
jameshadfield, line 25: Is our intention to have other users run these rules (by specifying their own buckets), or is this nextstrain-only functionality?

If it's nextstrain only, then I would advocate moving this rule to a snakemake file in `nextstrain_profiles/preprocess`. 

Similarly, if the downloading of preprocessed files is nextstrain only, I think it'd be simpler for them to live in `nextstrain_profiles/nextstrain`. This would make the pipeline conceptually easier to understand, and the default workflows wouldn't have to worry about ruleorders etc.

jameshadfield (commented):
Thanks @rneher -- looks pretty good. I have two high-level comments.

1. Is this functionality nextstrain-speicific, or is the intention that others will leverage this? (See in-line comments for more on this). 

2. We are adding more and more filter steps (currently we have `prefilter`, `refilter`, `filter` and n x `subsample`. There's scope for simplification here. I'd be in favor of removing `rule filter` and have `rules subsample` do per-subsample filtering but this is best left for another PR (and there are backwards-compat considerations here).

workflow/snakemake_rules/main_workflow.smk
rneher: I kept these files as gzip since they are shared via GISAID while the remainder is just for us...

workflow/snakemake_rules/main_workflow.smk
rneher: that is a good idea. We can move the up and download rules to `local_rule` files.

rneher:
regarding your comment on the extra filter rules. I think there is a case to be made for having the subsampling and filtering be separate rules. just to keep them simple... there are too many filters right now but as you say, best left for another PR.

workflow/snakemake_rules/main_workflow.smk
jameshadfield: Ahh. good point.

jameshadfield:
> We can move the up and download rules to local_rule files.

8d2bf53  does this (via `custom_rules`, but same idea). I think it simplifies things a bit, but am equally happy for it to be dropped if people disagree. Tested via the following DAGs:
```
snakemake -f auspice/ncov_global.json  --profile my_profiles/example  --dag | dot -Tpdf > dag.default-example.pdf
snakemake -f auspice/ncov_oceania.json  --profile nextstrain_profiles/nextstrain  --dag | dot -Tpdf > dag.default-nextstrain.pdf
snakemake -f upload  --profile nextstrain_profiles/preprocess --dag | dot -Tpdf > dag.default-preprocess.pdf
snakemake -f auspice/ncov_oceania.json  --profile nextstrain_profiles/nextstrain-rhino/ --dag | dot -Tpdf > dag.default-rhino.pdf
```

jameshadfield (approved):

rneher:
Thanks!

rneher:
One more thought. if we add the `upload` rule to the main workflow on aws, the alignment will always be done and provided on S3 along with the core build.

nextstrain_profiles/preprocess/rules.yaml
```diff
@@ -0,0 +1,5 @@
+# Note that `custom_rules` from multiple configuration YAMLs are combined by replacement,
+# not by concatenation
+custom_rules:
+  - nextstrain_profiles/nextstrain/download_preprocessed.smk
```
huddlej, line 4: @jameshadfield I'm sorry if I missed a step in the conversation above, but does the preprocess profile need to have the download rules if the goal of this profile is to re-run analyses without the download?

huddlej:
> One more thought. if we add the upload rule to the main workflow on aws, the alignment will always be done and provided on S3 along with the core build.

I like this idea. Based on the reorganization of rules in this PR, we could do this with the following steps:

1. include the `preprocess/rules.smk` in our main Nextstrain profile
2. call the `upload` target in the ncov-ingest workflow along with [the current deploy-to-staging target](https://github.com/nextstrain/ncov-ingest/blob/4f5aeaf9d54616dd4b18744661f5108d127fa77d/bin/rebuild-staging#L56)

For clarity, it might be better to move the `upload` rule into [the existing export_for_nextstrain rules](https://github.com/nextstrain/ncov/blob/master/workflow/snakemake_rules/export_for_nextstrain.smk), since uploading to S3 is what the deploy to staging does anyway.

Would we still want a separate preprocess profile if the upload happens automatically as part of the daily builds and the rule lives in the existing export rules file?

Slightly off-topic, but I really like having the S3 download rules in their own separate place.

nextstrain_profiles/preprocess/rules.yaml
jameshadfield: Yeah it does, but it's a little convoluted why this is. Without it (and therefore the `download_<STEP>` rules aren't available) then the ruleorder blocks which refer to them in `preprocess/rules.smk` and `nextstrain-scicore/localrules.smk` cause problems. The former is easy to remove, but the latter file is coming via a `localrules` config entry in a configfile imported by `preprocess/config.yaml`. So we could switch that latter import to be via `custom_rules`, but then I think we'd need to add these rules to it as well:

```
  - workflow/snakemake_rules/export_for_nextstrain.smk
  - nextstrain_profiles/nextstrain/subsampling_ranges.smk
```

jameshadfield:
> Would we still want a separate preprocess profile if the upload happens automatically as part of the daily builds and the rule lives in the existing export rules file?

No - I think we can remove a profile in this case (all else being equal, i'd prefer less profiles). 

> I really like having the S3 download rules in their own separate place.

Glad you also found this easier to reason with!

> For clarity, it might be better to move the upload rule into the existing export_for_nextstrain rules

Or move all of them to the nextstrain profile ;)

nextstrain_profiles/preprocess/rules.yaml
jameshadfield: We might be able to simply switch `custom_rules` to `localrules` (in `nextstrain_profiles/preprocess/rules.yaml`), which would be cleaner.

rneher:
The one thing that I currently don't see through is how to have different rule orders in the same profile. The preprocess profile and nextstrain profile have different rule orders right now. If we merged the upload rule into main nextstrain, the preprocess profile could go. And the core build would run fine and do the alignments (and then hopefully upload). but other builds would need to include the download rules and define the other rule order.

rneher:
the commit I just pushed allows toggling between the upload/download behavior by including a different set of rules. 

https://github.com/nextstrain/ncov/blob/bc2581210117042c69cd9683f292b01702e666e0/nextstrain_profiles/nextstrain/builds.yaml#L6

the current state is what nextstrain would run on aws (after adding the upload target). if you wanted to run locally using the downloaded files, you'd need to include the download_rules

rneher:
added upload to deployment as per @huddlej suggestion.

jameshadfield:
This is looking great. I'm in favor of merging -- and then I can start working on #519 😄

rneher:
ok. we'll merge this tonight after today's run started and adjust ingest + testing over the weekend.

PR #518, opened 2020-12-02
Allow parameters to define ambiguous date exclusion criteria

jameshadfield:
This allows both the main filtering step as well as individual subsampling schemes to define if ambiguous dates are to be excluded, and if so, what level of ambiguity is to be excluded.

By having this in the subsampling steps, we allow flexibility so that (e.g.) we can allow ambiguous sequences for a particular subset of (important) samples.

NOTE: requires augur version `10.1.0` or higher

### DIFF
--- a/defaults/parameters.yaml
+++ b/defaults/parameters.yaml
@@ -50,7 +50,8 @@ filter:
   min_length: 27000
 
   # Omit sequences with incomplete date annotations, and USA seqs without a state.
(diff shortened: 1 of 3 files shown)


huddlej (approved):
This looks excellent, @jameshadfield! Thank you for putting it together.

Could we also update the conda environment (`workflow/envs/nextstrain.yaml`) in this PR to require the latest Augur (10.1.1) so people will have the right version when they use `--use-conda`?

huddlej:
@jameshadfield I just added the new augur version requirement and rebased this branch onto the latest master to resolve conflicts.

I've tested this with the getting-started data. If everything looks ok to you, still, we can merge this.

PR #519, opened 2020-12-02
Allow config-defined multiple sequence & metadata files

jameshadfield:
A common use case is the desire to start from multiple sequence / metadata files. For instance, a build starting with the GISAID data and a set of new (non-public) sequences.

~This extends the workflow to allow `config["metadata"]` and `config["sequences"]` to be an array. If so, then an additional rule (`combine_input_metadata` and `combine_input_sequences`, respectively) are run before analysis steps.~

This PR extends the workflow logic to be able to take a number of inputs (matched TSVs + FASTAs) by allowing `config["sequences"]` and `config["metadata"]` to be a dictionary (see examples below). Each input, if there are multiple, is referred to as an "origin" (via wildcards). Each origin has it's own separate filter / align /  diagnostic / mask steps (see DAGs below). This is by design as it will allow us to maintain pre-computed alignments / masked-alignments of S3 (gisaid) data and then replace the appropriate local computation steps, whilst still allowing seaparate (private) sequence / metadata inputs.

Note that metadata is combined via `scripts/combine_metadata.py` which respects the order of metadata files supplied, and thus will overwrite conflicting (non-empty) fields, such that the last file supplied takes precedence.

### Example: multiple inputs

The following input config will produce the following DAG

```yaml
sequences: 
  gisaid: "data/sequences.fasta"
  private: "data/example_sequences.fasta"
metadata:
  gisaid: "data/metadata.tsv"
  private: "data/example_metadata.tsv"
```

image

### Future: use pre-computed masked alignment

In the future, this workflow will be able to be converted to use the pre-computed files in #516, using a DAG along the lines of

image

### Bug - cannot overwrite string value of `config["metadata"]` with a dict?

Ideally, the former syntax for defining starting files (e.g. `sequences: "data/sequences.fasta"`) would remain in `defaults/parameters.yaml` and custom profiles could overwrite this with a dict, as needed. This would maintain backwards compatibility with the many builds out there. However it seems you cannot overwrite a string value with a dict:

```
TypeError in line 32 of /Users/naboo/projects/nCoV/nz/ncov/Snakefile:
'str' object does not support item assignment
  File "/Users/naboo/projects/nCoV/nz/ncov/Snakefile", line 32, in <module>
```

### Todo - commented out rules

The rules `excluded_sequences`, `align_excluded`, and `diagnose_excluded` have been commented out here as I am unsure how they are used. If someone could comment on how they use them then I'll add the appropriate wildcards and restore this functionality. (The `--min-length 50000` argument in `rule excluded_sequences` makes me think these rules aren't being used!)

### Todo - potentially unnecessary `combine_masked` step

In the case where there is a single sequence input, the logic in this PR adds a `collect_masked` step between `mask` and the subsampling steps. I believe this could be skipped in this case. (Were the config inputs to always be dicts this would be simple.)

### Feature - Custom filtering options for each input origin

The `config["filter"]` options now allow a dictionary of values specific to the origin. Using the above example,
```yaml
filter:
  min_length: 27000
  private:
    min_length: 20000
```
Would allow the private dataset to include shorter sequences, which may be important for certain scenarios where you want to see lower-quality sequences.

### Feature - ability to skip diagnostic step for certain input origin

Adding the following to the config YAML
```yaml
diagnostic:
  skip:
     - "private"
```
Will skip the QC filtering steps (rules `diagnostic` and `refilter`) for the "private" input sequences only. Similar to custom filtering, this can be important to allow lower-quality sequences which are of interest to be included. 

### Note
The longer-term plan is for augur to natively allow multiple metadata/sequence inputs, which would simplify this logic and remove a couple of rules.

### DIFF
--- a/defaults/parameters.yaml
+++ b/defaults/parameters.yaml
@@ -7,6 +7,7 @@
 conda_environment: "../envs/nextstrain.yaml"
 
 # These are the two main starting files for the run
+# Each can take an array (and the files will be merged appropriately) or a string.
 sequences: "data/sequences.fasta"
 metadata: "data/metadata.tsv"
 
--- a/defaults/parameters.yaml
+++ b/defaults/parameters.yaml
@@ -7,7 +7,7 @@
 conda_environment: "../envs/nextstrain.yaml"
 
 # These are the two main starting files for the run
-# Each can take an array (and the files will be merged appropriately) or a string.
+# Each can take either a string (file path) or a dict of input "origins", each specifying a filepath
 sequences: "data/sequences.fasta"
 metadata: "data/metadata.tsv"
 
@@ -47,6 +47,12 @@ filter:
 # The number of sequences per group determines the run time of a single alignment job.
 partition_sequences:
   sequences_per_group: 150
+
+# Diagnostic settings
+# You can opt-out of running diagnostics for certain inputs (be careful!)
+# diagnostic:
+#   skip:
+#     - "origin_name_here"
 
 # Mask settings determine how the multiple sequence alignment is masked prior to phylogenetic inference.
 mask:
--- a/scripts/combine_metadata.py
+++ b/scripts/combine_metadata.py
@@ -1,0 +1,80 @@
+import argparse
+from augur.utils import read_metadata
+from Bio import SeqIO
+import csv
+
+EMPTY = ''
+
+# This script was written in preparation for a future augur where commands
+# may take multiple metadata files, thus making this script unnecessary!
+#
+# Merging logic:
+# - Order of supplied TSVs matters
+# - All columns are included (i.e. union of all columns present)
+# - The last non-empty value read (from different TSVs) is used. I.e. values are overwritten.
+# - Missing data is represented by an empty string
+
+def parse_args():
+    parser = argparse.ArgumentParser(
+        description="Custom script to combine metadata files",
+        formatter_class=argparse.ArgumentDefaultsHelpFormatter
+    )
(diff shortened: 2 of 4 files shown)


jameshadfield:
Note that we can rather easily replace the local generation of a masked alignment for the "gisaid" origin by adding this rule to a snakefile (inspired by #516)

```python
ruleorder: download_gisaid_masked>mask

rule download_gisaid_masked:
    message: "Downloading metadata and fasta files from S3"
    output:
        sequences = "results/masked_gisaid.fasta"
    shell:
        """
        aws s3 cp s3://nextstrain-ncov-private/masked.fasta.gz - | gunzip -cq > {output.sequences:q}
        """
```

rneher:
I think general direction is very useful. I have few thoughts...

 * you need to change quite a bit in the alignment section. We are planning to change this anyway (the download PR #516  for example uses the reference alignment). Would be easier to build this PR on that one. 
 * we might just want to delete the part on the excluded sequences and deal with them separately. 
 * the combination of metadata could have been done by `pandas.concat` https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.concat.html.  I think what you have done is good and serves the purpose -- I was just wondering why you went down that road.

jameshadfield:
Thanks @rneher -- i'll rework this on-top of #516 once it's gone into master so let's prioritise getting that merged.

rneher:
@jameshadfield #516 is now merged. I think it would be good to re-up this.

jameshadfield:
Closed in favor of #549

PR #520, opened 2020-12-10
Remove unconnected word, "either" from naming clades docs.

donkirkby:
Thanks for publishing this project. I noticed a small mistake in the documentation of how clades are named, so this fixes that. Perhaps you could also consider clarifying the 20% threshold for naming a new clade. Presumably you select a set of samples within some time window, is that easy to describe? Also, our lab was initially confused about how you could have more than five clades if they all had to be over 20% prevalance. We concluded that a clade can later drop below 20%, leaving room for new clades to emerge. If that's right, it might be helpful to be explicit in your description.

Thanks again.

### DIFF
--- a/docs/naming_clades.md
+++ b/docs/naming_clades.md
@@ -11,7 +11,7 @@ The nomenclature used by Nextstrain to designate clades for SARS-CoV-2 is driven
 
 #### Definition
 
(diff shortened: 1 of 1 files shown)


emmahodcroft:
Hi @donkirkby - Thanks for this PR and for spotting the extra word.

And you're right, apologies for the ambiguity here which is unintentional. We definitely do mean that it can be at 20% at any point in time, but it does not need to stay there (it won't 'unbecome' a clade, if so). I'll try to make this clearer!

PR #521, opened 2020-12-17
Use an absolute path to the conda environment

huddlej:
Updates the expected format of the `conda_environment` config parameter to be a path relative to the top-level Snakefile instead of a path relative to the main workflow files in `workflow/snakemake_rules/`. This change fixes a bug where user-defined workflow rules in profiles (e.g., `wa_profile/download_preprocessed.smk`) are nested in a different directory structure than the main workflow rules and cannot find the conda environment based on a relative path.

This commit explicitly defines the `conda_environment` path as relative to the top-level directory and then builds an absolute path from that relative path such that all rules will reference the same file regardless of their directory structure.

### DIFF
--- a/Snakefile
+++ b/Snakefile
@@ -1,5 +1,6 @@
 import copy
 from datetime import date
+import os
 from os import environ
 from socket import getfqdn
 from getpass import getuser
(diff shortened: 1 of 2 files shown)


huddlej:
I think this issue might have affected you, @rneher, since you asked about it on Slack. At the time, I couldn't see how the original relative path could have caused this specific bug. This commit seems to have fixed the same issue in @cassiawag's [ncov-wa-build](https://github.com/blab/ncov-wa-build).

Snakefile
```diff
@@ -1,5 +1,6 @@
 import copy
 from datetime import date
+import os
 from os import environ
```
cassiawag, line 4:
```suggestion
```
To prevent redundancy.

rneher:
Sorry for missing this. This looks good to me -- I was previously mainly puzzled by the path (we don't actually run this with `--use-conda`)

rneher (approved):

PR #522, opened 2020-12-21
Allow build names to use alphanumeric characters

trvrb:
This allows for build names like `A222V` or `501Y-V1`. When experimenting with this just now I couldn't figure out why Snakemake was refusing to build until I caught this wildcard constraints issue. I imagine this would surprise others as well.

### DIFF
--- a/Snakefile
+++ b/Snakefile
@@ -66,9 +66,9 @@ BUILD_NAMES = list(config["builds"].keys())
 
 # Define patterns we expect for wildcards.
 wildcard_constraints:
-    # Allow build names to contain alpha characters, underscores, and hyphens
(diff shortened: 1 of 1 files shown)


emmahodcroft:
Hm! I have a memory of looking over a PR that @huddlej did of this! Did we just forget to merge it? Apologies if so... Super strange! Anyway, we should definitely include this or the original (I imagine they are basically the same :) )

cassiawag:
FWIW, the WA ncov builds have had this modification for months now and works just fine: https://github.com/blab/ncov-wa-build/blob/master/Snakefile#L78

huddlej (approved):
This worked for me! Merging now.

PR #523, opened 2020-12-22
Print failed logs

cassiawag:
Adds Snakemake's `show-failed-logs` option to Nextstrain and example config files. This flag will automatically print log files of any failed jobs to the screen for easier debugging.


huddlej (approved):
This worked for me on all the modified profiles. Thank you, @cassiawag, for this helpful addition to the workflow!

PR #524, opened 2020-12-22
Allow numbers in build names

### LINKED ISSUE #482: Build name wildcard does not support numbers
**Current Behavior**

@rneher points out that build names cannot currently have numbers in them or Snakemake will produce an error that is difficult to interpret. These errors are caused by [the overly strict build name wildcard constraints](https://github.com/nextstrain/ncov/blob/667b0ac18863813989dac5d7c6e23034c9609115/Snakefile#L28).

**Expected behavior**

We should support numbers in build names, especially for cases when we want to place dates in the build name (e.g., for historical builds of specific time periods).

**Possible solution**

Changing the current constraints regex to `build_name = r'(?:[_a-zA-Z0-9-](?!(tip-frequencies|gisaid|zh)))+'` should fix the problem. I have not tested this yet though to confirm.

huddlej:
This is a second attempt to add support for numbers in build names after #522 unexpectedly broke the Nextstrain team's builds. This PR reduces the constraints on build names for most builds and adds rule-specific wildcard constraints to Nextstrain-dated builds. These constraints prevent dates from being in Nextstrain build names so the standard Nextstrain builds can be saved as dated copies. To prevent ambiguity between the dated JSONs rule and the main workflow's finalize rule, this commit also specifies a preferred rule order for the Nextstrain builds.

(**Updated on November 24, 2021**)
I tested this new code with the following build config (`builds.yaml`):

```yaml
inputs:
  - name: example-data
    metadata: data/example_metadata.tsv
    sequences: data/example_sequences.fasta.gz

default_build_name: my-build-2021-11-24
```

And then the following command:

```bash
snakemake --configfile builds.yaml -p -j 1
```

I also confirmed that the Nextstrain profile's dryrun worked as expected:

```
snakemake --profile nextstrain_profiles/nextstrain-gisaid all_regions -np
```

These changes should allow people to name their builds nearly anything they like including dates (e.g., `global-2020-12-22`) without breaking the existing Nextstrain naming scheme for dated builds.

- [x] Tested with CI
- [ ] [Tested with full Nextstrain open build](https://github.com/nextstrain/ncov/runs/4621236854)

Resolves #482

### DIFF
--- a/Snakefile
+++ b/Snakefile
@@ -133,7 +133,7 @@ config["conda_environment"] = CONDA_ENV_PATH
 wildcard_constraints:
     # Allow build names to contain alpha characters, underscores, and hyphens
     # but not special strings used for Nextstrain builds.
-    build_name = r'(?:[_a-zA-Z-](?!(tip-frequencies)))+',
+    build_name = r'(?:[a-zA-Z0-9-_](?!(tip-frequencies)))+',
     date = r"[0-9][0-9][0-9][0-9]-[0-9][0-9]-[0-9][0-9]",
     origin = r"[a-zA-Z0-9-_]+"
 
--- a/Snakefile
+++ b/Snakefile
@@ -131,11 +131,11 @@ config["conda_environment"] = CONDA_ENV_PATH
 
 # Define patterns we expect for wildcards.
 wildcard_constraints:
-    # Allow build names to contain alpha characters, underscores, and hyphens
+    # Allow build names to contain alphanumeric characters, underscores, and hyphens
     # but not special strings used for Nextstrain builds.
-    build_name = r'(?:[a-zA-Z0-9-_](?!(tip-frequencies)))+',
+    build_name = r'(?:[a-zA-Z0-9_-](?!(tip-frequencies)))+',
     date = r"[0-9][0-9][0-9][0-9]-[0-9][0-9]-[0-9][0-9]",
-    origin = r"[a-zA-Z0-9-_]+"
+    origin = r"[a-zA-Z0-9_-]+"
 
 localrules: clean
 
--- a/Snakefile
+++ b/Snakefile
@@ -133,9 +133,9 @@ config["conda_environment"] = CONDA_ENV_PATH
 wildcard_constraints:
     # Allow build names to contain alphanumeric characters, underscores, and hyphens
     # but not special strings used for Nextstrain builds.
-    build_name = r'(?:[a-zA-Z0-9_-](?!(tip-frequencies)))+',
-    date = r"[0-9][0-9][0-9][0-9]-[0-9][0-9]-[0-9][0-9]",
-    origin = r"[a-zA-Z0-9_-]+"
+    build_name = r'(?:[-a-zA-Z0-9_](?!(tip-frequencies)))+',
+    date = r"\d{4}-\d{2}-\d{2}",
+    origin = r"[-a-zA-Z0-9_]+"
 
 localrules: clean
 
--- a/workflow/snakemake_rules/export_for_nextstrain.smk
+++ b/workflow/snakemake_rules/export_for_nextstrain.smk
@@ -23,6 +23,8 @@ import requests
 import json
 from workflow.lib.persistent_dict import PersistentDict, NoSuchEntryError
 
+ruleorder: dated_json > finalize
+
 def get_todays_date():
     from datetime import datetime
     date = datetime.today().strftime('%Y-%m-%d')
@@ -135,6 +137,11 @@ rule dated_json:
         dated_tip_frequencies_json = "auspice/{prefix}_{build_name}_{date}_tip-frequencies.json"
     benchmark:
         "benchmarks/dated_json_{prefix}_{build_name}_{date}.txt"
+    wildcard_constraints:
+        # Allow build names to contain alphanumeric characters, underscores, and hyphens
+        # but not special strings used for Nextstrain builds.
+        build_name = r'(?:[a-zA-Z0-9-_](?!(tip-frequencies|\d{4}-\d{2}-\d{2})))+',
+        date = r"[0-9][0-9][0-9][0-9]-[0-9][0-9]-[0-9][0-9]"
     conda: config["conda_environment"]
     shell:
         """
--- a/workflow/snakemake_rules/export_for_nextstrain.smk
+++ b/workflow/snakemake_rules/export_for_nextstrain.smk
@@ -140,8 +140,8 @@ rule dated_json:
     wildcard_constraints:
         # Allow build names to contain alphanumeric characters, underscores, and hyphens
         # but not special strings used for Nextstrain builds.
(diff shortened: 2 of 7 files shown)


rneher:
@huddlej  -- I still get this error here:
```
\:> snakemake --profile nextstrain_profiles/nextstrain all_regions -n
Building DAG of jobs...
InputFunctionException in line 963 of /home/richard/Projects/nextstrain/ncov/workflow/snakemake_rules/main_workflow.smk:
KeyError: 'global_2021-01-10'
Wildcards:
build_name=global_2021-01-10
```

jacaravas:
Is there a reason this hasn't been merged yet?  Or did it just fall through the cracks?  I have had no issues using the code
    `build_name = r'(?:[_a-zA-Z0-9-](?!(tip-frequencies)))+',`
in the Snakefile "wildcard_constraints" for the last several months.

emmahodcroft:
I think this works fine for most people's builds, however, in the 'official' Nextstrain builds we also generate dated files - I think this is where the error comes in. (Unless something changed since Jan)

trvrb:
I was just testing out a `ncov/gisaid/global/6m` build and ran into this issue. I'd suggest bumping priority of addressing this.

jameshadfield (changes requested):
[ce0a763](https://github.com/nextstrain/ncov/pull/524/commits/ce0a7632936161e081c658075ec6e7e75501f988) is some good detective work @huddlej!

I've tested locally using snakemake 7.3.7 and can build a DAG without issue for the core nextstrain builds, however when changing the example profile to use a numeric build name I get an error:
```
## my_profiles/example/builds.yaml changed to include
builds:
    north-america_usa_washington_king-county_123: # name of the build; this can be anything
    
# error
Failed validating 'pattern' in schema['properties']['builds']['propertyNames']:
    OrderedDict([('pattern', '^(?:[_a-zA-Z-](?!(tip-frequencies)))+$')])
```

Updating `workflow/schemas/config.schema.yaml` lines 39-46 to be the following allowed things to work as expected:

```yaml
  builds:
    type: object
    minProperties: 1
    propertyNames:
      # Allow build names to contain alpha characters, underscores, and hyphens
      # but not special strings used for Nextstrain builds.  Also used in the
      # workflow's wildcard_constraints.
      pattern: "^(?:[a-zA-Z0-9-_](?!(tip-frequencies)))+$"
```

huddlej:
Thank you for testing this out, @jameshadfield! I've rebased onto master (to fix conflicts in the main workflow) and pushed commits to fix the schema validation and update the docs. I confirmed I could build DAGs for Nextstrain GISAID and open builds and have queued up [a trial run for open builds](https://github.com/nextstrain/ncov/runs/5996754585).

I also modified the `getting_started` profile's `builds.yaml` to name the build `global_6m` and confirmed that this built as expected with the both CLI/Docker and Snakemake/Conda. If the open builds work as expected, I'll plan to merge this.

workflow/snakemake_rules/export_for_nextstrain.smk
```diff
@@ -135,6 +137,11 @@ rule dated_json:
         dated_tip_frequencies_json = "auspice/{prefix}_{build_name}_{date}_tip-frequencies.json"
     benchmark:
         "benchmarks/dated_json_{prefix}_{build_name}_{date}.txt"
+    wildcard_constraints:
+        # Allow build names to contain alphanumeric characters, underscores, and hyphens
+        # but not special strings used for Nextstrain builds.
+        build_name = r'(?:[a-zA-Z0-9-_](?!(tip-frequencies|\d{4}-\d{2}-\d{2})))+',
+        date = r"[0-9][0-9][0-9][0-9]-[0-9][0-9]-[0-9][0-9]"
```
victorlin, line 144: Can we standardize between `\d{4}-\d{2}-\d{2}` and `[0-9][0-9][0-9][0-9]-[0-9][0-9]-[0-9][0-9]` here? They're equivalent regex patterns, would be nice to keep consistent.

Snakefile
```diff
@@ -131,11 +131,11 @@ config["conda_environment"] = CONDA_ENV_PATH
...
 # Define patterns we expect for wildcards.
 wildcard_constraints:
-    # Allow build names to contain alpha characters, underscores, and hyphens
+    # Allow build names to contain alphanumeric characters, underscores, and hyphens
     # but not special strings used for Nextstrain builds.
-    build_name = r'(?:[_a-zA-Z-](?!(tip-frequencies)))+',
+    build_name = r'(?:[a-zA-Z0-9_-](?!(tip-frequencies)))+',
     date = r"[0-9][0-9][0-9][0-9]-[0-9][0-9]-[0-9][0-9]",
-    origin = r"[a-zA-Z0-9-_]+"
+    origin = r"[a-zA-Z0-9_-]+"
```
victorlin, line 138: I had an opinion on hyphens at the end of regex sets here https://github.com/nextstrain/augur/pull/832#discussion_r793095816 but less of an opinion now, fine with this change.

workflow/snakemake_rules/export_for_nextstrain.smk
jameshadfield: Oh my! Strong vote for the `\d{4}-\d{2}-\d{2}` pattern 😂

Snakefile
huddlej: Oops, yes, I remember this now. I'm of the same mind as @tsibley about the convention of always placing hyphens at the end. Technically, we can always place them at the beginning instead and they will still get treated as literals but will be more "append-proof". Does that seem like a good compromise?

workflow/snakemake_rules/export_for_nextstrain.smk
huddlej: Agreed! Yes, I'll pick the less verbose option.

Snakefile
victorlin: Either way is fine to me. Though I don't think `-a-zA-Z0-9_` is much more "readable" than `a-zA-Z0-9-_`, so maybe it's best to have it at the end. I assume problems would be caught somewhere if we were to unintentionally append and make it a range like `_-~` in the future?

huddlej: Yeah, I would expect the readability to stem from the convention of first-or-last placement. I also don't expect us to update these too much, but it's good to have a convention within the team across projects that we're happy with. If you don't have a strong preference, I'll put the hyphen at the start to get the best of both worlds in terms of following convention and supporting appends to the end.

workflow/snakemake_rules/export_for_nextstrain.smk
huddlej: Fixed in 1b4beda

victorlin (approved):

workflow/snakemake_rules/export_for_nextstrain.smk
victorlin: I see, maybe another option would be `[0-9]{4}-[0-9]{2}-[0-9]{2}`?

tsibley: Yep!

PR #527, opened 2021-01-03
Clades update

trvrb:

This PR does a broad update of clades. Here, I used https://github.com/nextstrain/augur/blob/master/scripts/identify_emerging_clades.py to find clades that met either of the following frequency criteria (1 or 2) an I additionally labeled clades that met a third "variant of concern" criteria (3):
1. Greater than 20% global frequency for 2 or more months
2. Greater than 30% regional frequency for 2 or more months
3. VOC identified (applies to just 501Y.V1 and 501Y.V2)

This unifies regional and global clade naming so that `20A.EU1` becomes `20E` and consequently I eliminated region-specific subclades definitions. All clades are defined via `defaults/clades.tsv` and all emerging clades are defined via `defaults/subclades.tsv`.

### Testing
This has been pretty extensively tested on my laptop and through `nextstrain build`. This works for all regional targets.

Merging this requires following up with an update to https://github.com/nextstrain/nextclade with these new clade definitions and should have a blog post to pair with release.


rneher:
I added the 20A.EU1 back as subclade for the Europe build. I think it is better to keep it. It was never shown in any other build and I don't think it hurts keeping it for a while in the europe build.

trvrb:
Sounds good @rneher. In the above commit, I realized that we need to disambiguate between `20C/501T` (the spike mutation) and `20C/501T` (the nuc mutation) if we're going to have automated clade lookup. Thus, I think we need to specify "S" as in `20C/S.501T`. This change to specification has already been made in the draft blog post.

PR #528, opened 2021-01-04
Cleanup colors, outputs and description

trvrb:
This PR does a year-end cleanup of some cruft that's built up in the repo over the past year. Specifically, this PR:
1. Removes coloring by legacy Nextstrain clade (eg A2, B3, etc...) that was available as "Old Nextstrain Clade". In this labeling clade "A2" has been at >90% global frequency since June (about when we moved to the new system). This is no longer relevant to current builds and the legacy results can still be accessed via dated datasets.
2. Removes `/zh` build target. Early on we prioritized Chinese language translation at `/zh`. However, this has really not been relevant to the pandemic since April or May. This commit removes references to "zh" and removes this from the target list.
3. Removes `/gisaid` build target. GISAID maintains an iFrame linking to our latest build here: https://www.gisaid.org/epiflu-applications/phylodynamics/. It was originally desired to have some customization (enforce closed sidebar, no top-matter) and so we made a specific build target to accomplish this and kept this target at `/gisaid`. Subsequently Auspice has been updated to allow these customizations to be performed via URL query params. As such, we can now remove this build target to streamline workflow.
4. Revise footer description. This includes linking to `/sars-cov-2`, update subsampling description, improve region-specific links and other small fixes.

### Testing
I've tested this locally to ensure that the build completes.

This merge needs to wait on GISAID updating their iFrame link. I've emailed GISAID about this and hoping this change will be made soon. Additionally, when this is merged we can update Auspice (cc @jameshadfield) to remove the special casing of `/gisaid` in the URL path triggering sidebar to be closed (https://github.com/nextstrain/auspice/blob/master/src/reducers/controls.js#L285).


trvrb:
GISAID has kindly updated the iFrame link at https://www.gisaid.org/epiflu-applications/phylodynamics/. I believe this can be merged immediately at this point.

PR #529, opened 2021-01-06
remove dubious ORFs

rneher:
There is little evidenve that ORF14 and ORF10 matter. 

https://www.nature.com/articles/s41586-020-2739-1
"We also did not find evidence of translation of ORF14, which appears in some SARS-CoV-2 annotations"

"The SARS-CoV-2 ORF10 is not essential in vitro or in vivo in humans"
https://journals.plos.org/plospathogens/article?id=10.1371/journal.ppat.1008959


rneher:
@trvrb: any reason why these ORFs should stay? I think the tree/tooltips would be less cluttered with those removed. And given that they likely don't matter, I don't think we loose much.

PR #530, opened 2021-01-06
Add slack notification

rneher:
This copies the script to notify slack from ingest. I think the env vars are set the same way, so this might just work.


eharkins:
@rneher where does the script to notify slack get called / where are you hoping it will be run from?

rneher:
This is called from here:
https://github.com/nextstrain/ncov/blob/master/scripts/upload-to-s3

PR #531, opened 2021-01-07
Default to clades coloring

trvrb:

This moves default from geographic coloring to clades coloring for global and regional builds. I think this is in order at this point given necessary focus on expanding variants of concern.

### Testing
This was tested locally, but there's not a lot potential for errors.


emmahodcroft:
This makes sense to me. I think the geographic colouring was becoming a bit less helpful, and now we have good clades, I think it makes sense to highlight them!

emmahodcroft (approved):

PR #532, opened 2021-01-09
Provide an explicit max date for tip frequencies

huddlej:
## Description of changes

Prevent tip frequency pivots from extending too far into the future when no maximum date is provided by providing an explicit max date. Users can define the `max_date` in the frequencies section of the configuration (as with `min_date`) or leave this field empty. When the `max_date` is not defined by the user, it defaults to today's date.

This commit copies the `numeric_date` utility from TreeTime into the common functions for the workflow. The cost of duplicating this code is worth the benefit of not requiring Augur or TreeTime to be installed in the user's environment when they run Snakemake. This allows users to run the workflow with `--use-conda` from an environment without anything but Snakemake installed.

## Testing

Tested by with the getting started profile, to make sure that the correct date is used as an argument and that the frequency pivots do not extend beyond the maximum date.

### DIFF
--- a/workflow/snakemake_rules/common.smk
+++ b/workflow/snakemake_rules/common.smk
@@ -1,5 +1,30 @@
 """Small, shared functions used to generate inputs and parameters.
 """
+import datetime
+
+
+def numeric_date(dt=None):
+    """
+    Convert datetime object to the numeric date.
+    The numeric date format is YYYY.F, where F is the fraction of the year passed
+    Parameters
+    ----------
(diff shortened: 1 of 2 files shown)


trvrb:
I like this design of taking today's date unless something is specifically supplied as `config["frequencies"]["max_date"]`.

trvrb (approved):

trvrb:
This looks great to me. Tested locally with latest build from AWS Batch and everything works as it should. I believe this can be merged.

PR #533, opened 2021-01-13
Set weeks as pivot interval

trvrb:
### Description of proposed changes    

This PR updates `augur frequencies` with `--pivot-interval-units weeks` to give finer resolution pivots. Merging this should wait on the merge of Augur PR https://github.com/nextstrain/augur/pull/660 and the release of a new Augur version (so that AWS Batch has Augur change available).

PR #535, opened 2021-01-15
Updates to augur 10.3.0

cassiawag:
Updates `nextstrain.yaml` to specify the latest version of augur. Otherwise augur frequency fails with`use-conda` after weeks as the pivot interval (#533) was merged in.

Noticed this bug and tested it on the WA builds.


trvrb:
I thought this would need updating somewhere, but wasn't sure where in the repo this was stored. I'm going to go ahead and merge. Thanks Cassia!

PR #537, opened 2021-01-22
Add a "dump_config" rule to spew the final, merged config to stdout

tsibley:
Helpful for development, testing, and debugging of config settings when
layers of profiles/config files and command-line options are involved.

yaml is PyYAML, which is a transitive dep through Snakemake.

Even in --quiet mode, Snakemake prints cruft to stdout, so the explicit
start and end markers for the YAML document help delineate it.

### DIFF
--- a/Snakefile
+++ b/Snakefile
@@ -99,6 +99,11 @@ rule clean:
     shell:
         "rm -rfv {params}"
 
+rule dump_config:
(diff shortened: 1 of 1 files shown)


rneher:
I think this is useful. Thanks @tsibley!

PR #540, opened 2021-01-27
Filter out samples pre-screened for S dropout

trvrb:
Including "S dropout" samples biases prevalence of B.1.1.7 upwards. Labeling of samples in GISAID seems to be happening best in the US, but I think helpful to have this included in general.

This additional filtering is included in the Nextstrain profile rather than the default parameters because I imagine other groups will want to look specifically at trees of B.1.1.7.

Here, we see that prior to this change, we are listing 501Y.V1 in the US at 9% frequency at the most recent pivot:

uncorrected

while with this change, we list 501Y.V1 in the US at 6% frequency:

corrected

So, not a huge effect, but a simple step in the right direction.

### Testing
I did a full build locally via `snakemake` with this change in place.

### DIFF
--- a/nextstrain_profiles/nextstrain/builds.yaml
+++ b/nextstrain_profiles/nextstrain/builds.yaml
@@ -40,6 +40,10 @@ builds:
     region: South America
     auspice_config: "nextstrain_profiles/nextstrain/south-america_auspice_config.json"
 
+# remove S dropout sequences and sequences without division label in US
+filter:
+  exclude_where: "division='USA' purpose_of_sequencing='S dropout'"
+
 # if different exposure traits should be used for some builds, specify here
 # otherwise the default exposure in defaults/parameters.yaml will used
 exposure:


trvrb:
Hi @rneher and @emmahodcroft. Are you okay with this change? If we look at https://nextstrain.org/ncov/north-america?f_country=USA at the moment, we are placing US frequency of 501Y.V1 at 14% on Jan 8.

However, a more careful accounting of frequency that removes samples screened for S dropout estimates ~4% on Jan 8.

usa-prevalence

This is a big difference and I'd like to have the improved accuracy here.

emmahodcroft:
I think this is likely a good idea. I agree with just including in the Nextstrain profile so that others can make this decision separately. I also just think we need to be aware that essentially only the USA is providing this info at the moment so for other countries, estimates are skewed.
(I wish we had something like SPHERES for the EU so we could similarly at least informally suggest using such conventions!)

trvrb:
Thanks Emma! Yeah, this is an important caveat. This won't correct everything, but I think should be a step in the right direction. I'm going to go ahead and merge this now.

Issue #225, opened 2020-03-20
Metadata extension


t-neumann:
Hi,

I'm curious on how the `metadata.tsv` files is created - could you elaborate on that?

I don't see any fields commenting on sequencing platform used, assembly method and coverage, yet I feel like this would be also valuable information. While I also do not see these fields listed when browsing GISAID, I see that this information is at least optionally supplied when sequences are uploaded.

![image](https://user-images.githubusercontent.com/17272623/77139595-e04c3f80-6a76-11ea-8193-b9784d8d706d.png)

Is there a way to get to this?

tolot27:
I have to filter sequences dynamically based on coverage and sequencing platform (filter out low coverage MinION) and need this information as well.

oneillkza:
+1 to this. As far as I can tell, it involves running this tool: https://github.com/nextstrain/fauna

But I still don't see how they're getting it in the first place. I'm fairly sure they must be scraping it from GISAID, since the metadata only seems to appear when clicking on individual pages for each sequence.

trvrb:
I'm leaving this open. It would be nice to add a 1-sentence link to `ncov-ingest` from https://docs.nextstrain.org/en/latest/tutorials/SARS-CoV-2/steps/data-prep.html#background-contextual-sequences to just point out what's going on here. Though not at all urgent. cc @eharkins

eharkins:
Closed by #541

PR #541, opened 2021-01-27
#225 link to ingest from ncov tutorial

eharkins:
Addressing question about where metadata comes from by linking to ncov-ingest pipeline. cc @trvrb let me know if this language is what you were looking for. I left out this part since it's in the ncov-ingest README.md's first line:
> This is open source, but we are not intending to support this to be used by outside groups.


trvrb:
This looks good to me. Thanks @eharkins!

PR #542, opened 2021-01-28
Allow an optional split between S3 source and destination buckets

tsibley:
This lets trial runs pull the latest data from the production bucket but
upload back into non-production buckets so as not to affect our
production runs.


eharkins:
@tsibley this makes sense to me 👍

Issue #546, opened 2021-02-04
reserve more ram for priorities.py by default in snakefile


dpark01:
**Current Behavior**
When running a build with proximity-based sampling on modern day inputs (e.g. alignments with >400k genomes), I often find I need to feed much more RAM to the priorities.py calculations than what the main snakemake rules are reserving. If I have a build definition with more than one proximity score to calculate, the default ncov snakemake workflow will allow multiple proximity_score rules to run simultaneously (believing that they only takes 4GB of RAM each), resulting in the priorities.py processes being killed for requiring too much RAM.

**Expected behavior**
Given available RAM resources (assuming snakemake is told about them), the ncov workflow should not attempt to run more simultaneous priorities.py processes than the compute environment can handle.

**How to reproduce**
Steps to reproduce the current behavior:
1. Spin up a VM with 32GB RAM, pull the nextstrain/base docker image and ncov github repo
2. Pull a nextfasta/nextmeta data set with at least 400k ncov genomes.
3. As an example, pull the builds.yaml from https://github.com/CDCgov/spheres-augur-build/blob/master/spheres_profile/builds.yaml
4. Place an aligned, filtered, masked MSA in ncov/results/filered.fasta (skip all the upstream rules in the dag)
5. run `snakemake -j $(nproc) --resources mem_mb=32000 --profile <profile dir> results/california/subsampled_alignment.fasta` (skip all the downstream rules in the dag: this should just run a bunch of subsample and proximity_score rules
6. Watch as snakemake tries to run multiple proximity_score rules simultaneously. If your compute environment kills processes that exceed available RAM, priorities.py will be killed.
7. Fake it with `snakemake` with `--resources mem_mb=4000` and they will be run serially

**Possible solution**
The ncov snakemake should allot more [mem_mb to proximity_score](https://github.com/nextstrain/ncov/blob/master/workflow/snakemake_rules/main_workflow.smk#L435). In my observation, it should be increased to about 24GB per invocation to handle modern data sets, but given the growth rate of global data, this may need some headroom built in or dynamic calculation based on input size.

**Your environment: if running Nextstrain locally**
 - Operating system: `nextstrain/base:build-20210127T135203Z`
 - Version (e.g. `auspice 2.7.0`): github.com/nextstrain/ncov 5dbca8a45a64e39057c22163f154db981f7ed5c1

dpark01:
[...]

PR #548, opened 2021-02-04
Start collecting benchmark information for the proximity_score rule

tsibley:
This will let us see how much memory the step actually uses, so we have
some additional insight into adjusting the mem_mb resource declaration.

See also <https://github.com/nextstrain/ncov/issues/546>.


huddlej (approved):
This worked for me with the following command:

```
nextstrain build . all_regions -j 2 --profile nextstrain_profiles/nextstrain --config sequences=data/example_sequences.fasta metadata=data/example_metadata.tsv active_builds=europe
```

I forgot (or didn't realize) that Snakemake's memory profiling doesn't work on Mac, though, so the benchmarks were empty when I ran with straight-up `snakemake` instead of the CLI.

PR #549, opened 2021-02-05
Allow multiple input files

jameshadfield:
~This PR is an updated version of #519, but only up until the filtering stage (i.e. no subsampling yet).~ The general idea is to maintain backwards compatability with single-input workflows as well as allowing multiple inputs, each with their own filtering parameters etc, by specifying a dictionary of input files.

Inside the Snakemake rules, we use a wildcard `origin` to define different starting points. For instance, if we ask for the file `results/aligned_seqRun42.fasta` then `wildcards.origin="_seqRun42"` and we expect that the config has defined a sequences input via `config["sequences"]["seqRun42"]=<path to fasta>` (note the leading `_` has been stripped). If there's only one starting point, then this wildcard is empty. For instance, asking for `results/aligned.fasta` results in `wildcards.origin=""` and we expect `config["sequences"]=<path to fasta>`.

### Testing

These changes shouldn't change any currently existing (single-input) workflows & if people could verify this that would be really helpful!

### Snakemake questions

See in-line comments on this PR. 

### Running tutorial with multiple inputs:

Currently the example only works until the `filter` stage. Documentation is included in this PR at `docs/multiple_inputs.md`. But it should be runnable via:

```sh
tar xf data/example_multiple_inputs.tar.xz --directory data/
snakemake --profile my_profiles/example_multiple_inputs -f results/filtered_input1.fasta results/filtered_input2.fasta
snakemake --cores 4 --profile my_profiles/example -f results/filtered.fasta # single-input example should still work!
```

image

_Left: `example` profile, Right: `example_multiple_inputs`. Note how `input1` skips the diagnostic & refiltering steps_


scripts/combine_metadata.py
```diff
@@ -0,0 +1,79 @@
+import argparse
+from augur.utils import read_metadata
+from Bio import SeqIO
+import csv
+
+EMPTY = ''
+
+# This script was written in preparation for a future augur where commands
```
jameshadfield, line 8: Want to emphasise that as soon as `augur ...` can take multiple metadata inputs then this script can go away :)

workflow/snakemake_rules/main_workflow.smk
```diff
@@ -1,7 +1,46 @@
...
+        return f"results/aligned{wildcards.origin}.fasta"
+    return f"results/aligned-filtered{wildcards.origin}.fasta"
+
+
+
 rule download_sequences:
     message: "Downloading sequences from S3 bucket {params.s3_bucket}"
     output:
-        sequences = config["sequences"]
+        sequences = config["default_sequences"]
```
jameshadfield, line 43: My biggest question / unknown at the moment is how to allow specifig inputs (e.g. based on the `origin` wildcard) to download filew from S3 -- both here and in the `download_preprocessed.smk`.

This specific solution is not great - if one of the input meadata / sequences files is `sequences.fasta` or `metadata.tsv` then it can be downloaded from AWS (because it matches the defaults). This allows our runs to continue as normal, but would break the situation where someone defined their own `config.sequences` or `config.metadata` filenames _and_ used the AWS download functionality.

In the case of ``download_preprocessed.smk`, if included, for single-input runs (`wildcards.origin=""`) these download rules will run (which is great) but I want to somehow allow _one_ of the multi-inputs to use these rules. In other words, make snakemake aware that one of the input FASTAs can be downloaded from S3 instead of being computed locally.

workflow/snakemake_rules/main_workflow.smk
```diff
@@ -27,20 +66,42 @@ rule download:
...
+    log:
+        "logs/combine_input_metadata.txt"
+    conda: config["conda_environment"]
+    shell:
+        """
+        python3 scripts/combine_metadata.py --input {input.metadata} --output {output.metadata} 2>&1 | tee {log}
+        """
+
+def _get_unified_metadata(wildcards):
+    # TODO - there's another metadata file getter function!
```
jameshadfield, line 87: reminder for me

jameshadfield:
This PR now works for subsampling steps as well -- see the updated tutorial for instructions.

The biggest remaining issue is that you cannot download preprocessed files when there are multiple inputs. I can think of some rather hacky ways of doing so (defining input functions which look at config settings to choose which rule to run for which input) but nothing feels right yet. Suggestions welcome!

docs/multiple_inputs.md
```diff
@@ -0,0 +1,166 @@
...
+```yaml
+builds:
+  multiple-inputs:
+    subsampling_scheme: custom-scheme # use a custom subsampling scheme defined below
+
+subsampling:
+  custom-scheme:
+    # Use metadata key to include ALL from `input1`
+    allFromAus:
+      exclude: "--exclude-where 'aus=no'" # subset to sequences from input `aus`
```
trvrb, line 140: I might have expected a single new column of `input-source` with fields of `aus` vs `worldwide` rather than the "one-hot" encoding that is used here.

workflow/snakemake_rules/main_workflow.smk
trvrb: I don't think I mind special casing `sequences.fasta` and `metadata.tsv`. As long as this is well documented I don't think it's a bad pattern.

jameshadfield: I think the solution in #550 achieves the desired goal in a much nicer way than here.

P.S. The downloaded sequences file isn't special cased to `sequences.fasta` but rather it will download to `config["sequences"]` if that file isn't available.

docs/multiple_inputs.md
jameshadfield: I went with one-hot as I often have the use case where the sequence sets are overlapping -- e.g. some of my "private" sequences are _also_ in GISAID / NCBI, and I couldn't think of how to nicely express this in a single variable whilst facilitating subsampling based on this variable...

huddlej (commented):
This is excellent, @jameshadfield! I really like the direction of allowing multiple inputs with their own pre-filter step followed by custom subsampling and then a merging prior to analysis.

My biggest worry with this approach has more to do with the current complexity of the workflow with diagnostics, way too many filtering steps, optional downloads from S3, and optional application of travel history calculations. Maybe these are unavoidable complications, but I would love to get the workflow back to a simpler flow from raw data to a single filter, subsample, concat, and then tree building approach.

Specific comments in the "I like, I wish, and what if" style follow.

## I like that...

* you took the “document-driven design” approach here by developing the docs as you implemented
* sequences and metadata are named datasets (e.g., `aus`) and have entries in the config schema document
* it is possible to skip steps in the workflow with config options. I wonder if we could make this interface more generic with a top-level `skip_steps` key?
* filters can be applied to specific datasets by name (e.g., the `aus` custom filters). Should we consider a different name for these filters to avoid conflicts with defaults? These could be named `dataset_filters`, at the risk of adding too many parameters but in parallel with the proposed `datasets` key below.

## I wish that...

* `seq_per_group` did not need to be defined for the “all from Aus” use case. Maybe we can make this parameter optional? Ditto `group_by`. We need to allow users to more easily “subsample” all sequences.
* the workflow didn’t require a wildcard with an empty or an underscore prefix. We tend to reserve underscores as special delimiters in our filenames and expect our wildcard values to not include delimiters. I’m mostly worried about code like `return config['metadata'][wildcards.origin[1:]]` where the reader needs to know that the origin is prefixed by an underscore that has to be removed. I think this is a minor technical detail we can address though.
* the new helper functions in `main_workflow.smk` lived in the `common.smk` file instead. 

## What if...

* sequences and metadata were bundled more tightly into named datasets like the following such that updates within a dataset are more atomic (modify a single record in one data structure instead of remembering to modify separate data structures):

```yaml
# Keep these as top-level, backwards-compatible defaults.
sequences: "data/sequences.fasta"
metadata: "data/metadata.tsv"

# Augment defaults with more custom datasets.
datasets:
  aus:
    sequences: "data/example_sequences_aus.fasta"
    metadata: "data/example_metadata_aus.tsv"
  worldwide:
    sequences: "data/example_sequences_worldwide.fasta"
    metadata: "data/example_metadata_worldwide.tsv"
```

This approach would also allow us to check for definition of a top-level key instead of checking the type of the existing `sequences` and `metadata` keys. (Nevermind: I see that you propose this in https://github.com/nextstrain/ncov/pull/550.)

* we move away from negative exclude logic (e.g., `--exclude-where 'aus=no'`) toward positive query logic (e.g., `--query "aus == 'yes'"`)

jameshadfield:
Thanks @huddlej for the review!

> My biggest worry with this approach has more to do with the current complexity of the workflow with diagnostic

Agreed. This was the first time in a few months I've dived into nCoV and have a few ideas to improve things. The general pipeline (even with multiple inputs) can be broken into essentially 3 sequential stages (see below) so I'm hopeful we can make improvements here. For this PR I didn't want to play too much with this aspect of the pipeline.

1. Align, filter etc (inputs done in parallel) -> combined alignment
2. Subsample (can get complex!)
3. Tree + decoration

> [I wish that] the workflow didn’t require a wildcard with an empty or an underscore prefix. 

Yes -- this did feel a bit hacky. Mainly there to maintain backwards compat, but I realise that it's confusing (even to me) that `origin` sometimes includes a `_` and sometimes it's trimmed. 

> [I wish that] the new helper functions in main_workflow.smk lived in the common.smk file instead.

Done on #550 👍 

> [What if] sequences and metadata were bundled more tightly into named datasets

Glad you came up with this independently! I like the interface of #550. In that vein, I'm going to close this PR and move your other comments from this review over to that PR

PR #552, opened 2021-02-11
Use sequence index for augur filter rules

huddlej:
Adds a rule to generate an initial sequence index with the new augur index command and updates the workflow's augur filter rules to use this index.

Note that this implementation requires the download of the full sequence set even if the user has requested to start the workflow from the filtered alignment (since subsampling also needs the sequence index). If we chose to recalculate the sequence index for the filtered alignment, the contents of that new index would be different as it would represent a multiple sequence alignment with additional gap sequences.

### DIFF
--- a/nextstrain_profiles/nextstrain/builds.yaml
+++ b/nextstrain_profiles/nextstrain/builds.yaml
@@ -26,6 +26,7 @@ inputs:
     # ------------------------------------------------------
     # metadata: "s3://nextstrain-ncov-private/metadata_gisaid.tsv.gz"
     # sequences: "s3://nextstrain-ncov-private/sequences_gisaid.fasta.gz"
+    # sequence_index: "s3://nextstrain-ncov-private/sequence_index_gisaid.tsv"
     # aligned: "s3://nextstrain-ncov-private/aligned_gisaid.fasta.xz"
     # to-exclude: "s3://nextstrain-ncov-private/to-exclude_gisaid.txt.xz"
     # masked: "s3://nextstrain-ncov-private/masked_gisaid.fasta.fasta.xz"
--- a/workflow/envs/nextstrain.yaml
+++ b/workflow/envs/nextstrain.yaml
@@ -18,6 +18,6 @@ dependencies:
 - pip
 - pip:
   - awscli==1.18.45
-  - nextstrain-augur==10.3.0
+  - nextstrain-augur==11.1.2
   - nextstrain-cli==1.16.5
   - rethinkdb==2.3.0.post6
--- a/workflow/snakemake_rules/common.smk
+++ b/workflow/snakemake_rules/common.smk
@@ -110,6 +110,14 @@ def _get_unified_alignment(wildcards):
         return _get_path_for_input("filtered", "_"+list(config["inputs"].keys())[0])
     return "results/combined_sequences_for_subsampling.fasta",
 
+def _get_unified_sequence_index(wildcards):
+    if not config.get("inputs"):
+        return f"results/sequence_index.tsv"
+    elif len(list(config["inputs"].keys())) == 1:
+        return f"results/sequence_index{wildcards.origin}.tsv"
+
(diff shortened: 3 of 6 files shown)


cassiawag:
I tested this today on a branch with the [WA builds](https://github.com/blab/ncov-wa-build), and it ran through smoothly and sped up the build. 

To avoid downloading sequences.fasta, which is a nontrivial sized file that will only grow, you could calculate  "sequence_index.tsv" add it to the Nextstrain bucket. Then you could  add a a `rule download_sequence_index` to override `rule index_sequence` to the [download_preprocessed.smk](https://github.com/nextstrain/ncov/blob/master/nextstrain_profiles/nextstrain/download_preprocessed.smk).

cassiawag (commented):
I tested this today on a branch with the [WA builds](https://github.com/blab/ncov-wa-build), and it ran through smoothly and sped up the build. 

To avoid downloading sequences.fasta, which is a nontrivial sized file that will only grow, you could calculate  "sequence_index.tsv" and add it to the Nextstrain bucket. Then you could  add a `rule download_sequence_index` to override `rule index_sequence` to the [download_preprocessed.smk](https://github.com/nextstrain/ncov/blob/master/nextstrain_profiles/nextstrain/download_preprocessed.smk).

huddlej:
Thank you for testing this out, @cassiawag! I will look into adding the sequence index to the S3 bucket tomorrow.

rneher:
I guess there are a few conflict now that need resolving after the `multiple-downloads` PR was merged. Do we want to filter each input? or the combined fasta?

I have also made some progress with redoing the priorities based on Benjamin's work. This using the `index`

https://github.com/nextstrain/ncov/blob/redo-priorities/scripts/priorities.py

and should reduce memory requirements by a lot (the old one was routinely using 40Gb of memory)

jameshadfield:
Haven't tested this PR, but we'll have to index each input's `sequence` file for the `filter` step, and then _if_ we're using multiple inputs create a new index (or combine the indexes?) for the `subsampling` steps which use a unified alignment. 

The downloads do get a bit trickier, as in many cases we don't want to download the unaligned sequences (e.g. we want to start from a filtered alignment), but this should be possible by downloading the index as well.

Really important PR - thanks @huddlej 💯

rneher:
with the updated augur on aws, we are actually creating about 30 indices... each subsampling step creates its own index. so would be good to get this PR in. I imagine we could stick an index step after we combine inputs and then use it for all downstream operations.

huddlej:
I rebased this PR onto the new multiple inputs code on master. The workflow is pretty complicated right now and I'm not sure I caught all of the edge cases where the sequence index could be pulled in incorrectly. I have tested this code locally with the `getting_started`, `example_multiple_inputs`, and `nextstrain` profiles and the DAGs all look reasonable.

I also added support for uploading and downloading a precomputed sequence index to/from S3, but I haven't been able to test this.

@rneher Do these S3 uploads happen as part of the build shepard process or do you manually maintain those uploads yourself?

rneher:
the S3 upload happen by default (the workflow has targets `upload` and `all_regions`)

rneher:
I made a draft PR 

https://github.com/nextstrain/ncov/pull/561

that shows how indexing could work in the multi-download-nextalign branch. The only index that is reused is the one of the combined sequences. all others are touched only once can be made on the fly...

eharkins:
Have we run a full trial build (https://github.com/nextstrain/ncov/actions/workflows/trial-build.yml) for this and / or #561 ? Would that be helpful in testing this further?

huddlej:
That's a great idea, @eharkins. We'll probably close this PR in favor of @rneher's [simpler PR based on the new nextalign workflow](https://github.com/nextstrain/ncov/pull/561). We should probably focus all of our efforts on testing and refining that PR instead of this one.

huddlej:
This PR was superseded by #562.

PR #554, opened 2021-02-16
Don't filter samples collected today

### LINKED ISSUE #553: Sequences sampled today are incorrectly removed (filtered)
**Current Behavior**
Sequences with a sampling date the same as the analysis day are incorrectly filtered by `rule filter`

**Expected behavior**
They shouldn't be removed.

**Possible solution**

Short term: add one day to the code here: https://github.com/nextstrain/ncov/blob/master/workflow/snakemake_rules/main_workflow.smk#L263 to allow 

Long term: Change `augur filter` to keep sequences with the same date as the provided `--max-date` argument.

jameshadfield:
Closes #553 by making the argument supplied to `augur filter`'s `--max-date` represent tomorrow, not today!

### DIFF
--- a/workflow/snakemake_rules/main_workflow.smk
+++ b/workflow/snakemake_rules/main_workflow.smk
@@ -266,7 +266,7 @@ rule filter:
         exclude_where = lambda wildcards: _get_filter_value(wildcards, "exclude_where"),
         min_date = lambda wildcards: _get_filter_value(wildcards, "min_date"),
         ambiguous = lambda wildcards: f"--exclude-ambiguous-dates-by {_get_filter_value(wildcards, 'exclude_ambiguous_dates_by')}" if _get_filter_value(wildcards, "exclude_ambiguous_dates_by") else "",
-        date = date.today().strftime("%Y-%m-%d")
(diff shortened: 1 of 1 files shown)


huddlej (commented):
This solution looks good.

Can we just remove the max date filter entirely though? No max date filter should be the same as allowing all records up through "today". Was there a use case for wanting to exclude records from the future? For example, is this how we currently exclude records with invalid dates?

jameshadfield:
> Was there a use case for wanting to exclude records from the future? 

I believe this has come up in the past (cc @emmahodcroft & @lmoncla if I recall corectly). Either way, I think it's reasonable to exclude sequences sampled in the future, as these will be annotation errors & cause issues for TimeTree.

emmahodcroft:
Yes it's super handy for us to exclude dates in the future as these are obviously misdated. Luckily because the mistake is usually pretty obvious (I actually try to reach out anytime I see this), they're fixed pretty quickly. If so, when the new date comes in from GISAID it means it automatically gets included again, with nobody having to remember to add/remove from `exclude.txt`. (Of course before this, it's automatically excluded so requires no action from our end except to alert the submitters). 

It's really hard to keep up with when things should be removed from `exclude.txt` so as much as possible that can happen automatically there is super useful!

PR #555, opened 2021-02-17
Multiple downloads nextalign

rneher:
incorporate nextalign and streamline the ncov workflow...

This PR is on top of @jameshadfield  PR #550  and optionally uses nextalign for alignment. It also removes the `prefilter` and `refilter` steps which each take about 30min.


rneher:
@jameshadfield Does this work for you? This should be quite a bit faster since it does two filter steps less and reduce 32Gb less disk space. And aligning should be faster as well...

jameshadfield:
Looks awesome - cherry-picked these commits onto the new `master` branch in PR #558 and am currently testing on a full nCoV build.

PR #558, opened 2021-02-19
Nextalign

jameshadfield:
This is @rneher's PR #555 cherry-picked onto `master`. Commit 8140454  wasn't used, but was incoporated in 
b9968bc which also removed other code references to now-removed rules. 

---

Tested locally and the speed-ups of nextalign are amazing (🙏 @ivan-aksamentov). The new DAG is _much_ simpler. My only real addition over #555 is to re-enable the ability to skip the diagnostic step on a per-input basis (detailed in the multiple-inputs tutorial). 

Currently running a multiple-input nCoV build on AWS ~and will report back with results.~ which worked well 💯 . We are now in the outrageous position where aligning takes ~25min and each filtering step takes ~37min!


rneher:
Thanks, @jameshadfield. I'll add a few more things to PR

 * translating a genes
 * `mutation_summary.tsv`
 * explicit translations to avoid having these `68X` mutations

PR #560, opened 2021-02-20
Allow filtering by PANGO lineage

trvrb:
This is a simple swap from `authors` to `pangolin_lineage` in the Nextstrain profile filters. I've found it increasingly common to want to point people to trees highlighting B.1.525, etc... and this accomplishes this easily. 

I've chosen to swap out `authors` rather than just add `pangolin_lineage` just because the page footer was seeming over verbose. If folks think that `authors` is required we should keep it in. We'll still have originating lab, submitting lab and authors in the tooltip.

### Testing
I applied this on top of the latest build artifacts from AWS and it worked just fine. (There's not much there to break)

### DIFF
--- a/nextstrain_profiles/nextstrain/africa_auspice_config.json
+++ b/nextstrain_profiles/nextstrain/africa_auspice_config.json
@@ -121,7 +121,7 @@
     "division",
     "location",
     "host",
-    "author"
+    "pangolin_lineage"
   ],
   "panels": [
     "tree",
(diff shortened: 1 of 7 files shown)


emmahodcroft:
I think this is a really sensible idea Trevor! I haven't filtered by author in months really, it's become too diverse. On the other hand there's definitely been times I would have used a pango lineage search. So I think this us a good replacement to make.

emmahodcroft (approved):

PR #561, opened 2021-02-20
Nextalign rebased index

rneher:
simplified indexing after combining data sets


jameshadfield (approved):
This looks good to me -- added a one-line fix as this wasn't working for builds without nextalign (see commit msg for why). 

Reducing the number of pre-subsampling filtering steps to one-per-input helps a lot, as we can simply rely on `augur filter` to automatically index the fasta rather than add more rules which would be needed if we had multiple filters. 

There are a number of nice improvements here, including explicit files detailing insertions and gene (aa) alignments.

There were two points which confused me slightly:

1. Why are we running `rule translate` _and_ `rule aa_muts_explicit`? Would it be better to expand the number of genes that the latter runs on, and use that instead of `translate` (in the case where nextalign is available)?

2. I didn't quite get the moving of rules into `core_workflow.smk`, which seems to never be imported by any workflows?!?

I've tested using the example workflow with & without `nextalign` and it's working well.

huddlej:
Can we close the older PRs now (https://github.com/nextstrain/ncov/pull/558 and https://github.com/nextstrain/ncov/pull/561)?

PR #562, opened 2021-02-22
Nextalign rebased index seq pr

### LINKED ISSUE #546: reserve more ram for priorities.py by default in snakefile

rneher:
Split the priorities calculation into a calculation of proximity (in batches to save memory) and priorities. 

this should make the calculation of priorities more flexible and more transparent and limit the memory requirements to a few Gb instead of 50Gb. 

### Related issue(s)  
Fixes #546


nextstrain_profiles/nextstrain-scicore/cluster.json
```diff
@@ -20,12 +20,16 @@
...
   "align": {
-    "n": 16,
-    "mem": "128G",
+    "n": 8,
+    "mem": "16G",
     "qos": "6hours",
     "time": "05:59:00"
   },
   "proximity_score": {
     "mem": "64G"
```
huddlej, line 29: Can this be reduced to `2G` now?

scripts/get_distance_to_focal_set.py
```diff
@@ -0,0 +1,183 @@
...
+    focal_seqs_dict = calculate_snp_matrix(focal_seqs, consensus = ref)
+
+    fh_seqs = open(args.alignment, 'rt')
+    seqs = SimpleFastaParser(fh_seqs)
+
+    # export priorities
+    fh_out = open(args.output, 'w')
+    fh_out.write('strain\tclosest strain\tdistance\n')
+
+    chunk_size=10000
```
huddlej, line 154: This is excellent! Would parameterizing this variable in the script's command line arguments effectively allow users to tune memory required by this script? That could be helpful for folks who may have different memory constraints.

scripts/get_distance_to_focal_set.py
```diff
@@ -0,0 +1,183 @@
+"""
+Mask initial bases from alignment FASTA
```
huddlej, line 2: Can we delete this old docstring or update it to reflect this script's function?

scripts/explicit_translation.py
```diff
@@ -0,0 +1,43 @@
+
```
huddlej, line 1: This is also a really useful script. Do you see this approach eventually replacing the current augur translate approach? We could update augur translate to alternately accept existing translations instead of inferred ancestral nucleotide sequences...

huddlej (commented):
I'm still testing this on our SLURM cluster, so I can't comment on whether everything works or not. Inline comments below cover minor or prospective discussion points.

workflow/snakemake_rules/main_workflow.smk
```diff
@@ -96,60 +96,72 @@ rule diagnose_excluded:
...
+            genes = ','.join(config.get('genes', ['S'])),
+            basename = "seqs{origin}"
+        log:
+            "logs/align{origin}.txt"
+        benchmark:
+            "benchmarks/align{origin}.txt"
+        threads: 8
+        shell:
+            """
+            {params.bin} \
```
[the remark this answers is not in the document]
rneher: this is more a historical hangover

workflow/snakemake_rules/main_workflow.smk
```diff
@@ -96,60 +96,72 @@ rule diagnose_excluded:
...
-        """
-        mafft \
-            --auto \
-            --thread {threads} \
-            --keeplength \
-            --addfragments \
-            {input.sequences} \
-            {input.reference} > {output} 2> {log}
-        """
+if "use_nextalign" in config and config["use_nextalign"]:
```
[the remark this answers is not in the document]
rneher: I'd keep it for know and work towards handling this as a method in augur align (along with `--addframents` for mafft)

workflow/snakemake_rules/main_workflow.smk
```diff
@@ -96,60 +96,72 @@ rule diagnose_excluded:
...
+        params:
+            outdir = "results/translations",
+            bin = config["nextalign_bin"],
+            genes = ','.join(config.get('genes', ['S'])),
+            basename = "seqs{origin}"
+        log:
+            "logs/align{origin}.txt"
+        benchmark:
+            "benchmarks/align{origin}.txt"
+        threads: 8
```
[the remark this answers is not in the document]
rneher: there is a modest benefit of using 16cores. I reduced it to 8 because scheduling became easier. I think CPU utilization with 8 cores is 650% and more like 1000% with 16 cores.

workflow/snakemake_rules/core_workflow.smk
```diff
@@ -0,0 +1,405 @@
+rule combine_inputs:
```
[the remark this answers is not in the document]
rneher: sorry. that was an accidental commit of a file floating around... I removed it.

nextstrain_profiles/nextstrain-scicore/cluster.json
rneher: I'd make if 4 or 8, but certainly doesn't need to be 64 anymore. 
```
(nextstrain) [neher@login20 ncov2]$ more benchmarks/proximity_score_*
::::::::::::::
benchmarks/proximity_score_africa_region_late.txt
::::::::::::::
s       h:m:s   max_rss max_vms max_uss max_pss io_in   io_out  mean_load       cpu_time
268.1170        0:04:28 1178.01 1767.21 1166.00 1167.89 0.00    0.00    95.12   255.49
::::::::::::::
benchmarks/proximity_score_asia_region_late.txt
::::::::::::::
s       h:m:s   max_rss max_vms max_uss max_pss io_in   io_out  mean_load       cpu_time
315.9518        0:05:15 1161.40 1773.18 1154.78 1157.08 0.00    0.00    90.20   285.40
::::::::::::::
benchmarks/proximity_score_europe_region_late.txt
::::::::::::::
s       h:m:s   max_rss max_vms max_uss max_pss io_in   io_out  mean_load       cpu_time
281.7032        0:04:41 1211.82 1846.09 1199.81 1201.70 0.00    0.00    90.53   255.44
::::::::::::::
benchmarks/proximity_score_north-america_region_late.txt
::::::::::::::
s       h:m:s   max_rss max_vms max_uss max_pss io_in   io_out  mean_load       cpu_time
333.0056        0:05:33 1176.44 1763.07 1174.81 1178.41 0.00    0.00    94.62   315.55
::::::::::::::
benchmarks/proximity_score_oceania_region_late.txt
::::::::::::::
s       h:m:s   max_rss max_vms max_uss max_pss io_in   io_out  mean_load       cpu_time
230.3788        0:03:50 726.38  1234.78 714.36  716.25  0.00    0.00    97.65   225.43
::::::::::::::
benchmarks/proximity_score_south-america_region_late.txt
::::::::::::::
s       h:m:s   max_rss max_vms max_uss max_pss io_in   io_out  mean_load       cpu_time
249.8558        0:04:09 701.34  1252.14 689.50  693.10  0.00    0.00    89.35   223.41
```

scripts/get_distance_to_focal_set.py
rneher: indeed. this should be a parameter. I don't think there much of a performance cost if this parameter is 1000 or larger. but it will start requiring unreasonable amounts of memory if it is in excess of 100k. In between, I wouldn't expect this to matter much.

scripts/get_distance_to_focal_set.py
rneher: oops...

scripts/explicit_translation.py
rneher: `augur translate` has its value. the downside here is that if a sequence has a frameshifting deletion (sequence errors) it will be omitted from the nextalign translation. TreeTime will then infer it as the ancestral sequence. This is likely the best guess in most cases. but might not be the desired behavior in all cases. My suggestion would  be to use this script for a while and see when and where it is better/worse than `augur translate`. (we might also want to rename it :) )

rneher:
Thanks so much, @huddlej ! Super useful. I'll sign off for the day -- feel free to work on it!

nextstrain_profiles/nextstrain-scicore/cluster.json
huddlej: I left this at 4G, but I added a note that the memory use seems to scale at about 0.15 MB * chunk size (based on only on two data points of chunk size = 5000 and 10000). Eventually we could choose to implement the mem_mb as a function of chunk size to make this more flexible.

huddlej (approved):
This is working well for me, @rneher. Even though the `nextalign` package for OS X hasn't made it through Bioconda yet, we should probably merge this as is. We can always update the default config to use nextalign once it is available on all platforms.

rneher:
Thanks, John! I am happy for this to be merged. MacOS binaries are available as one-line download. Do we need to adjust documentation?

huddlej:
> MacOS binaries are available as one-line download.

Yeah, I'm mostly worried about people who manage their workflow environment with `--use-conda` who won't have a clean path to nextalign installation. Worst case, these folks (including myself) can copy the appropriate binary into their `.snakemake/conda/xyx/bin/` directory.

> Do we need to adjust documentation?

I'll review the docs with these changes in mind to see what needs updating.

PR #565, opened 2021-02-25
Improve and expand support for remote input files and unify local path and remote URL handling

### LINKED ISSUE #662: Download starting files for Open build via HTTPS instead of S3
**Context**
Initial downloads for runs of the Open profile would be faster and cost Nextstrain less in AWS outbound transfer costs.

**Description**
The files we publish as starting points for the Open profile (see #628) are currently downloaded/uploaded directly from/to S3 using `aws s3` pointed at `s3://nextstrain-data/files/ncov/open/`. Instead, download these via our CloudFront endpoint for that bucket instead, https://data.nextstrain.org/files/ncov/open/, possibly using the supporting code prototyped in #565.

Uploads would still have to go directly to S3, but we'd want to start invalidating CloudFront's cache on upload. Switching from `aws s3` to `nextstrain remote upload` would handle invalidation this for us, and I believe `nextstrain` is already expected to be available in the environment for this build. There are some features of the current `scripts/upload-to-s3` that `nextstrain remote upload` wouldn't provide, and we should identify those and consider the impact of losing them.

The build would still have to support downloads directly from S3 for the GISAID profile and (presumably) other custom profiles.

tsibley:

Uses Snakemake's existing support for remote files, along with some of our own code to make the local vs. remote distinction ~transparent to the workflow.

- Simplifies workflow rules by eliminating special "download" rules
- Expands existing support (just `s3://`) to include `http(s)://` and `gs://`
- Completes support for compressed remote files for all input stages
- Leaves downloaded files compressed on disk for all stages

Full details in the commit messages.

## Related issue(s)

Resolves #662.

## Testing

I've tested locally by hand and using the existing `tests/different-inputs.t`. I have not yet run a full trial build on AWS Batch, but don't foresee issues. I'd appreciate :eyes: and testing from folks who run the workflow routinely.

## Release checklist

I don't consider this PR to introduce backwards incompatible changes (at least in anything we're promising to keep stable) but would appreciate others' thoughts here.

One point of UX to consider is that remote files will be downloaded by Snakemake to local paths the corresponding to the remote URL without its scheme, e.g. `https://data.nextstrain.org/files/ncov/open/metadata.tsv.gz` is downloaded to `data.nextstrain.org/files/ncov/open/metadata.tsv.gz`. I've added the patterns `/nextstrain-data/` and `/data.nextstrain.org/` to the `.gitignore` to exclude expected inputs, but other input URLs will produce other local paths which aren't easily ignored.

This pull request does introduce new features:


jameshadfield (commented):
A whole lot nicer - thanks Tom! Tested using a few different (ncov/open) endpoints and it works well, as well as running `different-inputs.t`.

The downloading of remote files into folders based on their address isn't ideal as I can see someone accidentally committing private metadata this way, but also isn't a dealbreaker by any means. I had a look at overriding [some snakemake methods](https://github.com/snakemake/snakemake/blob/0054691052ca35b3806c7eb9d055724d19681f3c/snakemake/remote/__init__.py#L203-L207) but I'm not familiar enough with all the moving parts there to get it working. 

It wasn't clear to me how to force snakemake to re-download the resources - force running a rule which has a remote input uses the local copy. (Perhaps it's being clever and checking that the remote hasn't changed, but given that I can run the step without credentials I don't think so.) Deleting the downloaded files does the trick, of course.

workflow/snakemake_rules/remote_files.smk
```diff
@@ -0,0 +1,89 @@
...
+    return producer
+
+
+SCHEME_REMOTES = {
+    "https": _remote_producer(lambda HTTP: HTTP.RemoteProvider().remote),
+    "http": _remote_producer(lambda HTTP: partial(HTTP.RemoteProvider().remote, insecure = True)),
+    "s3": _remote_producer(lambda S3: S3.RemoteProvider().remote),
+    "gs": _remote_producer(lambda GS: GS.RemoteProvider().remote),
+    "": lambda: lambda local_path, **kwargs: local_path,
+}
```
ivan-aksamentov, line 46: I am curious whether there is a way in Snakemake to autodetect the provider from the scheme automagically and to not hardcode a subset (somewhat arbitrarily chosen. Why not Azure or Dropbox for example?), but to use everything available in Snakemake. Deferring the choice to Snakemake itself might also simplify the scheme/path logic below.

I don't know Snakemake very well, but I am surprised his is not automatic.
Could not find anything about autodetection in the docs either:
https://snakemake.readthedocs.io/en/stable/snakefiles/remote_files.html

tsibley: @ivan-aksamentov Snakemake tries to offer auto-detection through its [`AUTO` provider](https://snakemake.readthedocs.io/en/stable/snakefiles/remote_files.html#auto), but it's broken.

It enumerates and loads all the various `snakemake.remote.*` modules on each call to `AUTO.remote()`. This slows things down unnecessarily, but isn't a deal breaker. However, during enumeration, it protects against unsupported modules due to missing deps, but not supported modules which fail to initialize because of missing required credentials/config from the environment or missing required parameters to the remote provider or `AUTO.remote()`. This alone makes it completely unusable. I can't imagine how anyone is using it in this state. I suspect no one is.

(And even with a functioning `AUTO`, we'd still need some handling for local paths.)

tsibley: I'll add: We could make a better functioning `AUTO` equivalent ourselves, by similarly dynamically building the scheme → remote function mapping, but it didn't seem worth it to support use cases that aren't concrete right now. The implementation in this PR could certainly be further developed in that direction in the future, however.

workflow/snakemake_rules/remote_files.smk
```diff
@@ -0,0 +1,89 @@
...
+    first call of the producer.
+    """
+    # Name of first argument to *init*.
+    remote_name = init.__code__.co_varnames[0]
+    remote = None
+
+    def producer():
+        nonlocal remote
+        if not remote:
+            remote = init(import_module(f"snakemake.remote.{remote_name}"))
```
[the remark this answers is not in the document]
tsibley: Right, the intention was to let Snakemake's WorkflowError be raised if optional deps weren't available, as those errors already describe the right deps to install. It's a good point though that they don't describe _how_ to install them, which might be tricky to do since "it depends" is the answer.

`boto3` gets pulled in as a transitive dep from several other deps already, both in our Conda environment and the Docker image. The Google deps are not in the Docker image, but they are in the Conda environment.

I thought this was a reasonable state to be in, but I can update the Docker image if desired.

I'll update the remote inputs docs to mention that optional deps may need to be installed for remote file support.

docs/remote_inputs.md
```diff
@@ -17,6 +17,10 @@ The open (GenBank) data is publicly available at three endpoints:
 **Our intention is to make GenBank intermediate files open and available for everyone to use, and to keep these files up-to-date.**
 The paths for specific files are the same under each endpoint, e.g. `https://data.nextstrain.org/files/ncov/open/metadata.tsv.gz`, `s3://nextstrain-data/files/ncov/open/metadata.tsv.gz`, and `gs://nextstrain-data/files/ncov/open/metadata.tsv.gz` all exist.
 See below for a list of files that exist.
+If you're running workflows on AWS or GCP compute that fetch this data, please use the S3 or GS URLS.
+Otherwise, please use the https://data.nextstrain.org URLs.
+
+Note that even though the `s3://nextstrain-data/` and `gs://nextstrain-data/` buckets are public, the defaults for most S3 and GS clients require _some_ user to be authenticated, though the specific user/account doesn't matter.
```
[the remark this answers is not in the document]
tsibley: I'm not sure if we can catch that exception or not without further digging. One thing we could do here, at least for S3, would be to add a special case for the `nextstrain-data` bucket that forces anonymous access by disabling request signing.

However, a simpler alternative would be to document that if want to use a public S3 or GS bucket anonymously, you should instead configure an input like `https://nextstrain-data.s3.amazonaws.com/files/ncov/open/metadata` or `https://storage.googleapis.com/nextstrain-data/…`.

tsibley:
@jameshadfield 

> The downloading of remote files into folders based on their address isn't ideal as I can see someone accidentally committing private metadata this way, but also isn't a dealbreaker by any means. I had a look at overriding [some snakemake methods](https://github.com/snakemake/snakemake/blob/0054691052ca35b3806c7eb9d055724d19681f3c/snakemake/remote/__init__.py#L203-L207) but I'm not familiar enough with all the moving parts there to get it working.

I think we could indeed monkey-patch this into storing remote files under a single consistent subdir, but it'd be a bit of work and a bit of a hack to account for all the callers of `local_file()` that assume the current behaviour. I don't think it's worth it.

> It wasn't clear to me how to force snakemake to re-download the resources - force running a rule which has a remote input uses the local copy. (Perhaps it's being clever and checking that the remote hasn't changed, but given that I can run the step without credentials I don't think so.) Deleting the downloaded files does the trick, of course.

It does check the mtime of the remote file by making a `HEAD` request (many identical requests, actually). If let Snakemake download a file from S3, then modify the mtime of the local copy to be older than the remote, and run Snakemake again with no credentials, it fails when it goes to check the remote mtime. I suspect you had ambient AWS credentials available that you didn't realize? Or maybe something else is going on.

workflow/snakemake_rules/remote_files.smk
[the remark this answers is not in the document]
jameshadfield: Yeah, simply adding some docs should be enough here. I didn't realise it appears as a transitive dep in most situations, I guess that means I have a strange setup myself!

docs/remote_inputs.md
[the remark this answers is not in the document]
jameshadfield: The latter, I think. If we end up doing config-file checking (I hope so!) I could imagine covering this use case at that step.

workflow/snakemake_rules/remote_files.smk
tsibley: I added this reasoning as a comment at the top of `remote_files.smk` so it's recorded for future reconsideration. Thanks for spurring me to explain it.

docs/remote_inputs.md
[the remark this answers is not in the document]
tsibley: Added in 2ae1cbc.

workflow/snakemake_rules/remote_files.smk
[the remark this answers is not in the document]
tsibley: Addressed in fc5c4583.

workflow/snakemake_rules/remote_files.smk
```diff
@@ -0,0 +1,109 @@
...
+
+    The call to *init* (that is, the production of the remote input function)
+    is deferred until the first time the producer is called.
+
+    If the remote module is not available because required dependencies aren't
+    installed, then a :exc:`WorkflowError` exception object is returned on the
+    first call of the producer.
+    """
+    # Name of first argument to *init*.
+    remote_name = init.__code__.co_varnames[0]
```
huddlej, line 48: What if we defined `remote_name` as an argument to `_remote_producer` instead of using code introspection like this? I mostly was surprised to see a variable name used as a variable itself and wonder if Future Me (or others) will be similarly confused later.

workflow/snakemake_rules/common.smk
```diff
@@ -40,45 +40,17 @@ def _get_path_for_input(stage, origin_wildcard):
...
-        elif path_or_url:
-            return path_or_url
-        else:
-            return f"results/filtered_{origin_wildcard}.fasta.xz"
-
-    raise Exception(f"_get_path_for_input with unknown stage \"{stage}\"")
+    input_file = config.get("inputs", {}).get(origin_wildcard, {}).get(stage, "")
+
+    if input_file:
+        return path_or_url(input_file, keep_local=True)
```
huddlej, line 46: This is a major improvement!

.gitignore
```diff
@@ -10,6 +10,10 @@ defaults/colors.tsv
 defaults/colors_*.tsv
 stats.json
 
+# Downloaded remote files from sources we expect
+/nextstrain-data/
+/data.nextstrain.org/
```
huddlej, line 15: I wish these remote data directories could be easily nested in the existing `data/` directory, to follow the pattern we've been demonstrating to users at workshops and in the documentation. I see though that this is a Snakemake issue and not an issue with this PR.

huddlej (approved):
This is an excellent update to the workflow and nicely clarifies the original logic. I only have a couple of clarifying questions/comments included inline below. None of these are blocking though, so I would be happy to see this merged now.

docs/remote_inputs.md
```diff
@@ -17,6 +17,15 @@ The open (GenBank) data is publicly available at three endpoints:
 **Our intention is to make GenBank intermediate files open and available for everyone to use, and to keep these files up-to-date.**
 The paths for specific files are the same under each endpoint, e.g. `https://data.nextstrain.org/files/ncov/open/metadata.tsv.gz`, `s3://nextstrain-data/files/ncov/open/metadata.tsv.gz`, and `gs://nextstrain-data/files/ncov/open/metadata.tsv.gz` all exist.
 See below for a list of files that exist.
+If you're running workflows on AWS or GCP compute that fetch this data, please use the S3 or GS URLS.
+Otherwise, please use the https://data.nextstrain.org URLs.
```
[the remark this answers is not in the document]
tsibley: Ahh. S3 is significantly cheaper (and probably faster?) inside of AWS. Ditto for GS inside of GCP.

I'll add a note!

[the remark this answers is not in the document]
tsibley: 895e5ec4

workflow/snakemake_rules/common.smk
```diff
@@ -40,45 +40,17 @@ def _get_path_for_input(stage, origin_wildcard):
...
+
+    if stage in {"metadata", "sequences"}:
+        raise Exception(f"ERROR: config->input->{origin_wildcard}->{stage} is not defined.")
+    elif stage in {"aligned", "masked", "filtered"}:
+        return f"results/{stage}_{origin_wildcard}.fasta.xz"
+    else:
+        raise Exception(f"_get_path_for_input with unknown stage \"{stage}\"")
```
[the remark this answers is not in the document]
tsibley: Yeah. That was the existing error, and I think it would mostly be for devs who added support for a stage to once place but not another.

.gitignore
tsibley: Same. I looked into it, and to do so is technically possible, but involves either a lot of monkey-patching Snakemake or writing our own multi-protocol `snakemake.remote.AbstractRemoteProvider` implementation class that supports HTTPS, S3, GS, etc without proxying to the existing single-protocol implementations. I just don't think it is worth it.

workflow/snakemake_rules/remote_files.smk
tsibley: I considered that exact `remote_name` argument, but to me, small, contained uses of introspection like this to make tiny lambdas more scannable seems like a perfect use case. I realize it's not a widely-used technique in most Python in the wild, though. I hoped the documentation and comment would abate any future confusion.

Here's what it'd look like to add a `remote_name` argument. It seems to me that it takes the same quality and quantity of effort to understand what's happening as with the introspection version, but I'm maybe bias here.

```diff
diff --git a/workflow/snakemake_rules/remote_files.smk b/workflow/snakemake_rules/remote_files.smk
index 2e143d9a..a8fb49b0 100644
--- a/workflow/snakemake_rules/remote_files.smk
+++ b/workflow/snakemake_rules/remote_files.smk
@@ -25,14 +25,13 @@ from importlib import import_module
 from urllib.parse import urlsplit, urljoin
 
 
-def _remote_producer(init):
+def _remote_producer(remote_name, init):
     """
-    Produce a remote input function by loading a :class:`snakemake.remote`
-    module and passing the module to an *init* function.
+    Produce a remote input function by loading the :class:`snakemake.remote`
+    module named by *remote_name* and passing the module to an *init* function.
 
-    The module is named by the first argument of the *init* function.  *init*
-    will be called with the loaded module and is expected to return a remote
-    input function suitable for use inside :func:`path_or_url`.
+    *init* will be called with the loaded module and is expected to return a
+    remote input function suitable for use inside :func:`path_or_url`.
 
     Returns a "producer" function which when called with no arguments returns
     the remote input function.
@@ -45,7 +44,6 @@ def _remote_producer(init):
     first call of the producer.
     """
     # Name of first argument to *init*.
-    remote_name = init.__code__.co_varnames[0]
     remote = None
 
     def producer():
@@ -58,10 +56,10 @@ def _remote_producer(init):
 
 
 SCHEME_REMOTES = {
-    "https": _remote_producer(lambda HTTP: HTTP.RemoteProvider().remote),
-    "http": _remote_producer(lambda HTTP: partial(HTTP.RemoteProvider().remote, insecure = True)),
-    "s3": _remote_producer(lambda S3: S3.RemoteProvider().remote),
-    "gs": _remote_producer(lambda GS: GS.RemoteProvider().remote),
+    "https": _remote_producer("HTTP", lambda HTTP: HTTP.RemoteProvider().remote),
+    "http": _remote_producer("HTTP", lambda HTTP: partial(HTTP.RemoteProvider().remote, insecure = True)),
+    "s3": _remote_producer("S3", lambda S3: S3.RemoteProvider().remote),
+    "gs": _remote_producer("GS", lambda GS: GS.RemoteProvider().remote),
     "": lambda: lambda local_path, **kwargs: local_path,
 }
```

tsibley: In any case, happy to commit the `remote_name` version above if you think it's better.

huddlej: Thank you for placing these side by side! From this view, the explicit remote argument is slightly easier to interpret, because it is clearer to me how that value is being used as a variable.

.gitignore
huddlej: No more monkeys...

tsibley: …jumping in the code.

workflow/snakemake_rules/remote_files.smk
tsibley: Got it. Repushed to make this change.

ivan-aksamentov (approved):

PR #571, opened 2021-03-03
WIP: Use metadata-only filtering in subsample jobs

huddlej:
Minimize inspection of large FASTA files by subsampling only with metadata and a sequence index and outputting only strain names for each subsample rule. Uses the new metadata-only filter interface proposed for Augur to collect the resulting subsampled strains and output a single FASTA file of all distinct strains.

As noted in 3216c5f, some subsample jobs require priority scores for another subsampled set and these score calculations require subsampled sequences. That same commit as a rule to extract only the subsampled sequences for a specific set as needed.

Although the goal of metadata-only filtering is to dramatically speed up this workflow, this PR does not address a major bottleneck in the current workflow which is the amount of time required by augur filter to write sequences to disk.

### Related issue(s)
Related to https://github.com/nextstrain/augur/pull/679

### Testing
I've tested this workflow with the Nextstrain Europe build and also the multiple inputs example data.

CI tests will fail because this PR requires an Augur development branch (code that is not available in the Docker image). To test this PR locally, run Snakemake with the conda mode like so:

```
snakemake --use-conda --cores 4 --profile nextstrain_profiles/nextstrain --config active_builds=europe
```

### DIFF
--- a/nextstrain_profiles/nextstrain/builds.yaml
+++ b/nextstrain_profiles/nextstrain/builds.yaml
@@ -31,7 +31,7 @@ inputs:
     # sequences: "s3://nextstrain-ncov-private/sequences_gisaid.fasta.gz"
     # aligned: "s3://nextstrain-ncov-private/aligned_gisaid.fasta.xz"
     # to-exclude: "s3://nextstrain-ncov-private/to-exclude_gisaid.txt.xz"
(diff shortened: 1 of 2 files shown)


huddlej:
This PR will be superceded by work on a related PR that adds a new form of priority calculation.

PR #572, opened 2021-03-03
Stream combined sequences to disk instead of loading into memory

huddlej:

Instead of reading all sequences into memory to identify duplicates and then write out to disk, stream sequences from the multiple input files to disk and check for duplicates on the fly with a hash of the sequence content indexed by strain name. Hashes require far less memory than the actual sequences, so we are able to output a full list of duplicate sequences to stderr instead of the first duplicate.

Mismatched duplicates are treated as errors. All of such strains are reported at the end and an error is thrown.

### Related issue(s)  
Related to #571 (this commit content is copied from work done in that PR)

### Testing

```bash
# Combine and dedup the same file. This should write out a single copy of each strain sequence.
$ python3 scripts/combine-and-dedup-fastas.py \
  --input data/example_sequences_aus.fasta data/example_sequences_aus.fasta \
  --output test.fasta

# Confirm that both inputs and the output have the same number of records.
$ grep "^>" data/example_sequences_aus.fasta | wc -l
91
$ grep "^>" test.fasta | wc -l
91

# Make a modified copy of the original input sequences.
sed 's/AAGGNNNN/AAGGAAAA/' data/example_sequences_aus.fasta > example_sequences_aus_modified.fasta

# Combine and dedup the original and modified inputs.
# This should output the mismatches sequences and return a non-zero error code.
$ python3 scripts/combine-and-dedup-fastas.py \
  --input data/example_sequences_aus.fasta example_sequences_aus_modified.fasta \
  --output test.fasta
ERROR: Detected the following duplicate input strains with different sequences:
    Australia/VIC330/2020
    Australia/VIC969/2020
    Australia/VIC1045/2020
    Australia/VIC431/2020
    Australia/VIC329/2020
    Australia/VIC987/2020
    Australia/VIC05/2020
    Australia/VIC187/2020
    Australia/VIC970/2020
    Australia/VIC1120/2020
    Australia/VIC962/2020
    Australia/VIC1175/2020
$ echo $?
1
```

### DIFF
--- a/scripts/combine-and-dedup-fastas.py
+++ b/scripts/combine-and-dedup-fastas.py
@@ -1,6 +1,9 @@
-from Bio import SeqIO
 import argparse
-from augur.align import read_sequences
+from Bio import SeqIO
+import hashlib
+import sys
+import textwrap
+
 
 if __name__ == '__main__':
     parser = argparse.ArgumentParser(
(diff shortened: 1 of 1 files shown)


rneher:
I haven't tested this yet, but what I don't fully understand is the behavior when duplicate strains are encountered. If the sequences differ, it will mark them as duplicates and write out the first encounter of the strain name. If the sequences are identical, the sequences is written out again, right? Is this the desired behavior?

huddlej:
Thank you for the extra eyes on this, @rneher! That was definitely not the intended behavior. I just fixed that bug, added some tests to the PR comment, and changed the behavior back to make mismatched duplicates an error instead of a warning.

joverlee521 (commented):
I was able to run this locally with the aligned GISAID fasta and the example Australian fasta without 🔥

jameshadfield (approved):
Tested (via the multiple-inputs example + making modifications to the inputs) and this looks great. Preventing laptop fires is always a good thing 😉

jameshadfield:
P.S. The [combine metadata](https://github.com/nextstrain/ncov/blob/master/scripts/combine_metadata.py) script similarly stores everthing in memory (probably in duplicate). This isn't as problematic as sequences, but that script could also be improved as needed.

huddlej:
Thank you all for the testing! I'll merge now.

PR #573, opened 2021-03-05
Use a Snakemake profile for Travis CI tests

huddlej:
The current CI tests override the default Nextstrain data with the example data by passing arguments to the Snakemake config. However, we recently deprecated the top-level `sequences` and `metadata` keys and recent versions of Snakemake no longer allow empty values to be passed for config keys (e.g., `input=`), so whenever we upgrade Snakemake in the Nextstrain Docker image, the current CI tests would start to fail.

Since there is no interface in Snakemake to override nested config keys from the command line, this PR moves the logic to override sequence data into a Snakemake profile that inherits all of the other Nextstrain build settings. This PR also changes the test build from "global" to "europe" to make sure CI tests a path through proximity-based subsampling.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,4 +12,4 @@ install:
   - nextstrain update
 script:
   - gunzip data/example_sequences.fasta.gz
-  - nextstrain build . all_regions -j 2 --profile nextstrain_profiles/nextstrain --config sequences=data/example_sequences.fasta metadata=data/example_metadata.tsv active_builds=global inputs=
+  - nextstrain build . all_regions -j 2 --profile nextstrain_profiles/nextstrain-ci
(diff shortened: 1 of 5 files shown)


jameshadfield:
> recent versions of Snakemake no longer allow empty values to be passed for config keys (e.g., input=

I didn't know this, but i'm happy to hear it. I had to[allow string values](https://github.com/nextstrain/ncov/blob/master/workflow/schemas/config.schema.yaml#L15) in the schema for this use case. Removing this makes the schema more powerful and gives nicer error messages. This [conditional](https://github.com/nextstrain/ncov/blob/master/Snakefile#L40) can also be simplified.

huddlej:
Thanks for pointing those out, @jameshadfield! 22f1571 removes both the schema allowance for input strings and the conversion of a missing input to an empty string in the Snakefile.

Assuming CI tests pass with these new changes, I'll plan to merge this.

PR #574, opened 2021-03-07
fix: names for diagnostic files

rneher:
Download rules for outputs of the diagnostic rule were not expanded correctly.


rneher:
@jameshadfield: the current pattern is a little confusing as the specification of one input triggers the download of several files. I am wondering whether one could make this more explicit that the downloads are "per rule" or make a generic rule for every file.

PR #576, opened 2021-03-09
Let nextstrain know how much memory the filter step requires.

ttung:
Inform snakemake how much the filter step requires so it doesn't overschedule the memory.

### Testing
<strike>Running a build now! :)</strike>Successfully ran a build.

### DIFF
--- a/workflow/snakemake_rules/main_workflow.smk
+++ b/workflow/snakemake_rules/main_workflow.smk
@@ -183,6 +183,10 @@ rule filter:
         sequences = "results/filtered{origin}.fasta"
     log:
         "logs/filtered{origin}.txt"
+    resources:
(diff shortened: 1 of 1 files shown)


huddlej:
Thanks for this PR, @ttung. Can you say more about which version of Augur you're using? As of [version 10.0.4](https://github.com/nextstrain/augur/blob/master/CHANGES.md#1004-6-november-2020), augur filter should not be loading all sequences into memory at any point. Since then we've tried to limit all interactions with the sequence data to single passes through the FASTA as a generator. If filter is still using more memory, that indicates a bug we need to fix with Augur.

ttung:
We are using whatever is in the `nextstrain/base` docker image.

huddlej:
Closing this since #582 adds memory constraints for the filter rules and other rules in the workflow.

PR #577, opened 2021-03-09
parse_metadata.py try to correct US cities to counties

eharkins:
Adds a config file so that we can suggest rules to correct US city locations to their corresponding county. Only works if the city is spelled right. Another option would be to just go ahead and add all of these as rules instead of suggesting them as they arise. For more context, see https://bedfordlab.slack.com/archives/CTZKJC7PZ/p1614851125181400.


emmahodcroft:
I haven't tried this but if this works, I think it would be a great idea! It would be fantastic to get things sorted out more into counties rather than having a mix and automating this would certainly help.

Two small notes:
- You may want to add "Houston Area" (I think that's what the Houston Hospital uses...) to those not to change - I think this may cover multiple counties in Texas, but currently we don't know as they don't say more. But might be better to leave this as-is rather than saying it's all in Houston city (Harris County).
- We need to be sure to reference https://simplemaps.com/data/us-cities in the code - it's a condition of using the CSV file! (See the webpage for more details - click download under the free version to see the agreement-thing.)

MoiraZuber:
This looks very nice! Two comments:

- I think this would easily be adaptable to not only correct city level -> county level, but also counties that are lacking the "County" specification (e.g. Texas, Parmer should be Texas, Parmer County)

- At this position in the code, this feature will only be able to capture new cases. All older instances of city level location will remain the same. We could probably just write a quick one-time script to produce all necessary rules for older cases. Another option would be to employ the function of parse_metadata that is responsible for fixing Belgium locations and divisions. This should _theoretically_ be adaptable for any country given a proper database, but in the case of the USA, this would require quite a lot of fiddling with duplicates I think...

eharkins:
Ah good points, I wasn't intending to capture 
> counties that are lacking the "County" specification (e.g. Texas, Parmer should be Texas, Parmer County)

but as Emma points out, as it is now this will capture (and not necessarily do what we want) some of the ambiguous cases such as
> Houston (city) which is in Harris County, and then we have Houston County (totally different place)

> So not sure there's a lot we can do here except maybe to flag particular locations somehow if we want to ensure they get a double-look by a human.

To be clear this just suggests rules so there is an element of human proofreading of the rules in that by default they don't get added. But maybe good to add in special cases so it doesn't always suggest a rule that's wrong for a particular location.

> We could probably just write a quick one-time script to produce all necessary rules for older cases.

I would say if we go this route then maybe it's worth it to spend that time making the one-time script add rules for old and new cases by looping over the US cities database, and leaving out / changing a few key examples that we know of such as Houston. And perhaps its possible to check for similar cases by checking:

```
Is the city in a county of the same name? If yes, then probably safe to add a County correction rule.
If not, is there such a county in that state that could create some ambiguity?
In this case don't add the rule, just log the information and we can work through these cases manually.
```
Though I would think this sort of approach would make the script a lot slower having to search for similar county names for each city, so it would probably only make sense to do in a one-time, all-rule-adding script as opposed to doing it in parse_metadata.py.

emmahodcroft:
>Is the city in a county of the same name? If yes, then probably safe to add a County correction rule.
If not, is there such a county in that state that could create some ambiguity?
In this case don't add the rule, just log the information and we can work through these cases manually.

That sounds very reasonable to me Eli!

rneher:
If anything, this related to `augur curate` or `ncov-ingest` at this point.

PR #580, opened 2021-03-11
WIP: Add mutation counts

rneher:
In our influenza analysis, we routinely count epitope mutations of various sorts. 

This PR adds similar functionality to SARS-CoV-2 builds by adding distance maps and a rule calling the `augur distance` command.

The current maps and names are preliminary.


trvrb:
This is really great @rneher! Thanks for putting it together. I just made two very small changes to get this mergable. One was to create a proper "coloring" for "S1 mutations" and the other was to temporarily remove `MoC_count` from distance maps. I do think we should develop something here, but I believe that S1 mutations is useful enough on its own that this can be merged as is.

I do think it would be preferable to call a single indel stretch as 1 "mutation" in this accounting, but this doesn't block this merge for me.

One thing to flag (for @jameshadfield) is that Auspice frequencies is treating `0` as missing data. You can see that 
```
    "Wuhan/Hu-1/2019": {
      "S1_mutations": 0
    },
```
as intended and this flows through to Auspice as:

Screen Shot 2021-03-11 at 5 27 14 PM

However, the frequencies panel is treating these as "Unassigned" rather than "0" as far as I can tell.

Screen Shot 2021-03-11 at 5 28 12 PM

This appears to be an Auspice bug and doesn't block merge.

rneher:
Yes, this will require an update to the augur distance command. Since we only do rather simple distance calculations here, we could just implement the deletion count in a little script.

trvrb:
Sounds great. Will be really useful to have this live. Thanks for adding Richard.

PR #581, opened 2021-03-11
make diagnostic use info in metadata

rneher:
This PR uses the newly available info in metadata to simplify and speed up the filtering for by clock deviation and SNP clusters

The threshold can be set as arguments (only in script yet), not the config. 

This will speed up the workflow by 20min or so. 

However, some other analysis use some of the output of the previous diagnostic and we need to port those over to use the `mutation_summary.tsv` (cc @emmahodcroft) before we can merge this.


rneher:
closing in favor of #667

PR #582, opened 2021-03-12
Specify memory requirements for most rules

huddlej:

Adds `mem_mb` resource requirements for high-memory rules [based on benchmarks from a full Nextstrain build](https://gist.github.com/huddlej/020afba80fa4965e3d4d9d89234b6949) and corresponding input file sizes. The most common source of moderate memory usage is loading metadata in augur (or other scripts) as a dictionary with augur's `read_metadata` function. The most common source of high memory usage is loading multiple sequence alignments as in the
`refine` and `ancestral` rules.

Multipliers for memory usage based on input file sizes are conservative, erring on the side of requesting more memory than less to avoid hard memory limits that can kill jobs. For example, at the time the main metadata file for Nextstrain builds uses 165 MB on disk and ~2200 MB in memory (as a Python dictionary), so the multipler used for rules that load metadata is 15.

Interestingly, scripts that load metadata into a pandas DataFrame use 25% of the memory (~500 MB) than scripts (or augur commands) that load metadata into a dictionary (~2 GB). Future work on the metadata infrastructure should consider using pandas exclusively and take advantage of [pandas' tools to reduce memory usage](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.memory_usage.html) by storing some column values as categorical (essentially enum) values instead of as distinct objects.

### Related issue(s)  

Related to #576

### Testing

Tested with a full Nextstrain build for Europe on a SLURM cluster with:

```bash
snakemake \
  all_regions
  --conda-frontend mamba \
  --profile nextstrain_profiles/nextstrain-rhino/ \
  --config active_builds=europe
```

### DIFF
--- a/workflow/snakemake_rules/export_for_nextstrain.smk
+++ b/workflow/snakemake_rules/export_for_nextstrain.smk
@@ -53,6 +53,11 @@ rule export_all_regions:
         colors = expand("results/{build_name}/colors.tsv", build_name=BUILD_NAMES),
     benchmark:
         "benchmarks/export_all_regions.txt"
+    resources:
+        # Memory use scales primarily with the size of the metadata file.
+        # Compared to other rules, this rule loads metadata as a pandas
+        # DataFrame instead of a dictionary, so it uses much less memory.
+        mem_mb=lambda wildcards, input: 5 * int(input.metadata.size / 1024 / 1024)
     conda: config["conda_environment"]
     shell:
         """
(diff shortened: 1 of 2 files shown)


huddlej:
@ttung Would you be able to try out this branch for your builds? In addition to specifying memory constraints for filter steps, this PR also adds constraints for all other high-memory rules and should help job scheduling.

I can't add you as a reviewer directly, but I'd love your feedback on this. If it works for you, I'll merge this and close #576, since this is a superset of those changes.

huddlej:
@ttung I'm going to merge this now, but we can refine it later if we need to adjust some of these constraints.

ttung:
I will definitely let you know if we see any more OOMs!  Thanks a bunch!

PR #583, opened 2021-03-13
Gap aware mutation counts

rneher:
this PR temporarily replaces augur distance by a custom distance calculation to avoid the overcounting of long gaps:

instead of
https://nextstrain.org/staging/ncov/scounts/global?c=S1_mutations

this now looks like this:
https://nextstrain.org/staging/ncov/scounts2/global?c=S1_mutations

### DIFF
--- a/scripts/mutation_counts.py
+++ b/scripts/mutation_counts.py
@@ -18,14 +18,23 @@ if __name__ == '__main__':
     parser.add_argument('--output', type=str, required=True, help="output json file")
     args = parser.parse_args()
 
+    if type(args.map)==str:
+        args.map = [args.map]
+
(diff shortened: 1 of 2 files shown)


rneher:
@huddlej @trvrb now that S1 coloring is advertised, do we want to merge this or wait for a fix in augur? I am worried the next build might again include partial Spike sequences which throws off the mutation count?

huddlej:
@rneher I'd vote to merge this and then release an updated Augur when that feature is implemented. The Augur release will always take longer to propagate to everyone and could benefit from the initial prototyping here.

I can work on the Augur update tomorrow afternoon.

PR #586, opened 2021-03-17
Adding Aggregate Design pattren

akbhangal:
Adding the aggregate design pattern to better handle encapsulation and composition - aggregate is using gisaid as aggregate root.

### Related issue(s)  

Fixes #  na
Related to #  na

### Testing
What steps should be taken to test the changes you've proposed?   Running and building steps should be same as before and generate *.jason file and test. On front end user should not notice any changes in output as only code is modified to support new design,

### DIFF
--- a/scripts/developer_scripts/apply_aggregate.py
+++ b/scripts/developer_scripts/apply_aggregate.py
@@ -1,0 +1,12 @@
+
+#Gisaid class is used to create aggregate that contain gisaid_epi_isl as root and 
+#encapsulate/privatize the entity strain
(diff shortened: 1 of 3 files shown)


huddlej:
Thanks @akbhangal, but this code doesn't address any of our immediate issues, so we're going to close this for now. We'd happily consider your submission in the future provided some more context for the issue(s) that it resolves. Check out [our contributor guide for more details on finding open issues to resolve](https://github.com/nextstrain/.github/blob/master/CONTRIBUTING.md).

PR #587, opened 2021-03-17
Add aggregate Design Pattren Unit for Region to handle compositions 

akbhangal:
Adding additional aggregate units for region in addition to previous enhancement. Strain is accessible via gisaid and region is independent but controls other entities like country, location, division.

### Related issue(s)  

Fixes #  
Related to #  

### Testing
What steps should be taken to test the changes you've proposed? Running and building steps should be same as before and generate *.jason file and test. On front end user should not notice any changes in output as only code is modified to support new design,


rneher:
@akbhangal

PR #590, opened 2021-03-19
WIP: Prototype a build showing an overview of all Pangolin lineages

huddlej:

Adds a Snakemake profile with a single build that subsamples an existing filtered alignment of GISAID data to one sample per Pangolin lineage. The resulting tree shows the general relationships among these lineages.

This is a proof-of-concept for an approach to a general view of all circulating lineages that could link out to lineage-specific builds.

### Testing

[An example of this build is also available on the staging server](https://nextstrain.org/staging/ncov/pangolin).


trvrb:
Thanks John! Doing a simple `--group_by pango_lineage` with `--seq_per_group 1` was a very clever approach to the design problem. Here, I've just cleaned a few things up to:

1. Provide color ordering that gives alphabetically adjacent lineages similar colors
2. Use `pango_lineage` rather than `pangolin_lineage` to match with metadata field
3. Provide PANGO-specific `clades.tsv` to help structure visualization
4. Provide PANGO-specific Auspice config to default to coloring by PANGO lineage

The resulting output can be seen at: https://nextstrain.org/staging/ncov/lineages?tl=pango_lineage

emmahodcroft:
This is pretty cool! Thanks for doing this John, it's very interesting.

I like Trevor's additions, and was also thinking the same thing as his point 1 - we know some of the pango lineage samples sometimes end up pretty random places in our trees, so it would be good if we could avoid inadvertently sampling one of those as a 'representative'.

I also like Trevor's ideas in 2 both for sizing the tips and in linking to more dedicated builds, perhaps at least for a few of the largest lineages.

rneher:
superseeded by nextclade tree.

PR #591, opened 2021-03-20
Include custom coloring for PANGO lineages

trvrb:
The default Auspice coloring looks pretty garbled for PANGO lineages. This ensures that the spectrum is spread across lineages and that lineages that are alphabetically similar have similar colors.

Existing:

existing

PR:

pr

Still a bit jumbled (mainly due to persistent B.1s), but not as bad.

This will require occasional updates to `color_ordering.tsv`, but I think this is okay as new PANGO lineages not present in `color_ordering.tsv` will just show up in shades of gray and so updates can be infrequent.

Trial outputs are available as:
- https://nextstrain.org/staging/trial/pango-colors/ncov/global?c=pango_lineage&d=tree
- https://nextstrain.org/staging/trial/pango-colors/ncov/africa?c=pango_lineage&d=tree
- https://nextstrain.org/staging/trial/pango-colors/ncov/asia?c=pango_lineage&d=tree
- https://nextstrain.org/staging/trial/pango-colors/ncov/europe?c=pango_lineage&d=tree
- https://nextstrain.org/staging/trial/pango-colors/ncov/north-america?c=pango_lineage&d=tree
- https://nextstrain.org/staging/trial/pango-colors/ncov/oceania?c=pango_lineage&d=tree
- https://nextstrain.org/staging/trial/pango-colors/ncov/south-america?c=pango_lineage&d=tree


emmahodcroft:
I really like how this looks @trvrb ! I think it's indeed, a big improvement. My only concern is that actually, new Pango lineages seem to appear at a pretty good clip (though I guess some get 'retired' as well (?)). How can we ensure that updating the `color_ordering.tsv` file happens regularly, and is really simple to do? (How did you generate this one - is it worth including a script in the repo to generate a new list very easily?)

emmahodcroft:
I noticed this PR got updated and was just wonding if there were any comments on how we can ensure the colours for lineages stay updated easily as new Pango lineages come online - for example, is there a script that can be easily run that generates a new colours file? Then we'd just need to remember to do it every month or so (or maybe even could set this up as a github action?)

AngieHinrichs:
If github actions can be triggered by releases of cov-lineages/pangoLEARN, all the better.  :)

trvrb:
I was producing the alphabetical lineage list by just going to https://cov-lineages.org/lineages.html, copying the whole page and cleaning up quickly in BBEdit. Took just a couple minutes. That said, a couple points:
1. If the `pango_lineage` entries in `coloring_ordering.tsv` are a bit stale, it's not the end of the world. We'll get a mostly nicely grouped picture with some gray dots thrown in. I think this is definitely an upgrade to existing color jumble.
2. I do think that scripting this would be good. However, I think the right strategy lies entirely in Augur / ncov. All we need to do is pull `pango_lineage` from the filtered metadata column and construct `colors.tsv` from this. No need to import from pangoLEARN etc... I could imagine and option in `assign-colors.py` that takes in a list of metadata columns that should just have simple alphabetical entries in the `colors.tsv` output.

I would like to plan to merge this PR as it stands and implement (2) in a subsequent PR.

PR #592, opened 2021-03-24
Enable workflows based on curated global context sequences

huddlej:

This PR proposes the following two changes to the ncov workflow to address the issue of how to start a custom SARS-CoV-2 analysis without starting from the full GISAID database of >800,000 sequences.

#### Prepare global context sequences for upload to S3

During the course of standard Nextstrain builds, output a list of strains for each subsampling job and use these lists to aggregate the final FASTA file of subsampled sequences while also exporting a metadata file for those sequences. Then, add the Nextstrain global build's subsampled sequences to the list of files to upload to S3. With the sequences and metadata available on S3, our team can start to use these as a global context for our own bespoke analyses.

We include the build name in the subsampled sequence/metadata outputs to distinguish these data from other data in the current flat S3 bucket naming approach (e.g., `global_subsampled_sequences.fasta` is clearer than `subsampled_sequences.fasta`).

#### Add support for workflows starting with curated contextual inputs

Modifies the path-or-URL logic for multiple inputs so users can start from a local "filtered" input file (e.g., one downloaded from GISAID instead of a remote file on S3) that represents a set of contextual sequences to use for a new analysis. Adds an example profile, `example_global_context`, with a build that contextualizes the example Australian samples with a curated global subsampled set. The new build's inputs look like this:

```yaml
inputs:
  - name: aus
    metadata: "data/example_metadata_aus.tsv"
    sequences: "data/example_sequences_aus.fasta"
  - name: global
    metadata: "data/global_subsampled_metadata.tsv.gz"
    filtered: "data/global_subsampled_sequences.fasta.gz"
```

Assuming that the curated inputs will be compressed when downloaded by the user, this PR also enables reading compressed sequences when combining and deduplicating multiple inputs.

A potentially useful alternative to the approach above would allow users to inject their global context sequences into the workflow _after_ all other subsampling (at the `combine_samples` step) to avoid combining and de-duping, realigning, etc. This functionality could be supported with another input keyword like `subsampled` like so:

```yaml
inputs:
  - name: aus
    metadata: "data/example_metadata_aus.tsv"
    sequences: "data/example_sequences_aus.fasta"
  - name: global
    metadata: "data/global_subsampled_metadata.tsv.gz"
    subsampled: "data/global_subsampled_sequences.fasta.gz"
```

### Testing

As a sanity check, I first prepared contextual sequences locally with the CI build using the following command:

```bash
snakemake \
  results/global/global_subsampled_sequences.fasta \
  results/global/global_subsampled_metadata.tsv \
  -j 4 \
  --profile nextstrain_profiles/nextstrain-ci/
```

Then, I prepared contextual sequences with full Nextstrain global build on the cluster with:

```bash
snakemake \
  results/global/global_subsampled_sequences.fasta \
  results/global/global_subsampled_metadata.tsv \
  -j 4 \
  --profile nextstrain_profiles/nextstrain/ \
  --config active_builds=global
```

Next, I compressed the global subsampled sequences and metadata from the full global build with gzip and copied these into my `data` directory on my local system (as `data/global_subsampled_sequences.fasta.gz` and `data/global_subsampled_metadata.tsv.gz`). Normally, these files would be automatically uploaded to our S3 bucket by the standard Nextstrain `upload` rule.

I ran the `example_global_context` build locally with these local curated contextual sequences.

```bash
snakemake -j 4 -p --profile my_profiles/example_global_context/
```

Finally, I uploaded the compressed global context files to the Nextstrain S3 bucket and updated the `example_global_context` build's inputs to look like:

```yaml
inputs:
  - name: aus
    metadata: "data/example_metadata_aus.tsv"
    sequences: "data/example_sequences_aus.fasta"
  - name: global
    metadata: "s3://nextstrain-ncov-private/global_subsampled_metadata.tsv.gz"
    filtered: "s3://nextstrain-ncov-private/global_subsampled_sequences.fasta.gz"
```

Then, I ran this example workflow again with the S3 inputs:

```bash
snakemake --forceall -j 4 -p --profile my_profiles/example_global_context/
```

### DIFF
--- a/my_profiles/example_global_context/builds.yaml
+++ b/my_profiles/example_global_context/builds.yaml
@@ -1,0 +1,52 @@
+
+## This YAML file is sparsely commented, with a focus on the parts relevant to multiple inputs
+## See my_profiles/example/builds.yaml for more general comments
+## See docs/multiple_inputs.md for a walkthrough of this config.
+
+# custom_rules:
+#   - my_profiles/example_multiple_inputs/rules.smk
+
+use_nextalign: true
+
+inputs:
+  - name: aus
+    metadata: "data/example_metadata_aus.tsv"
+    sequences: "data/example_sequences_aus.fasta"
+  - name: global
+    metadata: "data/global_subsampled_metadata.tsv.gz"
+    filtered: "data/global_subsampled_sequences.fasta.gz"
+    # TODO: ideally, we could support something like the following, to inject
+    # a collection of contextual sequences into the analysis after all other
+    # subsampling.
+    # subsampled: "data/global_subsampled_sequences.fasta.gz"
+
+builds:
+  global-context:
+    subsampling_scheme: custom-scheme # use a custom subsampling scheme defined below
+
+# STAGE 1: Input-specific filtering parameters
+filter:
+  aus:
+    min_length: 5000 # Allow shorter genomes. Parameter used to filter alignment.
+    skip_diagnostics: True # skip diagnostics (which can remove genomes) for this input
+
+# STAGE 2: Subsampling parameters
+subsampling:
+  custom-scheme:
+    # Use metadata key to include ALL from `input1`
+    australian_focus:
+      exclude: "--exclude-where 'aus!=yes'" # subset to sequences from input `aus`
+    # Include all global contextual sequences without subsampling.
+    global_context:
+      exclude: "--exclude-where 'aus=yes'" # i.e. subset to sequences _not_ from input `aus`
+
+files:
+  auspice_config: "my_profiles/example_global_context/my_auspice_config.json"
+  description: "my_profiles/example_global_context/my_description.md"
+
+skip_travel_history_adjustment: True
+
+traits:
+  global-context:
+    sampling_bias_correction: 2.5
+    columns: ["country"]
--- a/my_profiles/example_global_context/builds.yaml
+++ b/my_profiles/example_global_context/builds.yaml
@@ -9,9 +9,9 @@
 use_nextalign: true
 
 inputs:
-  - name: aus
-    metadata: "data/example_metadata_aus.tsv"
-    sequences: "data/example_sequences_aus.fasta"
+  - name: local
+    metadata: "data/example_metadata.tsv"
+    sequences: "data/example_sequences.fasta"
   - name: global
     # Define local paths to a pre-filtered global context.
     metadata: "data/global_subsampled_metadata.tsv.gz"
--- a/my_profiles/example_global_context/builds.yaml
+++ b/my_profiles/example_global_context/builds.yaml
@@ -13,8 +13,12 @@ inputs:
     metadata: "data/example_metadata_aus.tsv"
     sequences: "data/example_sequences_aus.fasta"
   - name: global
+    # Define local paths to a pre-filtered global context.
     metadata: "data/global_subsampled_metadata.tsv.gz"
     filtered: "data/global_subsampled_sequences.fasta.gz"
+    # Paths to files on S3 also work. For example:
+    #metadata: "s3://nextstrain-ncov-private/global_subsampled_metadata.tsv.xz"
+    #filtered: "s3://nextstrain-ncov-private/global_subsampled_sequences.fasta.xz"
     # TODO: ideally, we could support something like the following, to inject
     # a collection of contextual sequences into the analysis after all other
     # subsampling.
--- a/docs/change_log.md
+++ b/docs/change_log.md
@@ -3,6 +3,13 @@
 As of April 2021, we use major version numbers (e.g. v2) to reflect backward incompatible changes to the workflow that likely require you to update your Nextstrain installation.
 We also use this change log to document new features that maintain backward compatibility, indicating these features by the date they were added.
 
+## v3 (12 April 2021)
+
+  - Use Augur 11.2.0's metadata-only output arguments to aggregate subsampled sequences and metadata [#592][]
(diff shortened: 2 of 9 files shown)


tsibley (approved):
I haven't given this a detailed review, just a cursory one, but I talked about it just now with @huddlej and @eharkins on a call. It seems good to go for me after the trial build completes succesfully.

rneher:
this looks good to me. I haven't tested it in detail, but the proposed changes make sense and I don't think they'll break other builds. 

I noticed that you changed the subsampling logic slightly to use `strains.txt` and `augur filter` to do the de-duping. Seems sensible to me. 

regarding data sets. For North America this can be immediately useful. for other continents, this will be a rather sketchy skeleton initially.

rneher (approved):

my_profiles/example_global_context/builds.yaml
```diff
@@ -0,0 +1,56 @@
...
+inputs:
+  - name: aus
+    metadata: "data/example_metadata_aus.tsv"
+    sequences: "data/example_sequences_aus.fasta"
+  - name: global
+    # Define local paths to a pre-filtered global context.
+    metadata: "data/global_subsampled_metadata.tsv.gz"
+    filtered: "data/global_subsampled_sequences.fasta.gz"
+    # Paths to files on S3 also work. For example:
+    #metadata: "s3://nextstrain-ncov-private/global_subsampled_metadata.tsv.xz"
```
trvrb, line 20: Files currently on S3 are `gz` and not `xz`.

trvrb:
When I tried to test this I ran into the following. I used the `inputs` specified in the PR which included:
```
  - name: local
    metadata: "data/example_metadata.tsv"
    sequences: "data/example_sequences.fasta.gz"
```
where `sequences` is given as a compressed file. This resulted in running:

```
nextalign                 --jobs=1                 --reference defaults/reference_seq.fasta                 --genemap defaults/annotation.gff                 --genes S                 --sequences data/example_sequences.fasta.gz                 --output-dir results/translations                 --output-basename seqs_local                 --output-fasta results/aligned_local.fasta                 --output-insertions results/insertions_local.tsv
```
directly on the compressed file and ending up with an empty output file `results/aligned_local.fasta`.

If I swap `inputs` to:
```
  - name: local
    metadata: "data/example_metadata.tsv"
    sequences: "data/example_sequences.fasta"
```
this rule runs as it should and the workflow continues.

trvrb:
Other than this hiccup everything worked for me in my testing. Though a couple broader suggestions:

1. It seems reasonable to me to want to upload all 7 primary datasets as reference sets rather than just global. This would create S3 files `s3://nextstrain-ncov-private/oceania_subsampled_metadata.tsv.xz`, etc... Looks like this may be just additional entries here https://github.com/nextstrain/ncov/pull/592/files#diff-834a01151d9fee74a508cf26f793d2e96ee9c167853154abc0154e0fa04d911fR179? If I'm running an analysis of (say) B.1.526 in New York, I'd prefer a ~4000 sequence North America context to a ~4000 sequence global context. This doesn't have to be part of this PR.
2. I think we should start with uploading GISAID-derived files to `nextstrain-ncov-private` as this PR does, but we can use this same logic to upload to `nextstrain-ncov-public` bucket from public builds once these exist.

trvrb (approved):

my_profiles/example_global_context/builds.yaml
huddlej: Good catch! The new subsampled files will have the `.xz` extension, since [the upload rule uses this compression for sequence data](https://github.com/nextstrain/ncov/blob/3ef559a1e975a1ef10cc301b87b056308ac74d89/workflow/snakemake_rules/export_for_nextstrain.smk#L180-L184), too. You can see example outputs from the trial build run here:

```bash
$ aws s3 ls s3://nextstrain-ncov-private/trial/99d07b91a09eb3d05e220e93aca0e84dd0f09bbe/
2021-04-01 14:39:26   70963528 aligned_gisaid.fasta.xz
2021-04-01 14:48:42   43403016 filtered_gisaid.fasta.xz
2021-04-01 14:42:35     218436 flagged-sequences_gisaid.tsv.xz
2021-04-01 14:51:34     202784 global_subsampled_metadata.tsv.xz
2021-04-01 14:51:33     308348 global_subsampled_sequences.fasta.xz
2021-04-01 14:44:17   50371144 masked_gisaid.fasta.xz
2021-04-01 14:51:22   27834876 mutation_summary_gisaid.tsv.xz
2021-04-01 14:42:29   19595104 sequence-diagnostics_gisaid.tsv.xz
2021-04-01 14:42:35     100472 to-exclude_gisaid.txt.xz
```

Moving forward, I'd really like us to compress our sequences with bgzip so we can use samtools faidx (benefits of bgzip indexing outweighing benefits of slightly better compression from LZMA). The bgzip compression would produce a `.gz` extension for all of these uploaded files instead.

huddlej:
Thank you for testing this out, @trvrb! I agree about the strange behavior of the workflow with one compressed input vs. multiple inputs. Users should not be expected to know that there is a distinction there. I've added an issue for this at #600.

> It seems reasonable to me to want to upload all 7 primary datasets as reference sets rather than just global. This would create S3 files s3://nextstrain-ncov-private/oceania_subsampled_metadata.tsv.xz, etc... Looks like this may be just additional entries here [...snip...]

That's right. We'd replace the current hardcoded paths with an `expand` function call that takes a list of the regions we want to upload.

Since the current PR is working, but it makes a backward incompatible change of requiring Augur 11.3.0, I'll push an update to the change log and make a new release after merging.

PR #595, opened 2021-03-30
Calculate change in frequency for nodes in tree

### LINKED ISSUE #594: Annotate change in frequency for clades in Auspice JSONs
**Context**

One way to identify potentially concerning variants or clades is to calculate the change in frequency for those variants/clades over recent history. For influenza analyses, we've developed a script to perform these calculations and annotate them as "delta frequency" over the last 6 months.

**Description**

For SC2 analyses, we want a similar delta frequency analysis over a shorter time-span. This analysis can use KDE tip frequencies as inputs and a configuration parameter that determines how far back in the past to look for the previous frequency of each strain.

We should add a script and Snakemake rule to calculate delta frequency and also provide a configuration parameter so users can choose the past timepoint. The resulting outputs should be a node data JSON that gets exported into the final auspice JSON as a color-by attribute.

huddlej:

Adds a script, Snakemake rule, and configs to calculate the change in
frequency (i.e., "delta frequency") for nodes in a given tree over a
fixed period of time. The default time period is 4 weeks which
corresponds to 4 timepoints or "pivots" in the tip frequencies data. The
workflow exports this new `delta_frequency` annotation in auspice JSONs
for visualization as a color-by or scatterplot attribute.

### Related issue(s)  

Fixes #594

### Testing

Tested by CI and with the Nextstrain "global" build. Examples include scatterplot view:

image

And the tree view:

image

Based on the tree view, we may want to limit the delta frequency calculations to internal nodes instead of calculating them for all tips.

### DIFF
--- a/defaults/auspice_config.json
+++ b/defaults/auspice_config.json
@@ -102,6 +102,16 @@
       "key": "delta_frequency",
       "title": "Change in frequency",
       "type": "continuous"
+    },
+    {
+      "key": "logistic_growth",
+      "title": "Logistic growth",
+      "type": "continuous"
+    },
+    {
+      "key": "current_frequency",
+      "title": "Current frequency",
+      "type": "continuous"
     }
   ],
   "geo_resolutions": [
--- a/defaults/auspice_config.json
+++ b/defaults/auspice_config.json
@@ -99,11 +99,6 @@
       "type": "categorical"
     },
     {
-      "key": "delta_frequency",
-      "title": "Change in frequency",
-      "type": "continuous"
-    },
-    {
       "key": "logistic_growth",
       "title": "Logistic growth",
       "type": "continuous"
--- a/defaults/auspice_config.json
+++ b/defaults/auspice_config.json
@@ -97,6 +97,11 @@
       "key": "region_exposure",
       "title": "Region of exposure",
       "type": "categorical"
+    },
+    {
+      "key": "delta_frequency",
+      "title": "Change in frequency",
+      "type": "continuous"
     }
   ],
   "geo_resolutions": [
--- a/defaults/parameters.yaml
+++ b/defaults/parameters.yaml
@@ -110,8 +110,8 @@ logistic_growth:
 
   # Set the minimum and maximum current frequency a clade must for its logistic
   # growth to be calculated.
-  min_frequency: 0.0001
-  max_frequency: 0.5
+  min_frequency: 0.000001
+  max_frequency: 0.95
 
 #
 # Region-specific settings
--- a/defaults/parameters.yaml
+++ b/defaults/parameters.yaml
@@ -108,8 +108,9 @@ logistic_growth:
   # is calculated.
   min_tips: 50
 
-  # Set the maximum current frequency a clade must for its logistic growth
-  # to be calculated.
+  # Set the minimum and maximum current frequency a clade must for its logistic
+  # growth to be calculated.
+  min_frequency: 0.0001
   max_frequency: 0.5
 
 #
--- a/defaults/parameters.yaml
+++ b/defaults/parameters.yaml
@@ -102,15 +102,15 @@ logistic_growth:
   # Calculate logistic growth over the last N pivots which corresponds to N
   # times the amount of time represented by the `pivot_interval_units` in the
   # frequencies configuration.
-  delta_pivots: 4
-
-  # Set the frequency threshold for logistic growth calculations, to avoid
-  # numerical errors when calculating logit transformations.
-  frequency_threshold: 0.0001
+  delta_pivots: 6
 
   # Set the minimum number of tips a clade must have before its logistic growth
   # is calculated.
-  min_tips: 25
+  min_tips: 50
+
+  # Set the maximum current frequency a clade must for its logistic growth
+  # to be calculated.
+  max_frequency: 0.5
 
 #
 # Region-specific settings
--- a/defaults/parameters.yaml
+++ b/defaults/parameters.yaml
@@ -104,6 +104,25 @@ delta_frequency:
   # frequencies configuration.
   delta_pivots: 4
 
+  # Set the minimum number of tips a clade must have before its change in
+  # frequency is calculated.
+  min_tips: 25
+
+# Logistic growth settings
+logistic_growth:
+  # Calculate logistic growth over the last N pivots which corresponds to N
+  # times the amount of time represented by the `pivot_interval_units` in the
+  # frequencies configuration.
+  delta_pivots: 4
+
+  # Set the frequency threshold for logistic growth calculations, to avoid
+  # numerical errors when calculating logit transformations.
+  frequency_threshold: 0.0001
+
+  # Set the minimum number of tips a clade must have before its logistic growth
+  # is calculated.
+  min_tips: 25
+
 #
 # Region-specific settings
 #
--- a/defaults/parameters.yaml
+++ b/defaults/parameters.yaml
@@ -97,17 +97,6 @@ frequencies:
   stiffness: 20
   inertia: 0.2
 
-# Delta frequency settings
-delta_frequency:
-  # Calculate change in frequency over the last N pivots which corresponds to N
-  # times the amount of time represented by the `pivot_interval_units` in the
-  # frequencies configuration.
-  delta_pivots: 4
-
-  # Set the minimum number of tips a clade must have before its change in
-  # frequency is calculated.
-  min_tips: 25
-
 # Logistic growth settings
 logistic_growth:
   # Calculate logistic growth over the last N pivots which corresponds to N
--- a/defaults/parameters.yaml
+++ b/defaults/parameters.yaml
@@ -97,6 +97,13 @@ frequencies:
   stiffness: 20
   inertia: 0.2
 
+# Delta frequency settings
+delta_frequency:
+  # Calculate change in frequency over the last N pivots which corresponds to N
+  # times the amount of time represented by the `pivot_interval_units` in the
+  # frequencies configuration.
+  delta_pivots: 4
+
 #
 # Region-specific settings
 #
--- a/my_profiles/example/my_auspice_config.json
+++ b/my_profiles/example/my_auspice_config.json
@@ -6,11 +6,6 @@
   ],
   "colorings": [
     {
-      "key": "delta_frequency",
-      "title": "Change in frequency",
-      "type": "continuous"
-    },
-    {
       "key": "location",
       "title": "Location",
       "type": "categorical"
--- a/my_profiles/example/my_auspice_config.json
+++ b/my_profiles/example/my_auspice_config.json
@@ -6,6 +6,11 @@
   ],
   "colorings": [
     {
+      "key": "delta_frequency",
+      "title": "Change in frequency",
+      "type": "continuous"
+    },
+    {
       "key": "location",
       "title": "Location",
       "type": "categorical"
--- a/my_profiles/example_advanced_customization/auspice_config_swiss.json
+++ b/my_profiles/example_advanced_customization/auspice_config_swiss.json
@@ -82,11 +82,6 @@
       "key": "region_exposure",
       "title": "Region of exposure",
       "type": "categorical"
-    },
-    {
-      "key": "delta_frequency",
-      "title": "Change in frequency",
-      "type": "continuous"
     }
   ],
   "geo_resolutions": [
--- a/my_profiles/example_advanced_customization/auspice_config_swiss.json
+++ b/my_profiles/example_advanced_customization/auspice_config_swiss.json
@@ -82,6 +82,11 @@
       "key": "region_exposure",
       "title": "Region of exposure",
       "type": "categorical"
+    },
+    {
+      "key": "delta_frequency",
+      "title": "Change in frequency",
+      "type": "continuous"
     }
   ],
   "geo_resolutions": [
--- a/my_profiles/example_advanced_customization/auspice_config_switzerland_build.json
+++ b/my_profiles/example_advanced_customization/auspice_config_switzerland_build.json
@@ -82,11 +82,6 @@
       "key": "region_exposure",
       "title": "Region of exposure",
       "type": "categorical"
-    },
-    {
-      "key": "delta_frequency",
-      "title": "Change in frequency",
-      "type": "continuous"
     }
   ],
   "geo_resolutions": [
--- a/my_profiles/example_advanced_customization/auspice_config_switzerland_build.json
+++ b/my_profiles/example_advanced_customization/auspice_config_switzerland_build.json
@@ -82,6 +82,11 @@
       "key": "region_exposure",
       "title": "Region of exposure",
       "type": "categorical"
+    },
+    {
+      "key": "delta_frequency",
+      "title": "Change in frequency",
+      "type": "continuous"
     }
   ],
   "geo_resolutions": [
--- a/my_profiles/example_multiple_inputs/my_auspice_config.json
+++ b/my_profiles/example_multiple_inputs/my_auspice_config.json
@@ -82,11 +82,6 @@
       "key": "region_exposure",
       "title": "Region of exposure",
       "type": "categorical"
-    },
-    {
-      "key": "delta_frequency",
-      "title": "Change in frequency",
-      "type": "continuous"
     }
   ],
   "geo_resolutions": [
--- a/my_profiles/example_multiple_inputs/my_auspice_config.json
+++ b/my_profiles/example_multiple_inputs/my_auspice_config.json
@@ -82,6 +82,11 @@
       "key": "region_exposure",
       "title": "Region of exposure",
       "type": "categorical"
+    },
+    {
+      "key": "delta_frequency",
+      "title": "Change in frequency",
+      "type": "continuous"
     }
   ],
   "geo_resolutions": [
--- a/nextstrain_profiles/nextstrain-ci/builds.yaml
+++ b/nextstrain_profiles/nextstrain-ci/builds.yaml
@@ -32,3 +32,9 @@ subsampling:
       priorities:
         type: "proximity"
         focus: "region"
+
+# Set CI-specific growth settings, such that the small tree still gets logistic
+# growth values and we properly test the growth calculations script.
+logistic_growth:
+  min_tips: 0
+  max_frequency: 1.0
--- a/nextstrain_profiles/nextstrain/africa_auspice_config.json
+++ b/nextstrain_profiles/nextstrain/africa_auspice_config.json
@@ -31,6 +31,11 @@
       "type": "continuous"
     },
     {
+      "key": "delta_frequency",
+      "title": "Change in frequency",
+      "type": "continuous"
+    },
+    {
       "key": "location",
       "title": "Location",
       "type": "categorical"
--- a/nextstrain_profiles/nextstrain/africa_auspice_config.json
+++ b/nextstrain_profiles/nextstrain/africa_auspice_config.json
@@ -36,6 +36,16 @@
       "type": "continuous"
     },
     {
+      "key": "logistic_growth",
+      "title": "Logistic growth",
+      "type": "continuous"
+    },
+    {
+      "key": "current_frequency",
+      "title": "Current frequency",
+      "type": "continuous"
+    },
+    {
       "key": "location",
       "title": "Location",
       "type": "categorical"
--- a/nextstrain_profiles/nextstrain/africa_auspice_config.json
+++ b/nextstrain_profiles/nextstrain/africa_auspice_config.json
@@ -31,11 +31,6 @@
       "type": "continuous"
     },
     {
-      "key": "delta_frequency",
-      "title": "Change in frequency",
-      "type": "continuous"
-    },
-    {
       "key": "logistic_growth",
       "title": "Logistic growth",
       "type": "continuous"
--- a/nextstrain_profiles/nextstrain/asia_auspice_config.json
+++ b/nextstrain_profiles/nextstrain/asia_auspice_config.json
@@ -36,6 +36,16 @@
       "type": "continuous"
     },
     {
+      "key": "logistic_growth",
+      "title": "Logistic growth",
+      "type": "continuous"
+    },
+    {
+      "key": "current_frequency",
+      "title": "Current frequency",
+      "type": "continuous"
+    },
+    {
       "key": "location",
       "title": "Location",
       "type": "categorical"
--- a/nextstrain_profiles/nextstrain/asia_auspice_config.json
+++ b/nextstrain_profiles/nextstrain/asia_auspice_config.json
@@ -29,7 +29,12 @@
       "key": "S1_mutations",
       "title": "S1 mutations",
       "type": "continuous"
-    },    
+    },
+    {
+      "key": "delta_frequency",
+      "title": "Change in frequency",
+      "type": "continuous"
+    },
     {
       "key": "location",
       "title": "Location",
--- a/nextstrain_profiles/nextstrain/asia_auspice_config.json
+++ b/nextstrain_profiles/nextstrain/asia_auspice_config.json
@@ -31,11 +31,6 @@
       "type": "continuous"
     },
     {
-      "key": "delta_frequency",
-      "title": "Change in frequency",
-      "type": "continuous"
-    },
-    {
       "key": "logistic_growth",
       "title": "Logistic growth",
       "type": "continuous"
--- a/nextstrain_profiles/nextstrain/builds.yaml
+++ b/nextstrain_profiles/nextstrain/builds.yaml
@@ -267,3 +267,7 @@ subsampling:
       priorities:
         type: "proximity"
         focus: "region_late"
+
+# Define frequencies parameters.
+frequencies:
+    recent_days_to_censor: 7
--- a/nextstrain_profiles/nextstrain/europe_auspice_config.json
+++ b/nextstrain_profiles/nextstrain/europe_auspice_config.json
@@ -36,6 +36,16 @@
       "type": "continuous"
     },
     {
+      "key": "logistic_growth",
+      "title": "Logistic growth",
+      "type": "continuous"
+    },
+    {
+      "key": "current_frequency",
+      "title": "Current frequency",
+      "type": "continuous"
+    },
+    {
       "key": "location",
       "title": "Location",
       "type": "categorical"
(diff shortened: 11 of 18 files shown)


huddlej:
The build from the example screenshots above is [available in the staging bucket and through the Nextstrain development server](https://dev.nextstrain.org/staging/ncov/global-with-dfreq?l=scatter&scatterY=delta_frequency).

trvrb:
And separate from the issue of exactly how to calculate growth rate from vector of node (clade) frequencies, there is an issue of how to treat tips. I see that you've made the decision to not give attach a growth rate to tips. This is justified in a lot of ways, however, I _think_ it would be better to cast growth rates to tips. Take a look at this zoomed portion of B.1.1.7 for example:

Screen Shot 2021-03-31 at 11 45 33 AM

Every internal branch gets an estimate of growth rate, but the final branch subtending each tip does not. These 2 tip internal branches will be about as wonky as single tip branches. This causes it to look like the internal bits of B.1.1.7 aren't growing, but they really are.

Here, I'd set a tip count threshold (perhaps something like 10) in which only internal nodes with 10 or more subtending tips would have the logistic growth directly calculated. Other branches (with 9 or fewer subtending nodes) would inherit the closest calculated "upstream" growth rate estimate. This will emphasize single-tip branches coming from basal expansions, but I think this is okay.

I'd also calculate current frequency for tips in addition to 2-plus-subtending-tip internal nodes. This will highlight the contribution of more recent tips.

Both these will result in (I believe) more useful scatterplots where you can actually mouseover tips.

huddlej:
> This PR looks like it does a straight-forward additive change in frequency? Ie going from 5% to 10% in 4 weeks would get labeled as dfreq = 0.65?

Yes, this is just the original dfreq formulation from flu, so 5 to 10% in 4 weeks (0.0833 years) would be `(0.1 - 0.05) / 0.0833` or ~0.6 delta frequency.

> I would recommend a Fisher-Pry transform. [...snip...]
> I'd set a tip count threshold (perhaps something like 10) in which only internal nodes with 10 or more subtending tips would have the logistic growth directly calculated. Other branches (with 9 or fewer subtending nodes) would inherit the closest calculated "upstream" growth rate estimate. [...snip...]
> I'd also calculate current frequency for tips in addition to 2-plus-subtending-tip internal nodes.

Ok, I’ve added this logistic transform and regression to the delta frequency script as an option along with a user-configurable option to limit these calculations to nodes with at least 10 tips. The script propagates the values of parents to nodes that are too small to have values calculated. I reannotated the original tree that I’ve shown in the PR with a `logistic_growth` field and also calculated the `current_frequency` field for all nodes and tips. I’ll show screenshots below of [this tree that is also available on the staging server](https://dev.nextstrain.org/staging/ncov/global-with-dfreq?c=logistic_growth&l=scatter&scatterX=current_frequency). The tree still includes the original `delta_frequency` or "Change in frequency" annotation.

[Here is the tree view colored by logistic growth with a minimum of 10 tips required for the growth calculation](https://dev.nextstrain.org/staging/ncov/global-with-dfreq?c=logistic_growth).

image

One issue here is that the color scale is linear, so growth >3 looks the similar across the whole tree and it is difficult to identify clades at the higher end of the distribution. The coloring of tips also obscures the signal of internal clades even if those tips have the same color as their parents. Would it be difficult to implement a log color scale in Auspice or should we define our own scale in a different way through a config file?

Next, we can look at [the scatterplot of logistic growth by sample date](https://dev.nextstrain.org/staging/ncov/global-with-dfreq?c=logistic_growth&l=scatter).

image

This view is complicated by older tips that have inherited the logistic growth values of their parent clades. We also see there is a wide range of growth values with the most extreme values representing small clades (those with <10 tips). This pattern seems consistent with the logit transformation’s greater emphasis on small changes at the tails of the input distribution. Just for comparison, I repeated these growth annotations but without propagating values from parents to children and got the following view that more clearly highlights rapidly growing smaller clades like the one at the top. I understand the need to include tip annotations in the tree view, though. For this kind of analysis, I wish that the scatterplot view had a way to toggle tips on an off the way it can toggle branches.

image

Back to the original tree with tip annotations, if we [plot growth by current frequency](https://dev.nextstrain.org/staging/ncov/global-with-dfreq?c=logistic_growth&l=scatter&scatterX=current_frequency), as shown below, we still see the wide range of growth values on the left, but we can also more clearly identify medium-sized clades (by current frequency) that have larger growth values (like 501Y.V1).

image

We can use this view to zoom into a recent 501Y.V1 subclade with higher growth. From the tree and map views, we can see that this growing clade has a declining subclade with a ORF1b:P218L mutation in West Africa and a growing subclade with a N:R204P mutation in Australia and central Europe.

image

One issue that you hinted at before is that propagating growth values to smaller clades can give the appearance that these small clades are doing as well as their parents when we really don’t know. This pattern is easier to see with the linear dfreq measurement and linear color scale for the following clade where its very small subclades appear to have the highest change in frequency, but they’ve really inherited their parent’s value.

image

Finally, [here is the correlation between the logistic growth values and linear change in frequency values](https://dev.nextstrain.org/staging/ncov/global-with-dfreq?c=logistic_growth&l=scatter&regression=show&scatterX=delta_frequency). For early detection of small clades that are rapidly growing, the logistic growth values provide a much better signal.

image

If this looks like a better direction, I'll update all the Auspice config files to reference these new logistic growth and current frequency fields.

rneher:
Thanks a lot for pushing this, @huddlej !

The large negative values after log-transform could be driven by the KDE smoothing kernel and will depend rather sensitively on the value of `pc`. Maybe we should clamp the range of valid values. 

The behavior of frequency trajectories near one are likely never a perfect logistic. Geographic and demographic penetration will slow a variant down before reaching 100%. But logit should still be a better transform than the simple log that we have used in the past. log transform has the advantage though that variants will receive smaller growth values once they become common (growth from 90 to 95% is arguably not terribly important). 

Overall, frequencies are normalized such that they sum to one at each pivot, right? I have been thinking of using the explicit derivative of your smoothing kernel, but this is complicated by unequal weighing pivot by pivot.

huddlej:
> The large negative values after log-transform could be driven by the KDE smoothing kernel and will depend rather sensitively on the value of `pc`. Maybe we should clamp the range of valid values.

That makes sense. For the figures above, I used the default value from the diffusion frequencies code (`pc=1e-4`). With `pc=1e-3`, we get the following range:

image

With `pc=5e-3`, we get the following range:

image

> Overall, frequencies are normalized such that they sum to one at each pivot, right? I have been thinking of using the explicit derivative of your smoothing kernel, but this is complicated by unequal weighing pivot by pivot.

That's right. We could always switch to a more flexible implementation though.

trvrb:
Thanks for all the detailed work here @huddlej. Starting with a replication check using different methodology to check that final numbers make sense. Here, I looked at the base of B.1.1.7 which has a logistic growth rate of 4.0 in your analysis (https://dev.nextstrain.org/staging/ncov/global-with-dfreq?c=logistic_growth&f_clade_membership=20I/501Y.V1&l=scatter). It looks like this is measured in `num_date` units and so represents a logistic growth rate of 4.0 per year.

If I take the same 3984 strains present in your dataset and run my logistic rate estimate I get:

frequencies-B 1 1 7

global-projection

This rate of 0.04 per day corresponds to a rate of 13.0 per year, so about 3X what you have. But can be maybe explained by timespan, if we limit to Jan 15 to Feb 15 we get:

global-projection

(This is a little funny of comparison in that the data goes until Feb 15 in this example, but the final pivot is March 22, so the final 5 weeks in the PR example is almost all projection from the KDE.)

I followed up here by directly looking at the final 4 pivots of clade 501Y.V1 via Auspice which are 27%, 29%, 30%, 32%. Logit transform followed by linear regression gives a slope and logistic growth rate of 1.1. 

I think I need to run things myself to dig in deeper here to figure out what's going on with the difference between values of 1.1 and 4.0, but do you have any thoughts on what may be causing this John?

trvrb:
Also, I think we're going to want 6 pivots (weeks) back. If you look at the latest live dataset, the final 2 weeks have ~30 genomes and the 2 weeks before this have ~330. I'm comfortable making estimates using 4 weeks of data, but the most recent two weeks shouldn't count for much here. We see consistently that the now cast of B.1.1.7 frequency based on KDEs is less than really frequency once the data comes in. For example: https://nextstrain.org/ncov/global/2021-03-16 has pivot from 2021-02-28 at 36%, while https://nextstrain.org/ncov/global/2021-03-01 has this same pivot at 25%.

Either in addition to or instead of this, we could also pull back the latest pivot we calculate in general. This may be more principled actually. As we don't trust the current estimate. Maybe lop off the final pivot in current tip frequencies and move to 5 pivots back for growth estimates?

huddlej:
> (This is a little funny of comparison in that the data goes until Feb 15 in this example, but the final pivot is March 22, so the final 5 weeks in the PR example is almost all projection from the KDE.)

I think this is a bug with how I ran the build on March 22 but did not download the latest GISAID data. Hopefully, this isn't going to be an issue in production, but I'll doublecheck.

> I followed up here by directly looking at the final 4 pivots of clade 501Y.V1 via Auspice which are 27%, 29%, 30%, 32%. Logit transform followed by linear regression gives a slope and logistic growth rate of 1.1 per year.
>
> I think I need to run things myself to dig in deeper here to figure out what's going on with the difference between values of 1.1 and 4.0, but do you have any thoughts on what may be causing this John?

It seems like the issue might be differences in how we calculate our date units. One small point is that the delta frequency script asks for "4 pivots back from the last pivot" to get the frequency 4 weeks back, so we end up with 5 data points to fit.

I attached a Jupyter notebook showing the frequencies and pivots for 501Y.V1 and how the model fitting works for those data:

 - [2021-04-05-summarize-logit-transformed-frequencies-for-501Y.V1.pdf](https://github.com/nextstrain/ncov/files/6260086/2021-04-05-summarize-logit-transformed-frequencies-for-501Y.V1.pdf)

In your example of 32% latest frequency and earliest of 27% frequency, the delta time between those pivots is `0.064`. The logit transform of those values is `logit(0.32) = -0.75` and `logit(0.27) = -0.99`. So, the slope between those two points is `(-0.75 - (-0.99)) / 0.064 = 3.75`. I've rounded all of the inputs here, but in the more precise calculations in the notebook the value is closer to 3.93.

> Either in addition to or instead of this, we could also pull back the latest pivot we calculate in general. This may be more principled actually. As we don't trust the current estimate. Maybe lop off the final pivot in current tip frequencies and move to 5 pivots back for growth estimates?

That is a great idea. With this change, then the `--delta-pivots` argument of the current script would change to be "pivots back from the last pivot" and we'd add an argument like `--drop-last-pivots` that defaults to 1?

trvrb:
Thanks for the well thought out notebook @huddlej. This really makes things clear. Seeing how close to perfect fit the linear regression is on the logit scale is really comforting. There's perhaps a bit of evidence for the final pivot to be a slightly below expectation, but it's really not as big of an effect as I was expecting:

Screen Shot 2021-04-05 at 2 49 41 PM

> With this change, then the `--delta-pivots` argument of the current script would change to be "pivots back from the last pivot" and we'd add an argument like `--drop-last-pivots` that defaults to 1?

I was proposing an "upstream" change so that when `tip-frequencies.json` is calculated we'd move from `--max-date` of "today" to 1 week back from today here: https://github.com/nextstrain/ncov/blob/master/workflow/snakemake_rules/common.smk#L150. Ie we'd want to correct Auspice view so that the current most recent pivot isn't shown as we don't trust this pivot.

That said, the current B.1.1.7 example looks pretty spot-on without the final pivot being obviously off. I'm not sure I'd worry about this at the moment.

Given notebook, I'm quite convinced about the logistic growth calculations and think the current parameters (4 pivots back, so 5 pivots total) are probably good.

trvrb:
Also, thanks for all the work implementing the tip count cutoff for calculation of growth rate. I was a bit surprised by this behavior here:

Screen Shot 2021-04-05 at 3 07 04 PM

where the highest growth rate clade does not have anything special about it. It's just a clade of 10 tips with no nucleotide mutations and nothing particularly special about it.

Screen Shot 2021-04-05 at 3 08 11 PM

It looks like it just happens to have 10 recent tips associated with it that are sampled and so there is apparent clade growth:

Screen Shot 2021-04-05 at 3 12 35 PM

Maybe the only recourse here is to have a larger cutoff value? Like 15 instead of 10? Because even if we didn't propagate internal node growth rates to the tips, this would still register as a internal node with 10 tips with high growth rate.

huddlej:
> I was proposing an "upstream" change so that when tip-frequencies.json is calculated we'd move from --max-date of "today" to 1 week back from today 

Ah, that seems much nicer! 

> Maybe the only recourse here is to have a larger cutoff value? Like 15 instead of 10? Because even if we didn't propagate internal node growth rates to the tips, this would still register as a internal node with 10 tips with high growth rate.
>
> Or perhaps dropping the final pivot would help in this scenario? I can imagine these small clades being influenced by a particular very recent sample pulling up the final pivot.

I bet dropping the final pivot would help. Initially, I tried setting a minimum current frequency of 1% instead of a fixed number of tips (after talking with @kistlerk about her similar work). But it turned out that viewing these data with [current frequency on the x-axis](https://dev.nextstrain.org/staging/ncov/global-with-dfreq?c=logistic_growth&l=scatter&scatterX=current_frequency) helped focus on clades that are large enough to consider their high logistic growth as more relevant without applying any filters during the analysis.

I will re-run the analysis with the latest available data so the last pivot and available samples are more closely matched. We can see from those results whether we need to modify any filter thresholds or alter the pivots.

trvrb:
Thanks John. I had run this trail build https://github.com/nextstrain/ncov/actions/runs/717668977 from commit https://github.com/nextstrain/ncov/pull/595/commits/53053e2462870baaf9ad784dfb76aa1a1bca3ce1. You should be able to work from this:
```
nextstrain build --aws-batch --attach fb03df82-6055-4f0c-90ff-45329d713f5c .
```
I haven't looked at the results here yet.

huddlej:
@trvrb I ran a fresh build with the latest available GISAID data, to see if logistic growth annotations were less likely to pick up spurious signals from KDE distributions. As a sanity check, I inspected the frequency distribution of strains by their submission date. Although there were strains in the tree from early April (a couple of days ago), 48% of frequencies at the last timepoint came from strains submitted >7 days ago and 21% submitted more than 1 month ago.

image

This distribution suggests that calculating frequencies from one week back as the maximum date would be a better implementation. Next, I looked at the logistic growth values by sampled date and current frequency. The same issue that we noticed in the earlier analysis appeared here where the clade with the highest logistic growth was a small one (N=10 strains) with two more recent samples whose KDE distributions most likely drove the apparent recent growth.

image

To address this issue, I added a configuration parameter for frequencies called `recent_days_to_censor` that allows users to dynamically set the maximum date for frequency estimation such that the most recent days are excluded.

Without defining this parameter, its value defaults to 0 and the max date is `2021.2616438356165`. When I set this parameter value to 7 days, the max date is `2021.2424657534248`. I recalculated frequencies and logistic growth and still got the same results with the highest growth from the small clade above. I could move the max date back even farther, but this would seem to throw away information about larger clades that we care about.

(As an aside, while I was inspecting the frequencies command, I noticed that the KDE bandwidth is 0.05. At 0.00274 years per day, this bandwidth corresponds to 18 days. That seems kind of wide now that we have such densely sampled trees.)

Other options to tune parameters and avoid these spurious signals would be to:

* increase the numerical precision threshold from 1e-3 to 1e-2 to limit extreme growth values
* increase the minimum number of tips required for growth calculations
* set a minimum current clade frequency for growth calculation
* limit growth calculations to branches with amino acid mutations

I reverted the frequencies censoring back to original behavior and then I tried increasing the numerical precision threshold to 1e-2, as a way to avoid censoring data while limiting extreme logistic growth values. This change set an upper threshold of ~12 for the upper end and ~-7 for the lower end. The resulting tree view more clearly highlights recently growing clades.

image

The highest growth clades with current frequencies >1% are actually reasonable clades with many amino acid mutations.

image

Some larger clades with high growth rates did not have amino acid mutations.

image

Their growth appeared to be driven by smaller subclades with a couple of recent samples.

image

This result suggests that censoring the last couple of days in frequency calculations could still be helpful. Adding censoring of the last 7 days back to the frequency calculations, the same larger clade still has a relatively high growth rate, but the subclade that appeared to be driving that growth is no longer solely responsible now that its two latest strains do not contribute to frequencies. 

image

For what it is worth, this approach of censoring the last 7 days for frequency calculations and increasing the numerical precision threshold seems to produce the clearest signal from the tree view, too.

image

I’ve pushed [this latest tree to the staging server](https://dev.nextstrain.org/staging/ncov/global-with-dfreq?l=scatter&scatterX=current_frequency&scatterY=logistic_growth), if you want to explore it more.

trvrb:
Nice progress here John. However, I do think that the shrinkage from using `pc=1e-2` is too strong. See for example clade 20H/501Y.V2:

Screen Shot 2021-04-06 at 9 45 57 PM

The base of the clade is estimated to be growing as expected from just looking at frequency plot. However, subclades within it quickly dive to 0. My desired behavior would be a tree that's much more blocky. Ie we expect subclades within 20H/501Y.V2 to be growing at the same rate as the parent clade unless we have good evidence otherwise.

I'd suggest to go in the direction of something like `pc=1e-4` but then having a tip count cutoff of something like ~25.

huddlej:
Ok, I see what you mean, @trvrb. Here is [the same view for 501Y.V2 using `pc=1e-4` and `min_tips=25`](https://nextstrain.org/staging/ncov/global-with-dfreq?c=logistic_growth&f_clade_membership=20H/501Y.V2&l=scatter&label=clade:20H/501Y.V2).

image

I added the frequency threshold (`pc`) as a command line argument to the delta frequency script and as an option in the Snakemake config.

trvrb:
Thanks John! This is behaving rather nicely I think. I'm quite happy with the "blockiness" of the tree where moderately sized clades are being estimated to have similar logistic growth rates. A couple observations:

The clade with the highest logistic growth rate is a subclade of 20I/501Y.V1. And generally as expected subclades of 20I/501Y.V1 rank above much of other circulating viruses. You can see the growth of 501Y viruses here: https://nextstrain.org/staging/ncov/global-with-dfreq?c=gt-S_501&d=tree,frequencies&l=scatter&p=full&scatterY=logistic_growth, which is really quite striking.

501Y

I'm mostly okay keeping "Change in frequency" in, but it does have some odd behavior and I prefer how logistic growth is working. For example take a look at this parent clade of 20H/501Y.V2.

delta-freq

You can see that this parent clade (part of 20C in green) is estimated at higher "Change in frequency" than the 20H subclade (in orange). However, I think this just because this 20C parent clade is at higher frequency and getting carried along by a rising 20H subclade. If we look at logistic growth it nests in as expected and is mostly lower than 20H.

logistic

My only remaining complaint is color ramp, where for logistic growth the vast majority of nodes have values greater than -10, but because the minimum value is around -30 the color ramp is compressed to basically be oranges and reds. I think we should supply a min and max in the Auspice config JSON to ramp from -10 to 10.

trvrb:
As specified in the above commit I decided that linear delta frequency was just too confusing. For example, parent clades of B.1.1.7 were consistently getting assigned higher values of delta frequency. Generally, I think is too biased to assigning high values to high frequency clades, which is exactly what we don't want to target.

I spent a fair bit playing around with different parameters for logistic growth calculation. There were a few broad considerations:

1. There is a trade off between catching growth of biological distinct newly arisen clades and geographic sampling artifacts that cause a clade to look like it's suddenly taking off. My best strategy here was to increase the minimum clade count to 50 subtending tips.
2. The other sampling effect was to have a clade that's generally declining, but then to get a bump due to sampling of some recent sequences. The best strategy I came up with here was to increase delta pivots from 4 to 6. This increase did not hide important new clades, but instead dealt with older clades "curving" in their trajectories.
3. A major failure mode was basal clades that take off (or have subclades that carry them like 20B) and then having subclades that branch directly off of the basal clade that don't hit the minimum thresholds. Due to propagating down parent estimates to daughter nodes, this caused a bunch of dead 20B clades to have some of the highest logistic growth estimates. I dealt with this by having a low 50% threshold for "max frequency" above which a clade receives a logistic growth estimate of `nan`. This causes old clades in the tree to be grayed out in Auspice and I think is generally an okay strategy, but it would be nice if there was a way to fix these non-propagating clades more directly.
4. Some old clades have completely died out near the present. With pseudocounts applied to the logit transform, these dead basal clades were assigned logistic growth rates of `0`, which placed them in the middle of the pack, when they really should be ranked poorly. To deal with this, I added a criteria that sets logistic growth rate to `nan` for clades that have current frequency of less than 0.0001. This basically removes the need for finessing of pseudocounts (`pc`) passed to the logit transform function.

I'm pretty happy with the results. By my eye the trees generally have the appropriate behavior with known clades and also with clicking around to look directly at frequency behavior.

Results are available at:
- global: https://nextstrain.org/staging/ncov/global/add-delta-frequency?c=logistic_growth
- Africa: https://nextstrain.org/staging/ncov/africa/add-delta-frequency?c=logistic_growth
- Asia: https://nextstrain.org/staging/ncov/asia/add-delta-frequency?c=logistic_growth
- Europe: https://nextstrain.org/staging/ncov/europe/add-delta-frequency?c=logistic_growth
- North America: https://nextstrain.org/staging/ncov/north-america/add-delta-frequency?c=logistic_growth
- Oceania: https://nextstrain.org/staging/ncov/oceania/add-delta-frequency?c=logistic_growth
- South America: https://nextstrain.org/staging/ncov/south-america/add-delta-frequency?c=logistic_growth

trvrb:
I'm sure this could use some further finessing, but I'm happy enough with things as they stand that I'd now be okay merging. Thanks for all the detailed work on this PR John.

trvrb (approved):

rneher:
I think this generally works very nicely. But it does generate some odd dead branches:
image


I am wondering whether a more robust approach would be to consider only clades with tips in the last 3 months or so.

trvrb:
Thanks Richard. Your example here is the edge case I was having the most trouble dealing with. It's been difficult to get the small basal clades that don't hit the 50 tip cutoff to register appropriately. If we want the basal 20I/501Y.V1 clades to be appropriately red, it's really difficult not also have the parent clades to 20I also be inappropriately red. The actual parent clade to 20I/501Y.V1 should have high logistic growth, but small offshoots to this parent lineage should not.

That said, a cutoff for only calculating logistic growth rates for clades with tips sampled in the last ~3 months seems like it could be a generally helpful thing. Worth experimenting with.

Also, I see that the merged PR is failing the CI tests. Should look into this as well.

huddlej:
> Also, I see that the merged PR is failing the CI tests. Should look into this as well.

@trvrb, I rebased this branch on to master so we're up-to-date with the new CI dataset on master. 2f875d6 fixes the CI bug and 41e60ea enables logistic growth for the CI build by making the internal node requirements less strict.

 > I'm sure this could use some further finessing, but I'm happy enough with things as they stand that I'd now be okay merging.
> ...
> That said, a cutoff for only calculating logistic growth rates for clades with tips sampled in the last ~3 months seems like it could be a generally helpful thing. Worth experimenting with.

Do you want to continue experimenting in this branch or merge and then continue experimenting separately?

It seems like a higher minimum current frequency threshold (instead of a high minimum number of tips) would provide a way to get at the "clades with tips in the last 3 months" approach. 2130500 adds the minimum frequency as config parameter that we could tune for this purpose.

trvrb:
Thanks John! I just slightly updated the logic that applies the minimum frequency cutoff so that it effectively only calculates logistic growth for clades with sampled tips in the previous three months.

I've updated the outputs here: 
- global: https://nextstrain.org/staging/ncov/global/add-delta-frequency?c=logistic_growth
- Africa: https://nextstrain.org/staging/ncov/africa/add-delta-frequency?c=logistic_growth
- Asia: https://nextstrain.org/staging/ncov/asia/add-delta-frequency?c=logistic_growth
- Europe: https://nextstrain.org/staging/ncov/europe/add-delta-frequency?c=logistic_growth
- North America: https://nextstrain.org/staging/ncov/north-america/add-delta-frequency?c=logistic_growth
- Oceania: https://nextstrain.org/staging/ncov/oceania/add-delta-frequency?c=logistic_growth
- South America: https://nextstrain.org/staging/ncov/south-america/add-delta-frequency?c=logistic_growth

I think this is a definite improvement and there's no longer confusion about what it means to have positive logistic growth of a clearly dead clade. I'm quite confident we should merge this now.

trvrb:
This did reveal one idiosyncrasy to how `augur frequencies` currently works. I'll describe here and it can become a separate issue if you agree that behavior should be updated. Running the above results in situations like:

recent

where very recent tips are gray. These are generally viruses sampled in the past week. I looked into things and these viruses (like "Australia/NSW4424/2021" in global above) have a collection date after the `--max-date` handed to `augur frequencies`. This results in a frequencies vector in `tip-frequencies.json` that looks like:
```
 "frequencies": [
      0.0,
      0.0,
      0.0,
      ...
      0.0,
      0.0,
      0.0
]
```
with all zeros. This is not what I would have expected. I'd expect `--max-date` to define the final pivot in the frequencies vector but I'd expect `--metadata` passed in with dates after this date to still be counted. Same logic for `--min-date`. 

I don't think this makes a big difference however. But was surprising to me and slightly shifts clade frequencies.

trvrb:
(I just realized I was supposed to update the change log as part of PR with this as a feature. I'll edit `master` instead. Sorry about that.)

batson:
Love this new feature, @trvrb  and @huddlej .

For use cases with time periods where the absolute number of samples is low and/or the variant is at very high or very low frequency, a weighted-least squares regression in logit-space may be more appropriate. (The strategies you have taken in terms of thresholding may also address these issues, but it may be worth trying for your offline analyses @trvrb.)

The loss function would be

Sum (f(x_i) - y_i)^2/(var(y_i|x_i)^2

where x = week, y = logit(frequency of variant in week), and var(y|x) is due to finite sample effects, rescaled by the squared derivative of logit. This is implemented in [statsmodels](https://www.statsmodels.org/stable/examples/notebooks/generated/wls.html) and [sklearn](https://scikit-learn.org/stable/modules/generated/sklearn.linear_model.LinearRegression.html#sklearn.linear_model.LinearRegression.fit)

This can help with time windows lacking good data and also with low/high probabilities, while still being as fast as linear regression.

Back-of-the-envelope:

**Finite sample correction**: if you observe k out of n, then the variance of k/n is 

`k(n-k)/(n*n*(n+1))`

(If n is very low, you may want to use a beta(1/2,1/2) prior, which amounts to adding 0.5 to k and 1 to n.)

**Jacobian of logit** the logit will rescale a frequency x to log(x) - log(1-x), with derivative 1/x + 1/(1-x). Plugging in the observation k/n to this formula gives n/k + n/(n-k).

Put this all together to get a variance in the logit observation of approximately 

var(y|x) = k(n-k)/n * (1/k + 1/(n-k))^2

When k is small and n is large, this is about 1/k; when k = n/2, this is about 4/n.

The appropriate weights for the regression would be the reciprocals of those. So an error at small k would be weighted by about k, where an error for large k would be weighted by about n/4.

PR #596, opened 2021-04-02
Add change log

huddlej:

[Based on discussions in Slack about versioning](https://bedfordlab.slack.com/archives/C7SDVPBLZ/p1616698682018200), this PR proposes the use of a single change log to track major and minor releases of the ncov workflow for major and feature releases, respectively.

The updated change log here references a v2.0 of the workflow which we will create as a GitHub release after merging this PR.

The release process I'm proposing here would look like:

1. Propose a new PR that adds a feature and/or backward-incompatible change.
2. Include an update to the change log in the PR with the proposed new version and date and any Nextstrain tools that need to be updated (e.g., "Requires Augur >=11.3.0").
3. Merge the PR
4. Make a GitHub major or minor release

If this seems like too much of a hassle for feature releases, we can limit this process to just breaking changes.

### Related issue(s)  

Related to #592 where the proposed changes will require Augur 11.3.0 for the first time and could break workflows for users with older versions installed.


rneher (commented):
Thanks @huddlej! I think this is excellent!

trvrb:
Thanks @huddlej! I'd initially lean to follow @tsibley's Slack advice (https://bedfordlab.slack.com/archives/C7SDVPBLZ/p1616699017019200?thread_ts=1616698682.018200&cid=C7SDVPBLZ) given that the workflow is generally a continually updated thing.

I definitely don't want to complete minor version updates when:
- updating location defaults (https://github.com/nextstrain/ncov/commit/d8d7ef56a62d900bcedf806c5967e424997226d8)
- updating excluded strains (https://github.com/nextstrain/ncov/commit/d85bb130428fff95fdcffa0f8310f7b8d46e8e1e)

but maybe we could just agree that these don't need version updates.

More of a question are things like:
- adding S1 mutation count as a colorby (https://github.com/nextstrain/ncov/commit/7781e4963fadea7ac7ad0bbbc54677a40e5ef368)
- stream combined sequences to disk (https://github.com/nextstrain/ncov/commit/fd68c7a5f4bc05037e274bfc11476dd6a6a6714e)

which are nice little features and could well be documented as a minor version update, but will add a bit of overhead. However, it would be convenient to be able to glance at the changelog to see what's new.

Definitely agree that things like:
- multiple inputs (https://github.com/nextstrain/ncov/pull/550)
- curated global context (https://github.com/nextstrain/ncov/pull/592)

constitute a major version bump.

So, although initially leaning toward just major versions, after review I do see utility in minor versions (and there doesn't seem to be as many edge cases as I initially expected) and would be happy going in this direction.

A small number of other things:
1. For Augur and Auspice it's been very helpful to have version release scripts to make the version release process easier. With the proposed release process it's going to be easy to forget to make a GItHub release, etc...
2. I like having a `VERSION` file with just release number per @tsibley's advice (https://bedfordlab.slack.com/archives/C7SDVPBLZ/p1616699409020800?thread_ts=1616698682.018200&cid=C7SDVPBLZ).
3. How do we specify dependencies in a more computable fashion? I don't like having to have human text in `change_log.md` that specifies version of Augur (nextalign? MAFFT?), etc... that needs to be used, but I don't have an obviously better light-weight solution.

huddlej:
> So, although initially leaning toward just major versions, after review I do see utility in minor versions (and there doesn't seem to be as many edge cases as I initially expected) and would be happy going in this direction.

Yeah, I was torn on this, too. I don't see as much value in providing minor version tags as in providing a change log so users can discover new features. The tags seemed like one way to alert users to these changes.

> 1. For Augur and Auspice it's been very helpful to have version release scripts to make the version release process easier. With the proposed release process it's going to be easy to forget to make a GItHub release, etc...
> 2. I like having a `VERSION` file with just release number per @tsibley's advice

I was initially hoping to avoid maintaining an additional file since we'd need to write a release script to prevent that file from getting out of sync with the git tags. If we didn't have this separate file, we wouldn't need a separate script because [the GitHub new release form](https://github.com/nextstrain/ncov/releases/new) does everything we need.

That said, I just downloaded the tarball for [our first ncov release](https://github.com/nextstrain/ncov/releases/tag/v1.0), and once you're inside the corresponding ncov directory (which is named `ncov-1.0`), there is no other way to know which version of the code you have. If we see this as a major issue, then we should definitely include a `VERSION` file.

Since our major workflow development tends to happen at the level of a GitHub PR (instead of the command line), it would be nice if the release process tied into our standard PR review process. For example, when we've approved a new feature or breaking change, we could either manually trigger a release through GitHub Action's UI or through a bot that responds to comments (akin to [how Bioconda PRs get merged](https://github.com/bioconda/bioconda-recipes/pull/27446#issuecomment-803277601)). The manually trigger workflow seems like the simplest solution, but I bet @tsibley has ideas about this, too.

Even if we decide to develop a release script, can we still merge this PR and manually make the releases we need to get the global context PR merged? I don't want to hold up other work too much...

> How do we specify dependencies in a more computable fashion? I don't like having to have human text in change_log.md that specifies version of Augur (nextalign? MAFFT?), etc... that needs to be used, but I don't have an obviously better light-weight solution.

The contents of [`workflow/envs/nextstrain.yaml`](https://github.com/nextstrain/ncov/blob/master/workflow/envs/nextstrain.yaml) are the most computer-readable description of our workflow dependencies that gets bundled with the repo itself. If we maintained a separate `Dockerfile` for this repo that used the Nextstrain docker-base as its starting point, that would be a second way to surface those dependencies.

Not everyone uses the `--use-conda` flag to run the workflow, though, so having a textual description of Nextstrain software version dependencies seems helpful to include.

tsibley:
Some quick thoughts.

> Even if we decide to develop a release script, can we still merge this PR and manually make the releases we need to get the global context PR merged? I don't want to hold up other work too much...

+1 for proceeding without automation initially.

>> How do we specify dependencies in a more computable fashion? I don't like having to have human text in change_log.md that specifies version of Augur (nextalign? MAFFT?), etc... that needs to be used, but I don't have an obviously better light-weight solution.
>
> …so having a textual description of Nextstrain software version dependencies seems helpful to include.

+1 for a human-readable textual description as the starting point.

I think any more than that is way out of scope for this PR, but I'm definitely interested in continuing this discussion elsewhere. I've put a bit of thought previously into how `nextstrain build` might satisfy deps without locking into a single ecosystem, both in the context of required versions for our core components like Augur or Nextalign and also the wide world of other software that builds might want to use. Super short answer: I think we can support many existing dep managers (Conda, Docker, npm, etc) co-operating all at the same time with [Buildpacks](https://buildpacks.io/).

trvrb:
Thanks @huddlej and thanks @tsibley for the thoughts. I think okay from my perspective almost merge as is (and develop automation that includes a `VERSION` file at a later date). However, if we proceed with this, I'm going to quickly forget what the version release process looks like. Could you amend `dev_docs.md` to spell out the process for releasing a new version of the workflow?

huddlej:
@trvrb, a767904 updates the dev docs to include [a description of the proposed release process](https://github.com/nextstrain/ncov/blob/a767904/docs/dev_docs.md#releasing-new-workflow-versions) and 387f2d4 adds a custom pull request template for this repo with a release checklist, as a reminder for submitters and reviewers.

huddlej:
The most recent commits reflect our group decision to only make new major version releases for backward incompatible changes to the workflow. These commits also increase the visibility of our versioning process and change log by adding badges to the top-level of the repo's README. These links complement the existing link to our change log in the main README body.

PR #597, opened 2021-04-08
Fix proximity

rneher:
It turns out that the proximity rule wasn't doing what it was supposed to do, for two reasons

 - the forced inclusion of the reference strains in the focal set meant that for many sequences the closest hit is the root, which isn't very useful
 - there was a sign problem related to confusion of what is a high and low priority.

This PR should fix this, but I had to introduce a fudge to make the CI work.

### DIFF
--- a/nextstrain_profiles/nextstrain-ci/builds.yaml
+++ b/nextstrain_profiles/nextstrain-ci/builds.yaml
@@ -6,3 +6,29 @@ inputs:
   - name: gisaid
     metadata: "data/example_metadata.tsv"
     sequences: "data/example_sequences.fasta"
+
+builds:
+  # Override the default Nextstrain European build's subsampling scheme for more
+  # stable subsampling of a fixed dataset in continuous integration tests.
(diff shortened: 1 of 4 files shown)


huddlej (approved):
@rneher I'm not sure I understand the proximity code enough to vouch for the sign changes, etc. but the CI error you got revealed a legitimate bug in the CI test itself. d556216 fixes the CI subsampling bug (the commit message has more details, if you're interested).

9c1af84 changes the focal set script to make an empty focal set an error with a helpful error message to the user. If the user has an empty focal alignment in the best case, running through this part of the workflow is a waste of time for the user and they should skip proximity-based subsampling. In the worst case (like the CI situation), there is a bug in the subsampling that the user should dig into. If you feel this shouldn't be an error, I can revert that change.

An upside to the fix in the CI subsampling is CI now runs in ~1 minute instead of ~8 minutes! :tada:

rneher:
thanks, @huddlej !

PR #602, opened 2021-04-15
Adding New Singleton Design Pattern to Recency

akbhangal:
New change has Refactored exiting   construct-recency-from-submission-date.py file to implement Creational Design Pattern - Singleton Pattern in particular.

This new change of introducing pattern restricts the instantiation of the class to only one object. Hence this act as global point of access to the class.

## Related issue(s)

Fixes # NA
Related to # NA

## Testing

No additional preparation is required to unit test. Just rebuild the whole code and execute it. And this change as no impact on output therefore adding this change should produce the output similar to the previously generate results.

### DIFF
--- a/scripts/construct-recency-from-submission-date.py
+++ b/scripts/construct-recency-from-submission-date.py
@@ -39,7 +39,7 @@ class Recency:
         return Recency.__instance_recency 
     
     def toJson(self):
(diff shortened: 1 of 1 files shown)


rneher:
@akbhangal 

PR #603, opened 2021-04-15
Add New Structural Composite Design Pattern

akbhangal:
New change has Refactored exiting add_labels.py to implement Structural design pattern - Composite Design Pattern.

Goal is to reduce complexity and introduce new pattern that allows composing objects into hierarchical tree like structure and treat with singular and composite entities in same way. Also providing functionality to run methods recursively.

## Related issue(s)

Fixes #NA
Related to # NA

## Testing

No additional preparation is required to unit test. Just rebuild the whole code and execute it. And this change as no impact on output therefore adding this change should produce the output similar to the previously generate results.

### DIFF
--- a/scripts/add_labels.py
+++ b/scripts/add_labels.py
@@ -3,15 +3,34 @@ import json
 from Bio import Phylo
 from collections import defaultdict
 
-def attach_labels(d, labeled_nodes):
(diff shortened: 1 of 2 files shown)


rneher:
@akbhangal

PR #605, opened 2021-04-15
Support GISAID downloads

### LINKED ISSUE #600: Add support for either compressed or uncompressed inputs to alignments
**Context**

[As Trevor points out in this related PR](https://github.com/nextstrain/ncov/pull/592#issuecomment-817063537), we want to support starting the workflow from compressed input sequences. However, neither nextalign nor mafft currently support compressed inputs natively. The only way the current workflow supports compressed sequences is when there are multiple inputs that need to be merged into a single uncompressed output file.

**Description**

Users should be able to define a compressed input FASTA file for one or more inputs of a build and the alignment rules should handle these inputs in the same way they currently handle uncompressed inputs.

**Possible solution**

Ideally, nextalign and mafft would transparently support compressed inputs such that the following commands work without any additional effort from the user or the workflow:

```
nextalign --sequences data/sequences.fasta.gz --reference ...
```

However, this workflow could implement solutions for both aligners by checking for the compression status of the input file and piping its contents into the aligners. For example, nextalign supports reading files from `/dev/stdin` like so:

```
gzip -c -d {input.sequences} | nextalign --sequences /dev/stdin ...
```

One simple (imperfect) check would be to define a Snakemake `params` entry that checks the suffix of the input file and returns an appropriate binary to use for decompression of the input file. In the case that the input is not compressed, the binary could be `cat`. The shell command would look like this:

```
{params.sequence_extraction_command} {input.sequences} | nextalign --sequences /dev/stdin ...
```

huddlej:

This PR minimally modifies the existing workflow to support GISAID "nextregions" downloads (curated contextual sequences from Nextstrain regional builds) and custom selection downloads (e.g., selections of sequences from specific locations like "King County" in Washington State). These changes should also be reflected in updates to our ncov tutorial such that users can get a build setup with minimal effort.

Among the workflow changes are:

  - support for aligning compressed sequences with nextalign or mafft (so users do not need to uncompress sequences and metadata after extracting GISAID tarballs)
  - support for duplicate sequences with preference for first occurrence of the duplicate sequence (analogous to how the script for combining metadata overwrites earlier data with later data, since custom selections may have the same sequences present in the contextual set)
  - support for different submission date annotations in GISAID's downloaded metadata where the date can be set as "undefined"
  - updated example build for "global context" that could eventually replace the "example" and "example_multiple_input" builds as the one we reference in the tutorial

Open questions:

 - [x] Since we cannot store GISAID data in the repo, what data should we use for our "example" data for focal and contextual sequences?
 - [x] Should we strip the leading "hCoV-19/" from all inputs to avoid needing to define new settings throughout the workflow? Or do we switch to using "hCoV-19/" everywhere?

## Related issue(s)

Fixes #600

## Testing

 - [x] Tested by CI
 - [x] Tested locally with both mafft and nextalign using tarballs from North America "nextregions" download and a custom selection download for "King County".
 - [x] [Tested with a trial build of the whole Nextstrain workflow](https://github.com/nextstrain/ncov/actions/runs/807384248)

## Release checklist

This PR introduces backward incompatible changes in that it changes the default build and default subsampling scheme.

 - [x] Determine the version number for the new release by incrementing [the most recent release](https://github.com/nextstrain/ncov/releases) (e.g., "v2" from "v1"). -> v4


cassiawag:
I know this is a WIP, but I was so excited about it that I decided to try adapting it on a branch in the WA ncov builds. I ran into some unexpected behavior in which fasta's are combined prior to subsampling. This may be totally expected, but based on the example which provides specific subsampling settings for the different named inputs, I expected sample combination after subsampling.

huddlej:
> I ran into some unexpected behavior in which fasta's are combined prior to subsampling. This may be totally expected, but based on the example which provides specific subsampling settings for the different named inputs, I expected sample combination after subsampling.

There was definitely a bug in the subsampling rules, although the basic behavior of the workflow merging sequences before subsampling is by design. The workflow currently merges sequences and metadata prior to subsampling, so you can subsample from the combined data if you want to. When the workflow merges metadata, it adds calculated columns for each input dataset (e.g., `focal`, `contextual`, etc.) where the values of those columns are either `yes` or `no` depending on whether a given row comes from the corresponding input. These columns allow you to add a query to the subsampling rules that selects only the data from the input you want.

In the original example that you tried out, the rules looked like this:

```
subsampling:
  custom-scheme:
    focal:
      max_sequences: 100
    contextual:
      max_sequences: 100
```

These rules pulled 100 random sequences from the combined FASTA regardless of the original input file those sequences came from. The rules should have looked like this:

```
subsampling:
  custom-scheme:
    focal:
      max_sequences: 100
      query: --query "focal == 'yes'"
    contextual:
      max_sequences: 100
      query: --query "contextual == 'yes'"
```

These rules first filter the combined sequences to the specific dataset we want and _then_ subsample 100 sequences.

I've updated the rules in the example to match these latter rules with the queries. Thank you for catching this bug, @cassiawag!

fanninpm:
Does this mean that the nextstrain workflow will be compatible with the daily-updated .xz files provided directly by GISAID? (These files are under the "Download packages" section, _not_ the "Genomic epidemiology" section.)

huddlej:
@fanninpm, that's right! This PR will allow us to stream those compressed files directly into the workflow without first decompressing them.

fanninpm:
I just realized that the files I mentioned earlier were actually _tar_ archives that contain a `readme.txt` in addition to our desired payload, and that could potentially get messy.

huddlej:
Ah, yes, we haven't attempted to process tarballs and don't have any plans to. The messiness and ambiguity of that input format is definitely an issue.

fanninpm:
Regarding the .tar.xz files, I came up with a series Bash commands to extract and repackage our desired payload:

```bash
#!/bin/bash

ref_date=$(ls -lt metadata_tsv_*.tar.xz | head -n 1 | grep -Po "\d{4}_\d\d_\d\d")
tar xJOf metadata_tsv_${ref_date}.tar.xz metadata.tsv | gzip > metadata_${ref_date//_/-}.tsv.gz
tar xJOf sequences_fasta_${ref_date}.tar.xz metadata.tsv | gzip > sequences_${ref_date//_/-}.fasta.gz
```

NOTES:
- These only apply to the raw `.tar.xz` files that come directly from the GISAID authorities. No curation steps have been applied.
- You may choose to emit `.xz` files instead of `.gz` files if you're strapped for space (and have plenty of time to spare). (Just remember to replace the `gzip` command with the compression algorithm of your choice.)

huddlej:
By changing the default build from "global" to an "all inputs" build, this PR makes a backward incompatible change and will require a new release. We should prioritize this release over [the "drop backward compatibility" PR](https://github.com/nextstrain/ncov/pull/615) which needs a little more work.

rneher:
thanks John. I think this looks good overall -- though it reinforces my desire to split the core workflow from the bells and whistles before and after. Two quick questions:

1) `sanitize_sequences.py` is only run if there are multiple inputs, right?

2) wouldn't this

https://github.com/nextstrain/ncov/blob/support-gisaid-downloads/my_profiles/example_global_context/builds.yaml#L40

be more intuitive as 

`"--exclude-where 'origin != focal'" `

and just have one metadata column that specifies the origin?

huddlej:
Thank you for checking this out, @rneher!

> `sanitize_sequences.py` is only run if there are multiple inputs, right?

The sanitize sequences scripts gets run [just before alignment](https://github.com/nextstrain/ncov/blob/64cd8b8f2e0ce043c730aa4cb86750b25bbfe5cd/workflow/snakemake_rules/main_workflow.smk#L75) and also [just before combining and deduplicating sequences](https://github.com/nextstrain/ncov/blob/64cd8b8f2e0ce043c730aa4cb86750b25bbfe5cd/workflow/snakemake_rules/main_workflow.smk#L364). The pre-alignment step is what allows this to work for a single input and the pre-dedup step is what allows this to work for multiple inputs from different entry points.

It is possible that a user could start the workflow from either the `aligned` or `masked` entry point of `inputs` such that the sequences are not sanitized before running `filter` and include/exclude strain names would not be recognized. We use the same include/exclude files later in the `subsample` step (after strain names have been standardized), but this second use wouldn't prevent "include" strains from getting excluded during the `filter` step.

Just to be safe, we _could_ sanitize sequences before each entry point (e.g., before the diagnostic and mask steps for an `aligned` starting point and before the filter step for the `masked` entry point). But, in our actual use cases, we start the workflow from later entry points with sequences from our own workflow (files uploaded to S3 as part of running the Nextstrain profile). These files will have already gone through the sanitize step during the original alignment prior to upload to S3.

For the use case this PR addresses with sequences downloaded from GISAID, there is not much extra cost for users to always start from the `sequences` entry point. In this case, the different entry points are not an issue either.

> wouldn't [it] be more intuitive [to have] just have one metadata column that specifies the origin?

This idea was [discussed as part of the original "multiple inputs" PR](https://github.com/nextstrain/ncov/pull/549#pullrequestreview-587098923) where @jameshadfield noted:

> I went with one-hot as I often have the use case where the sequence sets are overlapping -- e.g. some of my "private" sequences are also in GISAID / NCBI, and I couldn't think of how to nicely express this in a single variable whilst facilitating subsampling based on this variable...

We could discuss this more in a separate issue, if the current implementation doesn't generally work well for folks. This would be a great place to do some user testing, actually...

jameshadfield (commented):
I haven't yet tested it, but these changes seem sensible. Really cool how `augur.io` lets us use streams. Brief comments (not really a review):

* Once #615 goes in (very shortly) this will need to be rebased, which will simplify things (e.g. no more conditional alignment rules).
* How close can we get to using xz compression throughout? 
    * Metadata: I think we could make `scripts/sanitize_metadata.py` write compressed output (perhaps by examining the output path). Subsequent rules should handle compressed metadata I believe.
    * Sequences: This PR allows us to use compressed sequence input, but from then on they're all uncompressed. We could have `nextalign` write to a tmp file and compress it all in one rule. I _think_ our downstream rules, e.g. `augur filter` will handle compressed files OK, but some scripts may not. Perhaps we could leave post-subsampled alignments uncompressed to ease the upgrade path?

huddlej:
Thank you for checking out this PR and the detailed comments on #615!

> Once #615 goes in (very shortly) this will need to be rebased, which will simplify things (e.g. no more conditional alignment rules).

I'd like to merge this PR before #615 and release v4 with GISAID support before we make the much bigger changes in #615 with v5. I can imagine people balking a little at the force switch to nextalign but still wanting to use the functionality here. It won't be too much of a pain to rebase the other PR onto this one later.

> How close can we get to using xz compression throughout?

For metadata, we could make the switch any time. For sequences, we need a little more work in Augur `tree`, `refine`, and probably `ancestral` to handle compressed inputs. Not all tree builders handle compressed inputs, so we'd need to stream to stdin where possible. Both `refine` and `ancestral` use TreeTime's internal functions to load alignments, so I'm not sure how these handle compressed files. If we can get to the point where we start workflows with subsampling from a sequence index and metadata, we won't need to worry as much about compressing sequences.

We should also consider using bgzip instead of xz for large files where random access through an index file is helpful.

cassiawag:
This PR doesn't support metadata that is xz compressed. I updated `rule download_metadata` in  `workflow/snakemake_rules/download.smk` to support this. Can push up the commit to the PR if this is desired behavior.

`rule download_metadata:
    message: "Downloading metadata from {params.address} -> {output.metadata}"
    output:
        metadata = "data/downloaded{origin}.tsv"
    params:
        address = lambda w: config["inputs"][_trim_origin(w.origin)]["metadata"],
        deflate = lambda w: _infer_decompression(config["inputs"][_trim_origin(w.origin)]["metadata"])
    shell:
        """
        aws s3 cp {params.address} - | {params.deflate} > {output.metadata:q}
        """`

huddlej:
Great catch, @cassiawag! This looks like [a bug that's been in `master` for a little while now](https://github.com/nextstrain/ncov/blob/7a71f49c76de7d269930ff431b4853abd8378dca/workflow/snakemake_rules/download.smk#L46-L56). Apparently `gunzip` on OS X (and maybe other OSs) can handle LZMA-compressed files (`.xz`) transparently but other implementations don't.

I just replaced the old code with your proposed change, confirmed it worked locally, and pushed it in e917368.

PR #611, opened 2021-04-19
Report Augur versions on error without `--use-conda`

huddlej:

If the workflow errors out and the user isn't using the `--use-conda` Snakemake flag, let the user know what Augur version they have installed and which version is expected in the Conda environment. Let users know about the `--use-conda` flag in this case as a possible solution to the workflow errors.

## Related issue(s)

Related to https://github.com/nextstrain/cli/issues/119

## Testing

Tested by installing an older version of Augur in an external Conda environment that did not support the `--output-strains` argument for `augur filter` (11.1.2 or before) and confirming that the new error message about `--use-conda` got printed when the filter rule failed.

## Release checklist

### DIFF
--- a/Snakefile
+++ b/Snakefile
@@ -8,6 +8,7 @@ from getpass import getuser
 from snakemake.logging import logger
 from snakemake.utils import validate
 from collections import OrderedDict
+import textwrap
 import time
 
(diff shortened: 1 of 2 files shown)


rneher:
@huddlej this might still be useful and has not merge conflicts!!!

PR #613, opened 2021-04-21
Upload regional contexts to S3

huddlej:

Updates the existing global context input of the upload rule to include
all Nextstrain regional builds. As a result, all subsampled sequences
and metadata for each regional build should get uploaded to S3 during
each run of the ncov workflow.

## Related issue(s)

Related to #592

## Testing

 - [x] Trial build through GitHub Actions

## Release checklist

No major updates or features for the public workflow.


huddlej:
The trial build uploaded the new context files to S3 as expected, so I'm merging this now.

PR #620, opened 2021-04-29
Use ordinal scale for S1 mutations

jameshadfield:
Traits with integer values are better displayed by auspice using
an ordinal scale. This will result in a legend with integer
entries rather than our current situation of floats.

image
_Left: this PR, right: current nextstrain.org_

### DIFF
--- a/nextstrain_profiles/nextstrain-genbank/auspice_config.json
+++ b/nextstrain_profiles/nextstrain-genbank/auspice_config.json
@@ -28,7 +28,7 @@
     {
       "key": "S1_mutations",
       "title": "S1 mutations",
-      "type": "continuous"
+      "type": "ordinal"
     },    
     {
       "key": "location",
--- a/nextstrain_profiles/nextstrain/africa_auspice_config.json
+++ b/nextstrain_profiles/nextstrain/africa_auspice_config.json
@@ -28,7 +28,7 @@
     {
       "key": "S1_mutations",
       "title": "S1 mutations",
-      "type": "continuous"
+      "type": "ordinal"
     },
     {
       "key": "logistic_growth",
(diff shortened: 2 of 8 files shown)


trvrb:
What happens when viruses with 20 `s1_mutations` exist? Does the ordinal legend end up as 0, 1, ..., 19, 20? I originally didn't have this an ordinal because I was assuming that ordinal would imply situations where `s1_mutations` entries include just 0, 1, 2, 7, 8 would make color ramp that has the same gradation going from 1 to 2 as from 2 to 7 as in an ordinal scale only rank matters.

I'd think that S1 mutations might be better covered as a continuous variable, but with defined intervals as sketched out in https://github.com/nextstrain/auspice/pull/1340.

huddlej:
This change reminds me of [the opposite change that just happened in the seasonal-flu repo for epitope mutations](https://github.com/nextstrain/seasonal-flu/commit/46f72c3c577d1ae345d177229a5407d08a4c6898#diff-a38a6442c0b203a5f4c306dd2bed6b2be88b2323ea03656a09ea57d8f2f022c5). :)

@trvrb's argument about ordinal values technically only considering rank instead of intervals is convincing ([Vega uses this technical definition of ordinal](https://vega.github.io/vega/docs/scales/#ordinal)). Although, I can imagine that we don't want to have to manually define bounds for each continuous whole-numbered variable. It would be nice if Auspice's automated segmentation of the data range used integer bounds when the inputs are integers. This latter feature is similar to both the "nice" and "bins" options of [Vega's quantitative scales](https://vega.github.io/vega/docs/scales/#quantitative).

On the flip side, a nice feature of using the ordinal values is that one can hover over a specific value in the legend and see the corresponding tips in the tree that match that exact value. With the continuous values, I can't tell as easily what values the highlighted tips have because they are in some range handled opaquely by the legend.

rneher:
I think integer legend values are definitely preferable. We have been using `ordinal` to mean `integer` and to me the most obvious solution would be to have auspice figure out integer bins that don't exceed 12. So if there the range is 0 to 16, we would have bins of size 2...

rneher:
we are at about 40 S1 mutations now, way past the point where integers are useful.

PR #626, opened 2021-05-06
Update location of GenBank data for downloading

tsibley:
Matches [a corresponding change in ncov-ingest](https://github.com/nextstrain/ncov-ingest/pull/164).  This new prefix will be
a consolidated, shared location for "open" build inputs and
intermediates.

### DIFF
--- a/nextstrain_profiles/nextstrain-genbank/builds.yaml
+++ b/nextstrain_profiles/nextstrain-genbank/builds.yaml
@@ -86,8 +86,8 @@ conda_environment: "workflow/envs/nextstrain_with_nextalign.yaml"
 # subsampling
 inputs:
   - name: genbank
-    metadata: "s3://nextstrain-data/ncov-ingest/genbank_metadata.tsv.gz"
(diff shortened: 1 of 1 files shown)


tsibley:
Marking draft for now as we still need to adjust intermediate and reference file locations, which is slightly more involved.

tsibley:
We will also need to discuss migration plans for the existing paths:

    s3://nextstrain-data/ncov-ingest/
    s3://nextstrain-data/ncov-intermediates/
    s3://nextstrain-data/ncov-reference-datasets/

Should we remove them after we switch to the new consolidated location? Or keep them around for some time afterwards? If we keep them around, for how long? They'll become stale, but at least existing builds using them won't suddenly start failing.

huddlej:
@rneher can confirm, but my understanding is that no one is using those buckets for production analyses since these paths have not been widely distributed. I'd vote to remove them so any builds relying on them fail immediately and users (probably only in our group) have to update their paths.

PR #628, opened 2021-05-07
Profile for Open Data builds

tsibley:
Much discussion around this has happened in Slack and Zoom. Some implementation details are in the commit messages.

This is just a draft of the most minimal changes we could make and still have the desired initial structure for the published data files. It's a starting point, but alternatives are definitely up for discussion.

Syntactically all seems ok, but I have not actually tested this by running the builds. That should happen before merge, once we're settled on an implementation.

## Todo

  * [x] What other build-specific files to include? Probably `{build_name}/aligned.fasta`, but what else? We wanted all the final Nextstrain JSON files duplicated under `files/ncov/open/…` as well, right?
  * [x] What uploads currently included should we remove? We want to nail down a reasonable interface we can support over time.
  * [x] The metadata + sequence inputs uploaded by ncov-ingest will often be out of sync with the latest ncov build. We need to decide how to handle this. @trvrb, @emmahodcroft, and myself (@tsibley) weighed in [on Slack](https://bedfordlab.slack.com/archives/C01LCTT7JNN/p1620421048121300).
  * [x] Build filenames don't include "open" vs. "gisaid" prefix, but they need to in order for the auto-deployed files not to conflict. These prefixes could be hardcoded in each _builds.yaml_ or we could adjust the workflow rules which define the filenames. See https://github.com/nextstrain/ncov/pull/643 for the latter.
  * [x] ~@jameshadfield Does the behaviour change to the `upload` rule and `S3_DST_ORIGINS` break any configurations you're running?~ [Nope.](https://github.com/nextstrain/ncov/pull/628#issuecomment-836119259)
  * [x] ~Fetch via https:// through CDN instead of directly from s3://.  We may need to deal with CloudFront invalidations on upload, however (but `nextstrain remote` would do this).~ Let's do this later separately; see issue #662.
  * [x] Run only major Nextstrain regional builds by default (leaving state-level builds for others)
  * [x] Decide whether to define auspice config files for each region's "open" build or not
  * [x] Decide wheher to inherit most build config values from the main `nextstrain` profile instead of repeating config params (e.g., builds, exposure, traits, subsampling)
  * [x] Extend the trial builds workflow to run open builds instead or additionally?
  * [x] Rename the `nextstrain` profile to `nextstrain-gisaid` to parallel `nextstrain-open`.
  * [x] Hardcode top-level redirects /ncov/{global,africa,...} → /ncov/gisaid/{global,africa,...} in nextstrain.org
  * [x] Remove multiple build sizes from the config
  * [x] Modify description of data use for Open builds (@trvrb)
  * [x] Blog post describing these new builds and the public files

## Cleanup

  * [ ] Remove all files under `s3://nextstrain-data/ncov-ingest/`¹
  * [ ] Remove all files under `s3://nextstrain-data/ncov-intermediates/`¹
  * [ ] Remove all files under `s3://nextstrain-data/ncov-reference-datasets/`¹
  * [ ] Remove all files matching `s3://nextstrain-ncov-private/*_gisaid.*`

¹ See [decision](https://github.com/nextstrain/ncov/pull/626#issuecomment-834596932)

## Release checklist

 - [ ] Update `docs/change_log.md` in this pull request to document these changes and the new version number.


tsibley:
@huddlej As we discussed yesterday, please take over this PR and make changes as you see fit!

workflow/schemas/config.schema.yaml
```diff
@@ -18,3 +18,51 @@ properties:
...
+    enum:
+      - gz
+      - xz
+
+  S3_DST_ORIGINS:
+    type: array
+    minItems: 1
+    items:
+      type: string
+      # Note: The {origin} wildcard in workflow rules includes a leading
```
jameshadfield, line 65: PR #615 removed this so that `{origin}` is a "normal" wildcard and the `_` character is specified in the input/output parameters of each rule.

nextstrain_profiles/nextstrain-open/builds.yaml
```diff
@@ -85,20 +85,13 @@ conda_environment: "workflow/envs/nextstrain_with_nextalign.yaml"
...
-    # metadata: "s3://nextstrain-ncov-private/metadata_gisaid.tsv.gz"
-    # sequences: "s3://nextstrain-ncov-private/sequences_gisaid.fasta.gz"
-    # aligned: "s3://nextstrain-ncov-private/aligned_gisaid.fasta.xz"
-    # to-exclude: "s3://nextstrain-ncov-private/to-exclude_gisaid.txt.xz"
-    # masked: "s3://nextstrain-ncov-private/masked_gisaid.fasta.fasta.xz"
-    # filtered: "s3://nextstrain-ncov-private/filtered_gisaid.fasta.xz"
+  - name: open
+    metadata: "s3://nextstrain-data/files/ncov/open/metadata.tsv.gz"
+    sequences: "s3://nextstrain-data/files/ncov/open/sequences.fasta.gz"
+    # aligned: "s3://nextstrain-files/files/ncov/open/aligned.fasta.xz"
```
jameshadfield, line 91: Is this a typo or do we actually have a `nextstrain-files` bucket?

Either way, won't only the metadata & sequences files be available, or are we maintaining an all-open-data alignment etc? My understanding is that this profile will only be able to start from `metadata` and `sequences`, but I may be out-of-date here!

jameshadfield:
> Does the behaviour change to the upload rule and S3_DST_ORIGINS break any configurations you're running?

Nope (I haven't used this functionality)

> What other build-specific files to include?

I would suggest the alignment, filtered alignment and auspice files (the latter of which would also be uploaded as a top-level object in the bucket so it's accessible by nextstrain.org).

workflow/schemas/config.schema.yaml
tsibley: Ah! Rebased onto latest `master` and updated the comment.

rneher:
not sure this is intentional, but the `config.yaml` doesn't have a `cores` specification and errors out if you run things like 

`snakemake --profile nextstrain_profiles/nextstrain-open data/downloaded_open.tsv`

which works in our current profiles. In most real usecases, `--cores` will be supplied though.

rneher:
(I am running this now locally)

rneher:
This ran locally without problems.

huddlej:
> not sure this is intentional, but the config.yaml doesn't have a cores specification and errors out if you run things like

I've struggled with this a bit, too. Maybe we should think about build configs (what the workflow runs) as separate from the Snakemake profiles (how the workflow is run). For example, the `nextstrain` profile is primarily the builds YAML but we have `nextstrain-scicore`, `nextstrain-rhino`, etc. profiles to change how the workflow runs in different environments.

What if we ran our workflows like this:

```bash
# Run Nextstrain builds on AWS Batch (e.g., with 96 cores, 180 GB memory,
# 16 tree threads, etc. currently hardcoded in the ncov-ingest commands).
snakemake --configfile nextstrain/builds.yaml --profile nextstrain/aws/

# Run Nextstrain builds on scicore cluster.
snakemake --configfile nextstrain/builds.yaml --profile nextstrain/scicore/

# Run Nextstrain builds on your old Macbook.
snakemake --configfile nextstrain/builds.yaml --cores 4 --use-conda
```

tsibley:
@huddlej +1 for config files meant only for customizing what is run vs. profiles meant for how to run it in a given computing environment. However, that is maybe out-of-scope for this PR?

jameshadfield:
Notes from running this build today, in addition to the commits pushed to this PR. cc @huddlej

* Each build defines an auspice config file, which is from the main nextstrain (GISAID) profile. The file `nextstrain-profiles/nextstrain-open/auspice_config.json`, while set in the config file, is never used. Observations:
    * Each of these config files is essentially identical. I think this could be simplified into a rule & generated as part of the pipeline.
    * A lot of colourings are missing, e.g. pango lineage, gisaid lineage etc. Some of these we could generate, and some could be removed. 
    * Lots of lat/longs are missing, largely due to metadata curation issues. I suggest removing `location` from the geo resolutions (and colorby), as the majority of data is lacking lat/longs here.
    * (I can fix all these if you're happy with these changes @huddlej?)
* #643 looks like it's soon to go in, which will allow us to define an auspice prefix and avoid URL clashes for display in nextstrain.org

workflow/snakemake_rules/common.smk
```diff
@@ -145,15 +145,37 @@ def _get_max_date_for_frequencies(wildcards):
...
+    # structure, while builds (global, europe, africa, etc) span all
+    # inputs/origins (and thus do).  In our desired outcome, the two kinds of
+    # files are comingled.  Without changing the main workflow, the mismatch
+    # has to be reconciled by the upload rule.  Thus, the upload rule enforces
+    # and uses only single-origin configurations.  How third-wave.
+    #   -trs, 7 May 2021
+    if len(config["S3_DST_ORIGINS"]) != 1:
+        raise Exception(f'The "upload" rule requires a single value in S3_DST_ORIGINS (got {config["S3_DST_ORIGINS"]!r}).')
+
+    origin = config["S3_DST_ORIGINS"][0]
```
jameshadfield, line 162: An alternative approach would be to hardcode that this rule only works for profiles defining a single input, which must be either `gisaid` or `open`. This would allow us to modify which files we upload, potentially starting with a small set of files for `open`. 

(Yes, I realise that I was the one who added the `S3_DST_ORIGINS` key in the first place.)

workflow/snakemake_rules/common.smk
```diff
@@ -145,15 +145,37 @@ def _get_max_date_for_frequencies(wildcards):
...
-    non-empty value.
-
-    Raises a :class:`KeyError` if none of the *keys* exist with a suitable
-    value.
-    """
-    for key in keys:
-        if config.get(key) not in {"", None}:
-            return config[key]
-    raise KeyError(str(keys))
+def _get_upload_inputs(wildcards):
```
jameshadfield, line 148: As the uploads are only intended for internal use (they're only used by the `export_for_nextstrain` rule) would they be better co-located in ` workflow/snakemake_rules/export_for_nextstrain.smk`?

tsibley: I do think they're only for internal use, and my first instinct was to put this `_get_upload_inputs` function alongside the `export_for_nextstrain` rule. But looking more widely, I saw that all input functions seemed to be colocated in `common.smk`, even when they're tightly coupled to a single caller, so I wanted to follow that convention. Happy to change that convention if we want.

tsibley:
Rebased onto latest master. Will add additional commits to start using `auspice_json_prefix`.

nextstrain_profiles/nextstrain/builds.yaml
```diff
@@ -1,3 +1,5 @@
+auspice_json_prefix: ncov_gisaid
```
jameshadfield, line 1: What's the best recommendation here for nextstrain.org redirects?

We can probably hardcode top-level redirects `/ncov/{global,africa,...}` → `/ncov/gisaid/{global,africa,...}`

Do we want to preserve all the old (timestamped) URLs, or change the filenames & add redirects accordingly? (I don't think it's worth it.) Going forward, newly produced timestamped datasets will be of the `ncov/gisaid/:region/YYYY-MM-DD` format, correct?

cc @eharkins

huddlej:
> * Each build defines an auspice config file, which is from the main nextstrain (GISAID) profile. The file nextstrain-profiles/nextstrain-open/auspice_config.json, while set in the config file, is never used. Observations:
>   * Each of these config files is essentially identical. I think this could be simplified into a rule & generated as part of the pipeline.
>   * A lot of colourings are missing, e.g. pango lineage, gisaid lineage etc. Some of these we could generate, and some could be removed.

Based on our recent Slack conversation, it seems like we could move most of the file-based configuration for Auspice into command line arguments that are parameterized through the `builds` definition. For example, I just diff'd the auspice config JSONs for the `nextstrain` profile and they are nearly identical, too.

Africa, Asia, Europe, and South America are identical. North America and Oceania are identical. Global is different.

The second group uses `division` for geo resolution instead of `country`.

```diff
--- nextstrain_profiles/nextstrain/africa_auspice_config.json	2021-04-29 12:55:39.000000000 -0700
+++ nextstrain_profiles/nextstrain/north-america_auspice_config.json	2021-04-29 12:55:39.000000000 -0700
@@ -128,7 +128,7 @@
   "display_defaults": {
     "color_by": "clade_membership",
     "distance_measure": "num_date",
-    "geo_resolution": "country",
+    "geo_resolution": "division",
     "map_triplicate": false,
     "branch_label": "clade",
     "transmission_lines": false
```

The global build enables map triplicate.

```diff
--- nextstrain_profiles/nextstrain/africa_auspice_config.json	2021-04-29 12:55:39.000000000 -0700
+++ nextstrain_profiles/nextstrain/global_auspice_config.json	2021-04-29 12:55:39.000000000 -0700
@@ -129,7 +129,7 @@
     "color_by": "clade_membership",
     "distance_measure": "num_date",
     "geo_resolution": "country",
-    "map_triplicate": false,
+    "map_triplicate": true,
     "branch_label": "clade",
     "transmission_lines": false
   },
```

It seems like we could allow users to specify a `geo_resolutions` per build (passing the value in the config to `augur export v2 --geo-resolutions`) and collapse all of these JSONs except the global into a single file. We could also add a flag to `augur export v2` for `--map-triplicate` and parameterize that from the builds. This would allow us to use a single auspice config JSON for all builds. Then we could more easily maintain the available color-by options by only needing to update a single config file.

Does that approach sound reasonable? I haven't diff'd the JSONs from this `open` branch yet, but I imagine a similar approach might work here...

nextstrain_profiles/nextstrain/builds.yaml
tsibley: Yep. We'll want to hardcode top-level redirects.

I don't recall discussion about redirects or not for the dated URLs, but I think it would be ok to leave them as-is. We can always move them and add redirects later if we change our mind.

The date stamping suffixes will remain the same and happen for both GISAID and Open builds.

tsibley:
Added commits to make trial builds launch both GISAID and Open, [for example](https://github.com/nextstrain/ncov/actions/runs/886309881), which will look like this in the GitHub Actions UI:

image

I've canceled both on the AWS Batch side, as I don't think we need to run them through quite yet, but we may want to do that before merge.

nextstrain_profiles/nextstrain-open/auspice_configuration.smk
```diff
@@ -0,0 +1,93 @@
...
+rule make_auspice_config:
+    message: "Making auspice config file {output.auspice_config}"
+    output:
+        auspice_config = "results/{build_name}/auspice_config.json"
+    run:
+        import json
+        from copy import deepcopy
+        config = deepcopy(base_config)
+        config["display_defaults"]["geo_resolution"] = "division" if (wildcards.build_name.startswith("north-america") or wildcards.build_name.startswith("oceania")) else "country"
+        config["display_defaults"]["map_triplicate"] = wildcards.build_name.startswith("global")
```
huddlej, line 91: This type of workflow configuration might be clearer and easier to discover and maintain if it lived in the builds configuration instead of Snakemake logic. For example, we can get nearly the same outcomes as this custom `.smk` file produces with a configuration file like this:

```yaml
# Define a default Auspice configuration for open builds.
files:
  auspice_config: nextstrain_profiles/nextstrain/open/auspice_config.json

# Define builds.
builds:
  global:
    subsampling_scheme: nextstrain_region_global
    # Override the default auspice config where map_triplicate=false.
    map_triplicate: true

  africa:
    subsampling_scheme: nextstrain_region
    region: Africa

  asia:
    subsampling_scheme: nextstrain_region
    region: Asia

  europe:
    subsampling_scheme: nextstrain_region
    region: Europe

  north-america:
    subsampling_scheme: nextstrain_region
    region: North America
    # Override the default Auspice config value of geo_resolution=country.
    geo_resolution: division

  oceania:
    subsampling_scheme: nextstrain_region
    region: Oceania
    # Override the default Auspice config value of geo_resolution=country.
    geo_resolution: division

  south-america:
    subsampling_scheme: nextstrain_region
    region: South America
```

In this approach, the auspice config JSON already exists as a valid config that will produce reasonable outputs for any builds and we don't need a custom rule to create it per build. We pass through arguments to `augur export v2` from the builds config to override the JSON from the command line. This approach moves the bulk of this `.smk` file into a JSON file and leaves the custom title building to be resolved.

nextstrain_profiles/nextstrain-open/builds.yaml
```diff
@@ -1,198 +1,107 @@
...
+# (Note: the build size `standard` is not appended, i.e. the original build name remains unchanged.)
+# Each build will have it's own subsampling scheme, which is a copy of the one specified by
+# the build but with the build size appended (e.g. build `global` specifies `nextstrain_region_global`,
+# build `global-xsmall will specify `nextstrain_region_global_xsmall`).
+# The copy of this scheme will be the same as defined in this file with one difference:
+# any `max_sequences` parameters less than 1.0 will be multiplied by the value supplied below. 
 build_sizes:
   standard: 4000
   small: 2000
   xsmall: 1000
```
huddlej, line 29: I'm not convinced that we need three different sizes per build. I understand that this was a feature of the original implementation, but I don't know the specific users who benefit from these different builds that wouldn't just as easily benefit from a single standard build.

I'm also worried that automating build definitions like we do here encourages us to hardcode configuration-level details in Snakemake source code (as in the `auspice_configuration.smk` above or in the [`subsampling_ranges.smk`](https://github.com/nextstrain/ncov/blob/master/nextstrain_profiles/nextstrain/subsampling_ranges.smk)). Just like any other software we develop, hardcoding configuration details in the source code makes those details difficult to discover and debug. In both of the `.smk` files I mention above, the corresponding functionality could be implemented as simpler YAML configuration if `augur` provided a couple of additional flags for `export` and `filter`, respectively.

As programmers, I think we see a hardcoded list of 50 states (or 7 regions in this case) and think we should automate the solution. In practice, a verbose configuration file is just as easy to generate once and then maintain as a static file. The frequency these files need to be updated does not require automated tooling to produce the updated files. A good example of this is [the SPHERES US builds file](https://github.com/CDCgov/spheres-augur-build/blob/master/spheres_profile/builds.yaml) that I generated once and that hasn't needed modification since.

My general thinking here is that workflows like this are complicated enough to follow and debug even in their simplest form. When we add automated configuration to the workflow, we make it that much harder to understand in the future for the short-term benefit of not manually editing verbose configuration files.

huddlej (commented):
@jameshadfield Thanks for pointing me at the auspice config implementation. I made some comments inline for each section. We should discuss more via a call, though. My general concern is that need to avoid putting configuration details in source code. The inline comments below expand on that general idea.

nextstrain_profiles/nextstrain-open/auspice_configuration.smk
jameshadfield: This would also work, and (like my solution) has the big advantage that we don't have a lot of nearly-identical auspice config files hanging around as we currently do for the `nextstrain` profile. If we go this way,  I would propose changing these keys to (e.g.) `default_geo_resolution` as `geo_resolution` is ambiguous. 

Generally, I find a per-profile snakemake file which contains all the auspice config logic (and includes comments) easier to comprehend than a per-profile auspice config JSON which is then modified by rules later on in ways that are hard to remember; however I can understand this is personal preference. 

One advantage of programmatically generating the config files is when we extend their contents to (e.g.) specify custom legend bounds for S1 mutations. This depends on intermediate files (e.g. what's the range of S1 mutations for this build?) which is harder to generate up-front in a config file.

nextstrain_profiles/nextstrain-open/builds.yaml
jameshadfield: Yup! This is why I added 9 lines of comments as I knew I wouldn't remember what was going on next time I looked through the config 😉

nextstrain_profiles/nextstrain-open/auspice_configuration.smk
huddlej: @jameshadfield and I just chatted about this. Here's a summary of our discussion:

  - Dynamically modifying `config["builds"]` and dynamically creating Auspice config JSONs are separate issues here. We primarily discussed the auspice configs. Whether we need multiple build sizes seems to depend on whether there is scientific value in creating these and then the technical solution is another separate issue.
  - We thought of different ways to define build-specific auspice config parameters that override some default
    1. Programmatically build a valid Auspice config JSON with a Snakemake rule that overrides values in a "base config" defined in the rule's scope (top-level of the Snakemake file)
    2. Expose command line arguments for `augur export v2` through the `builds` definitions (as in my examples above) and add any missing arguments to Augur as needed (like `--default-map-triplicate`)
    3. Implement config file overlays in `augur export` such that `--auspice-config` can accept multiple inputs and deep merge these files (analogous to how Snakemake merges multiple inputs to `--configfile`). Define a base config JSON with most of the contents and minimal build-specific JSONs (e.g., defining only `map_triplicates: true`) to be merged with the base config.
    4. Implement config file overlays in the Snakemake workflow with a conditional rule that merges multiple JSONs in the order they are defined. The script we write to handle this logic and the config interface paves the way for migrating that functionality to `augur export`.
  - We agreed that a single concrete auspice config JSON is easier to inspect and understand than a combination of a base auspice config and many overriding command line arguments to `augur export`.
  - We also agreed that our workflow would be more robust if its own configuration (the `builds.yaml`) wasn't bound tightly to Auspice specifics. For example, exposing a `map_triplicate` parameter in the builds config through an Augur command line argument creates a tight link from the workflow to Augur to Auspice. In contrast, if we specify Auspice settings through Auspice config JSONs, the workflow and Augur only need to know about those files.
  - Dynamically generating Auspice legends bounds for S1 mutations, for example, is an interesting case where the implementation requires domain knowledge that Auspice does not have and effectively promotes the "config" to "data" by producing a data-dependent configuration. This might be unavoidably complex, so the question is where to put the complexity.

nextstrain_profiles/nextstrain-open/builds.yaml
rneher: this is why I explored some way to template builds and a combinatorial way:

https://github.com/nextstrain/ncov-simple/blob/master/workflow/snakemake_rules/templated_build.smk

this isn't trying to make an argument for three different sizes, but generating the regions builds only takes 40 lines including full subsampling specs:
https://github.com/nextstrain/ncov-simple/blob/master/profiles/genbank/builds.yaml#L16-L56

nextstrain_profiles/nextstrain-open/auspice_configuration.smk
jameshadfield: Following conversations on Slack, 212b0d3 moves us to a simpler implementation of one auspice config per build. This discussion can shift over to https://github.com/nextstrain/ncov/issues/661 to allow this PR to be merged😄

tsibley:
I created #662 to describe switching downloads from S3 → HTTPS through CloudFront, and marked that todo item above as "complete". We can address it separately.

jameshadfield:
https://github.com/nextstrain/nextstrain.org/pull/350 adds the necessary redirects to nextstrain.org. It can be merged sometime after this PR is merged and a build is completed, so I've marked the todo item as complete.

jameshadfield:
I've rebased this branch onto master to ensure that the builds include the (updated) clades as expected. To test this (and everything, really) I've started [the GitHub workflow with name "2021-06-17"](https://github.com/nextstrain/ncov/actions/runs/945229142)

In conjunction with slack / zoom conversations a number of the todo items are now complete:
- Build config parameter overlap (nextstrain-open vs nextstrain-gisaid) can be addressed via #665
- The rarely used diagnostic files are no longer uploaded
- The auspice dataset JSONs and build-specific alignments are uploaded
- The consensus is that it's ok if the metadata + sequence inputs uploaded by ncov-ingest will be out-of-sync for a few hours with the latest nCoV build.
- The blog post can be started after merge (and merging gives us a good incentive to write it!)

rneher:
I added a preliminary description and still needs the links to the intermediates files.

workflow/snakemake_rules/common.smk
```diff
@@ -164,18 +164,20 @@ def _get_upload_inputs(wildcards):
...
         f"mutation-summary.tsv.xz":       f"results/mutation_summary_{origin}.tsv.xz",       # from `rule mutation_summary`
     }
 
     for build_name in config["builds"]:
         uploads.update({
-            f"{build_name}/sequences.fasta.xz": f"results/{build_name}/{build_name}_subsampled_sequences.fasta.xz",
-            f"{build_name}/metadata.tsv.xz":    f"results/{build_name}/{build_name}_subsampled_metadata.tsv.xz",
+            f"{build_name}/sequences.fasta.xz": f"results/{build_name}/{build_name}_subsampled_sequences.fasta.xz",   # from `rule combine_samples`
+            f"{build_name}/metadata.tsv.xz":    f"results/{build_name}/{build_name}_subsampled_metadata.tsv.xz",      # from `rule combine_samples`
+            f"{build_name}/aligned.fasta":      f"results/{build_name}/aligned.fasta",                                # from `rule build_align`
```
tsibley, line 176: Should we make this `aligned.fasta.xz` to be consistent with everything else, incl. the top-level `aligned.fasta.xz`? Seems like it, to me.

workflow/snakemake_rules/common.smk
```diff
@@ -164,18 +164,20 @@ def _get_upload_inputs(wildcards):
...
+            # export the auspice dataset which matches the subsampled sequences / metadata (see `rule finalize`)
+            f"{build_name}/{config['auspice_json_prefix']}_{build_name}.json":                 f"auspice/{config['auspice_json_prefix']}_{build_name}.json",
+            f"{build_name}/{config['auspice_json_prefix']}_{build_name}_tip-frequencies.json": f"auspice/{config['auspice_json_prefix']}_{build_name}_tip-frequencies.json",
+            f"{build_name}/{config['auspice_json_prefix']}_{build_name}_root-sequence.json":   f"auspice/{config['auspice_json_prefix']}_{build_name}_root-sequence.json"
```
tsibley, line 180: This will produce paths like:

    https://data.nextstrain.org/files/ncov/open/global/ncov_open_global.json
    https://data.nextstrain.org/files/ncov/open/global/ncov_open_global_tip-frequencies.json

What about instead producing paths like this:

    https://data.nextstrain.org/files/ncov/open/global.json
    https://data.nextstrain.org/files/ncov/open/global_tip-frequencies.json

with this mapping?

```python
            f"{build_name}.json":                 f"auspice/{config['auspice_json_prefix']}_{build_name}.json",
            f"{build_name}_tip-frequencies.json": f"auspice/{config['auspice_json_prefix']}_{build_name}_tip-frequencies.json",
            f"{build_name}_root-sequence.json":   f"auspice/{config['auspice_json_prefix']}_{build_name}_root-sequence.json"
```

workflow/snakemake_rules/common.smk
jameshadfield: Typo. Fixed.

workflow/snakemake_rules/common.smk
jameshadfield: Seems sensible - updated this PR with those recommendations.

workflow/snakemake_rules/common.smk
jameshadfield: Misread your original comment @tsibley - I've now resolved this via 69c4f77. I needed to add a new `compress_build_align` rule to generate the `results/{build_name}/aligned.fasta.xz` file which we then upload.

jameshadfield:
One more todo which needs to be done after merge of this PR: `ncov-ingest` [triggers builds](The ncov-https://github.com/nextstrain/ncov-ingest/blob/master/bin/launch-build#L6) from the now-renamed `nextstrain_profiles/nextstrain` profile, and should be updated to use both the `nextstrain-gisaid` and `nextstrain-open` profiles.

workflow/snakemake_rules/common.smk
tsibley: Thanks! Though it looks like the change you made wasn't quite what I meant. With the latest changes on this PR, the URLs are now:

```
https://data.nextstrain.org/files/ncov/open/global/global.json
https://data.nextstrain.org/files/ncov/open/global/global_tip-frequencies.json
```

but I meant URLs like:

```
https://data.nextstrain.org/files/ncov/open/global.json
https://data.nextstrain.org/files/ncov/open/global_tip-frequencies.json
```

Does that seem reasonable?

jameshadfield: It's a judgement call I think - both seem reasonable to me. I like the fact that the auspice JSONs are co-located with the intermediates from which they are generated.

tsibley: Nod. Ah, I like the fact that as `data.nextstrain.org/files/ncov/open/global.json` it more closely matches the `nextstrain.org/ncov/open/global` URL. I'm ok with either way, though. Do others have input? @trvrb?

docs/remote_inputs.md
```diff
@@ -0,0 +1,149 @@
...
+|                  | aligned   | s3://nextstrain-data/files/ncov/open/north-america/aligned.fasta.gz   |
+|                  | auspice   | nextstrain.org/ncov/open/north-america    
+| South-America sample    | metadata  | s3://nextstrain-data/files/ncov/open/south-america/metadata.tsv.gz    |
+|                  | sequences | s3://nextstrain-data/files/ncov/open/south-america/sequences.fasta.gz |
+|                  | aligned   | s3://nextstrain-data/files/ncov/open/south-america/aligned.fasta.gz   |
+|                  | auspice   | nextstrain.org/ncov/open/south-america    
+| Oceania sample   | metadata  | s3://nextstrain-data/files/ncov/open/oceania/metadata.tsv.gz    |
+|                  | sequences | s3://nextstrain-data/files/ncov/open/oceania/sequences.fasta.gz |
+|                  | aligned   | s3://nextstrain-data/files/ncov/open/oceania/aligned.fasta.gz   |
+|                  | auspice   | nextstrain.org/ncov/open/oceania                                |
```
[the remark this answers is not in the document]
trvrb: Would be nice. This would speed downloads and also save on bandwidth charges.

workflow/snakemake_rules/common.smk
```diff
@@ -145,15 +152,39 @@ def _get_max_date_for_frequencies(wildcards):
...
+        f"mutation-summary.tsv.xz":       f"results/mutation_summary_{origin}.tsv.xz",       # from `rule mutation_summary`
+    }
+
+    for build_name in config["builds"]:
+        uploads.update({
+            f"{build_name}/sequences.fasta.xz": f"results/{build_name}/{build_name}_subsampled_sequences.fasta.xz",   # from `rule combine_samples`
+            f"{build_name}/metadata.tsv.xz":    f"results/{build_name}/{build_name}_subsampled_metadata.tsv.xz",      # from `rule combine_samples`
+            f"{build_name}/aligned.fasta.xz":   f"results/{build_name}/aligned.fasta.xz",                             # from `rule build_align`
+            # export the auspice dataset which matches the subsampled sequences / metadata (see `rule finalize`)
+            f"{build_name}/{build_name}.json":                  f"auspice/{config['auspice_json_prefix']}_{build_name}.json",
```
jameshadfield, line 185: The trial-build fails at this line:

```
[33m./scripts/upload-to-s3 auspice/ncov_open_trial_2021-06-26_global.json s3://nextstrain-staging/files/ncov/open/trial/2021-06-26/global/global.json \| tee -a logs/upload.txt[0m
--
An error occurred (404) when calling the HeadObject operation: Not Found
054e9d3a96a701a3db508e0af7030e9b2c257fa2eb4d5dbf72bbf043c2c7979b auspice/ncov_open_trial_2021-06-26_global.json
0000000000000000000000000000000000000000000000000000000000000000 s3://nextstrain-staging/files/ncov/open/trial/2021-06-26/global/global.json
Uploading auspice/ncov_open_trial_2021-06-26_global.json → s3://nextstrain-staging/files/ncov/open/trial/2021-06-26/global/global.json
Unknown options:
[32m[Fri Jun 25 06:24:43 2021][0m
[31mError in rule upload:[0m
[31m    jobid: 0[0m
[31m    log: logs/upload.txt (check log file(s) for error message)[0m
[31m[0m
[31mRuleException:
CalledProcessError in line 214 of /nextstrain/build/workflow/snakemake_rules/export_for_nextstrain.smk:
Command 'set -euo pipefail;  ./scripts/upload-to-s3 auspice/ncov_open_trial_2021-06-26_global.json s3://nextstrain-staging/files/ncov/open/trial/2021-06-26/global/global.json \| tee -a logs/upload.txt' returned non-zero exit status 255.
File "/nextstrain/build/workflow/snakemake_rules/export_for_nextstrain.smk", line 214, in __rule_upload
File "/usr/local/lib/python3.7/concurrent/futures/thread.py", line 57, in run[0m
[31mExiting because a job execution failed. Look above for error message[0m
```

_This particular line may not be the bug, but this is the part of the code where the error originates_

jameshadfield: Fixed by 2556b61

PR #629, opened 2021-05-10
Auto-deploy our core builds to production instead of staging

tsibley:
Discussion in Slack¹ led to consensus to switch.  The reasoning is that
nearly all the time the build shepherd's are just copying files between
staging and production.  Any occasional bad build can be fixed by
rolling back to yesterday's build (from the dated JSON files).

¹ https://bedfordlab.slack.com/archives/C01LCTT7JNN/p1620422859124500

### DIFF
--- a/workflow/snakemake_rules/export_for_nextstrain.smk
+++ b/workflow/snakemake_rules/export_for_nextstrain.smk
@@ -157,18 +157,18 @@ except:
     # means that the Snakefile won't crash.
     deploy_origin = "by an unknown identity"
 
-rule deploy_to_staging:
+rule deploy:
     input:
         *rules.all_regions.input
     params:
-        slack_message = f"Deployed <https://nextstrain.org/staging/ncov|nextstrain.org/staging/ncov> {deploy_origin}",
-        s3_staging_url = config["s3_staging_url"]
+        slack_message = f"Deployed to {config['deploy_url']} {deploy_origin}",
+        deploy_url = config["deploy_url"]
     benchmark:
-        "benchmarks/deploy_to_staging.txt"
+        "benchmarks/deploy.txt"
     conda: config["conda_environment"]
     shell:
         """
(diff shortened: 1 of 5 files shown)


huddlej (approved):
This looks good! I like the simplicity of the new `deploy_url` parameter. After we merge this, I can update the new config reference to replace `s3_staging_url` with this new parameter.

workflow/snakemake_rules/export_for_nextstrain.smk
```diff
@@ -157,18 +157,18 @@ except:
     # means that the Snakefile won't crash.
     deploy_origin = "by an unknown identity"
 
-rule deploy_to_staging:
+rule deploy:
```
huddlej, line 160: Sorry, I typed too soon! Since this rule gets renamed, we should update [the reference to `deploy_to_staging` in the trial build config](https://github.com/nextstrain/ncov/pull/629/files#diff-3e38b7b89f28aa3eb17024c02a6ae6a641942711e90a07cbcd155935606384b9R40) and [the corresponding reference in ncov-ingest's rebuild-staging script](https://github.com/nextstrain/ncov-ingest/blob/04ca33cbed1f96320035b9f7ebcc6abf4fa25a72/bin/rebuild-staging#L58).

huddlej (changes requested):
I noticed a couple of little changes we need to make to our CLI commands before merging this.

workflow/snakemake_rules/export_for_nextstrain.smk
tsibley: Ah, thank you for that catch. Repushed. I thought I grepped both repos!

rneher (approved):

huddlej (approved):

tsibley:
Rebased and repushed to address the new configuration option docs since those have been merged to master since I forked this branch.

PR #635, opened 2021-05-13
Include lineage B.1.617 as named clade 21A

trvrb:
I've not yet tested this, but it really should just work. This is same definition used for "emerging lineages" for a little while now and it's seemed very stable.


huddlej (approved):
This worked for me!

image

@ivan-aksamentov Does nextclade need to be manually updated to reflect this new clade annotation or will this get picked up automatically by a scheduled build?

PR #636, opened 2021-05-14
Support compressed metadata, sequence indices, and early intermediate sequences

huddlej:

Working with multiple copies of full, uncompressed GISAID sequences has become unsustainable. This PR modifies early workflow outputs to use compressed sequences for aligned, masked, filtered, and combined sequence sets. This PR also changes metadata and sequence index outputs to compressed formats. Commit messages provide full details on the rationale of the implementation.

## Testing

- [x] Test by CI
- [x] Test a Nextstrain build on SLURM cluster
- [x] [Trial build on AWS](https://github.com/nextstrain/ncov/runs/2592498791?check_suite_focus=true)
- [x] [Second trial build on AWS after major bug fixes](https://github.com/nextstrain/ncov/runs/2615297965?check_suite_focus=true)

## Release checklist

### DIFF
--- a/Snakefile
+++ b/Snakefile
@@ -1,6 +1,7 @@
 import copy
 from datetime import date
 import os
+from pathlib import Path
 import sys
 from os import environ
 from socket import getfqdn
--- a/docs/change_log.md
+++ b/docs/change_log.md
@@ -5,7 +5,7 @@ We also use this change log to document new features that maintain backward comp
 
 ## New features since last version update
 
- - 17 May 2021: Compress metadata, sequence indices, and early intermediate sequences (aligned, masked, filtered, combined for subsampling, and subsampled files) to save disk space. ([#636](https://github.com/nextstrain/ncov/pull/636))
+ - 19 May 2021: Compress metadata, sequence indices, and early intermediate sequences (aligned, masked, filtered, combined for subsampling, and subsampled files) to save disk space. ([#636](https://github.com/nextstrain/ncov/pull/636))
  - 12 May 2021: Include S1 mutations and nextalign-based ancestral amino acid mutations in Auspice JSONs by default instead of requiring the now-unnecessary `use_nextalign` configuration parameter. ([#630](https://github.com/nextstrain/ncov/pull/630))
  - 12 May 2021: [Document all available workflow configuration parameters](https://nextstrain.github.io/ncov/configuration). ([#633](https://github.com/nextstrain/ncov/pull/633))
 
--- a/docs/change_log.md
+++ b/docs/change_log.md
@@ -5,6 +5,7 @@ We also use this change log to document new features that maintain backward comp
 
 ## New features since last version update
 
+ - 17 May 2021: Compress metadata, sequence indices, and early intermediate sequences (aligned, masked, filtered, combined for subsampling, and subsampled files) to save disk space. ([#636](https://github.com/nextstrain/ncov/pull/636))
  - 12 May 2021: Include S1 mutations and nextalign-based ancestral amino acid mutations in Auspice JSONs by default instead of requiring the now-unnecessary `use_nextalign` configuration parameter. ([#630](https://github.com/nextstrain/ncov/pull/630))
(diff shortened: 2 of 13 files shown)


huddlej:
Note to self: I still need to add support for the `upload` rule in the Nextstrain builds.

huddlej:
Thank you for these improvements, @rneher! The trial builds on AWS worked as expected (although it took ~11 hours to run), so we can merge this unless you are still testing, too.

huddlej:
With piped mask output to `xz -2`, run time is ~12 minutes instead of ~2 hours. With filter output to temp file followed by `xz` compression, run time is ~30 minutes instead of 2 hours.

Below is a summary of the problem with Augur's backend for managing compressed files (xopen) with a list of potential short- and long-term solutions.

[xopen uses the standard Python LZMA module to handle xz files](https://github.com/pycompression/xopen/blob/9edf101dfad2f91f7fc5d268c61dc0c49e65597c/src/xopen/__init__.py#L506-L507), but it only uses the `filename` and `mode` arguments. Python’s [LZMA library also includes support for a `preset` argument](https://docs.python.org/3/library/lzma.html#reading-and-writing-compressed-files) that controls the amount of CPU and memory used for compression. Although Augur’s `io` module passes through extra keyword arguments to the internal compression tool, this doesn’t help us because `xopen` does not support LZMA’s presets. [`xopen` only passes through its `compression` argument for gzip files](https://github.com/pycompression/xopen/blob/9edf101dfad2f91f7fc5d268c61dc0c49e65597c/src/xopen/__init__.py#L636-L637).

Ideally, we could just pass through a compression preset to the LZMA module, but in the absence of that functionality, we have to consider other options:

1. Switch from xz to gzip. This is a quick fix. It will introduce some breaking changes to workflows that depend on our hosting xz files.
2. Modify the ncov workflow to write FASTA output to stdout and compress that output with `xz -2` directly. This is a quick fix. It will be backward compatible.
3. Modify Augur to use an I/O library that parameterizes compression levels for different compression backends like LZMA. This is a long-term fix and will be backward compatible. If we used this approach, we’d probably want to introduce an environment variable that allows users to override the default Augur compression presets for different compression backends or a configuration file (a la Nextstrain CLI’s config) that supports these parameters.
	1. [fsspec](https://filesystem-spec.readthedocs.io/en/latest/api.html): to use this module, it seems we would need to [register our own LZMA compression function](https://docs.dask.org/en/latest/remote-data-services.html#compression) that hardcodes a compression preset. The default LZMA backend uses the default preset from the Python `lzma` module and there doesn’t appear to be an interface to pass kwargs through `fsspec.open` to the compression backend.
	2. [smart_open](https://github.com/RaRe-Technologies/smart_open#supported-compression-formats): as with fsspec, [this module would require us to define our own LZMA backend](https://github.com/RaRe-Technologies/smart_open#supported-compression-formats) (there isn’t one supported by default) that hardcodes the compression present we want to use.
4. Submit a patch to `xopen` that supports passing the `compression` argument from `xopen.xopen` through to the LZMA backend as the `preset`. This is a long-term fix and not guaranteed to work if the patch isn’t accepted.

ivan-aksamentov:
xopen is a single 600-line-ish python file, and we probably use like 10 of these lines. We could just copy-paste it into augur and modify to our needs.

How versatile the compression really needs to be? Do we really need a library and pay the associated abstraction costs? We could stick to one format/preset and hardcode it forever. Or, if yet another older format needs to be supported, then hardcode it too. Branch depending on filename or a flag. Can be done without a library. Thoughts?

huddlej:
Those are good points, @ivan-aksamentov, and worth discussing as a group.

It is _probably_ not too much work to maintain a few more lines of code that use Python's standard compression modules to handle compression ourselves. We have discussed supporting alternate file transport options, though, like S3, HTTP, etc. That type of functionality often gets bundled in these same libraries that support compression transparently, so it's worth figuring out the value of that functionality before we commit to rolling our own compression code.

PR #639, opened 2021-05-19
Handle multiple file inputs in the same precedence order.

ttung:

Currently, the metadata is merged preferring the LAST value seen, and the sequences are merged preferring the FIRST value seen.  This change makes it so that it's always the last value seen.

## Testing

We've been using this internally for our builds.

### DIFF
--- a/scripts/combine-and-dedup-fastas.py
+++ b/scripts/combine-and-dedup-fastas.py
@@ -24,8 +24,10 @@ if __name__ == '__main__':
     with open_file(args.output, "w") as output_handle:
         # Stream sequences from all input files into a single output file,
         # skipping duplicate records (same strain and sequence) and noting
-        # mismatched sequences for the same strain name.
-        for record in read_sequences(*args.input):
+        # mismatched sequences for the same strain name.  In order to
+        # prefer the latter files, we have to reverse the order of the
+        # files.
+        for record in read_sequences(*reversed(args.input)):
             counter += 1
             if counter % 10000 == 0:
                 print(f"Processed {counter} records")
--- a/scripts/sanitize_sequences.py
+++ b/scripts/sanitize_sequences.py
@@ -18,6 +18,8 @@ if __name__ == '__main__':
         pattern = ""
 
     with open_file(args.output, "w") as output_handle:
-        for sequence in read_sequences(*args.sequences):
+        # In order to prefer the latter files, we have to reverse the order of
+        # the files.
+        for sequence in read_sequences(*reversed(args.sequences)):
             sequence.id = re.sub(pattern, "", sequence.id)
             write_sequences(sequence, output_handle)


scripts/combine-and-dedup-fastas.py
```diff
@@ -24,8 +24,10 @@
     with open_file(args.output, "w") as output_handle:
         # Stream sequences from all input files into a single output file,
         # skipping duplicate records (same strain and sequence) and noting
-        # mismatched sequences for the same strain name.
-        for record in read_sequences(*args.input):
+        # mismatched sequences for the same strain name.  In order to
+        # prefer the latter files, we have to reverse the order of the
+        # files.
+        for record in read_sequences(reversed(*args.input)):
```
huddlej, line 30: I think you want the star expansion operator to precede the `reversed` call like `read_sequences(*reversed(args.input))`. The current syntax tries to expand a list into `reversed` which doesn't expect multiple arguments and throws a runtime error.

```bash
$ python3 scripts/combine-and-dedup-fastas.py  --input data/example_sequences_aus.fasta.xz data/example_sequences_worldwide.fasta.xz --warn-about-duplicates --output test.fasta
  File "/Users/jlhudd/miniconda3/envs/nextstrain/lib/python3.8/site-packages/augur/io.py", line 33, in open_file
    yield handle
  File "scripts/combine-and-dedup-fastas.py", line 30, in <module>
    for record in read_sequences(reversed(*args.input)):
TypeError: reversed expected 1 argument, got 2
```

huddlej (changes requested):
Great catch, @ttung! We definitely need this change. Since it will be a backwards incompatible change (in that it will change the behavior of people's current workflows, even though they've been relying on a bug :-/), we will plan to make a new release after we merged this.

I had one syntax comment below, but after that's fixed we can merge this.

scripts/combine-and-dedup-fastas.py
ttung: I was going to say, "i swear this is working in our setup."

And then I realized I did not port [our change out](https://github.com/chanzuckerberg/aspen/blob/trunk/src/backend/aspen/workflows/nextstrain_run/combine-and-dedup-fastas.py.patch) correctly.  #facepalm

huddlej (approved):
Works for me! I'll merge, update the change log, and make a new release tomorrow morning. Thanks again, @ttung!

ttung:
Thanks, @huddlej!

ttung:
I noticed that `scripts/sanitize_sequences.py` probably needs the same fix.

fanninpm:
> Also, you may be interested in the latest release of the workflow that supports reading GISAID data directly from tarballs based on your earlier feedback on a related PR. I'd be interested if this ends up working for your needs or if there's anything that could be improved there.

@huddlej Thanks for letting me know! I tried that with the raw GISAID .tar.xz files (the sequence .tar.xz archive is about an order of magnitude smaller than the curated .fasta.gz), but I ran into an `Exception: Where there’s SAMPLING_TRAIT we should always have EXPOSURE_TRAIT` later on down the line (which was found in our private metadata). I put `skip_travel_history_adjustment: True` in the builds.yaml file and am re-running the workflow.

huddlej:
Ok, that is good to know that this was an issue. We're [planning to remove this part of the workflow soon](https://github.com/nextstrain/ncov/issues/646), so this issue is helpful motivation/context for prioritizing that change.

PR #640, opened 2021-05-19
Support default GISAID metadata and sequences

huddlej:

We would like to support GISAID's standard metadata and sequence data formats from the "Download packages" interface of EpiCoV. This PR expands the existing "sanitize" scripts for metadata and sequences with the following major changes to support these data:

 - [x] Remove whitespace from strain names in sequence and metadata (whitespace is not allowed in record ids for FASTA deflines)
 - [x] Strip metadata from FASTA deflines (these metadata entries should be present already in the corresponding separate metadata file)
 - [x] Rename GISAID metadata columns to Augur-style names ("Virus name" -> "strain" and "Collection date" -> "date")
 - [x] Parse `region`, `country`, `division`, and `location` values from the single `Location` field in the GISAID metadata
 - [x] Resolve duplicate metadata records by preferring those with the latest GISAID accession number (with an option to produce an error with a list of all duplicates instead)
 - [x] Resolve duplicate sequence records by preferring the first sequence encountered (necessary for downstream components of the workflow when the combine-and-dedup step doesn't get run and duplicates exist in the sequence data)
 - [x] Ingest sequences (`.fasta`) and metadata (`.tsv`) files directly from GISAID tarballs (allows users to download data and run the workflow directly on those files without manually extracting/decompressing sequences and metadata)

Currently, these changes are applied to all inputs including data that are not in the GISAID default format.

## Testing

 - [x] Manually test sanitizer scripts locally with the full GISAID downloads
 - [x] Run a small build with a 200 random strains from the full downloads
 - [x] Tested with a full Nextstrain build on a SLURM cluster
 - [x] [Tested with AWS Batch trial build](https://github.com/nextstrain/ncov/runs/2677037937?check_suite_focus=true)

## Release checklist

 - [x] Determine the version number for the new release by incrementing [the most recent release](https://github.com/nextstrain/ncov/releases)  -> `v7`


emmahodcroft:
Apologies if this is a naive question, is this also stripping out `hCoV/` or whatever the header is - so that the root sequence & exclude files will work properly?
Apologies if this is in there and I just missed it while skimming changes 🙃

huddlej:
@emmahodcroft Yeah, we actually do that as part of the current sanitize scripts based on [these config parameters](https://github.com/nextstrain/ncov/blob/84d586ca12720e249de0e1f022f8d97c7ef184e7/defaults/parameters.yaml#L9-L11). We had to make that change to get [the "Augur input" format downloads to work with our standard include/exclude files](https://github.com/nextstrain/ncov/pull/605) and then this PR builds on those existing scripts.

emmahodcroft:
Thanks John! I had thought I'd seen this somewhere before, but I just couldn't remember exactly where. Apologies for the repeat!

huddlej:
This PR implements duplicate resolution for metadata using [a procedure described as a possible improvement for Augur's own metadata reader](https://github.com/nextstrain/augur/issues/725). If this approach to resolving duplicates seems reasonable, we can port this function into Augur.

This PR _does not_ implement duplicate resolution for sequences (at least, not yet!).

huddlej:
After some further modifications to the workflow, we can now run builds like the following that use GISAID tarballs directly as metadata and sequences inputs:

```yaml
# Define inputs with preferred sequences/metadata listed last.
inputs:
  - name: north-america
    metadata: data/ncov_north-america.tar.gz
    sequences: data/ncov_north-america.tar.gz
  - name: washington
    metadata: data/gisaid_auspice_input_hcov-19_2021_05_24_21.tar
    sequences: data/gisaid_auspice_input_hcov-19_2021_05_24_21.tar

# Define builds.
builds:
  washington:
    region: North America
    country: USA
    division: Washington
    subsampling_scheme: focal-contextual

# Define subsampling scheme.
subsampling:
  focal-contextual:
    focal:
      query: --query "division == '{division}'"
      max_sequences: 20
    contextual:
      query: --query "division != '{division}'"
      max_sequences: 20
      group_by: region year month
      priorities:
        type: proximity
        focus: focal
```

Internally, the `sanitize_sequences.py` script extracts the first `.tsv` or `.fasta` file from a given (compressed or uncompressed) tarball, decompresses that file as needed, and returns the corresponding buffer to be consumed downstream.

scripts/sanitize_metadata.py
```diff
@@ -1,32 +1,282 @@
...
+    # fields to be incomplete for some databases. By sorting across all fields,
+    # we use information from all available accession fields. If all fields
+    # contain missing values (e.g., "?"), we end up returning the last record
+    # for a given strain as a reasonable default.
+    sort_fields = [strain_field]
+    if len(accession_fields) > 0:
+        sort_fields.extend(accession_fields)
+
+    # Return the last record from each group after sorting.
+    return metadata.sort_values(sort_fields).groupby(strain_field, sort=False).last().reset_index()
```
rneher, line 147: Maybe this is a more transparent way of doing this:
https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.drop_duplicates.html

rneher (approved):
This looks good to me. Thanks John. I left one comment about the `pandas.drop_duplicates`

scripts/sanitize_metadata.py
huddlej: This implementation is a little more complicated because it tries to resolve duplicates by preferring records with the most recent accession from GISAID or GenBank instead of taking the first/last record by default. This prefers the latest metadata record, although the bigger issue is preferring the latest sequence record and we can't do that unless GISAID includes the record's accession with the sequence header.

I just expanded the inline comments around this code, though, to clarify the context of why it's sorting and then grouping the way it is.

huddlej: I didn't notice the `keep` argument for `drop_duplicates` before. That does make the code clearer and a little faster, too. Thanks, @rneher!

huddlej:
[Based on conversation in Slack](https://bedfordlab.slack.com/archives/CSKMU6YUC/p1622129937051300), I opted to not standardize column names in the metadata in favor of the simpler approach of manually renaming all columns.

PR #641, opened 2021-05-20
Run xz outside python.

ttung:

xopen+lzma has performance issues as it does not support passing compression levels to py-lzma.  This moves the compression out of the control of the python process.

## Testing

will kick off a nextstrain run.


huddlej (approved):
This works for me. Thanks for catching this bottleneck in the workflow, @ttung!

PR #536, opened 2021-01-22
Allow customisable auspice JSON prefixes

ignasiialemany:
### Description of proposed changes    
Pull request fix #488 . Add `prefix` to json files inside `auspice` folder. The `prefix` can be specified in `defaults/parameters.yml`.  If it is not specified in `parameters.yml` the code considers the default prefix "ncov" for backward compatibility .

PR #643, opened 2021-05-25
Support custom Auspice JSON prefixes

### LINKED ISSUE #488: Provide a configuration parameter for auspice JSON prefixes
**Context**

[Community builds sourced through nextstrain.org/community](https://nextstrain.org/docs/contributing/community-builds) currently require the prefix of each auspice JSON to match the name of the repository where they are hosted (e.g., `zika-columbia_tree.json` in the [zika-columbia repo](https://github.com/blab/zika-colombia/tree/master/auspice), `covidtracker_ca.json` in the [Biohub California build repo](https://github.com/czbiohub/covidtracker/tree/master/auspice)).

However, the auspice JSONs produced by the ncov workflows [all get prefixed by `ncov_`](https://github.com/nextstrain/ncov/blob/26a77de9594f7b1538a07d0c6c7532f5995c3224/Snakefile#L31). Users who want to provide their builds through the community interface need to manually rename their auspice JSONs or add a custom rule to the workflow to automate this.

**Description**

As an alternative to a custom renaming rule, we could provide a configuration parameter like `auspice_json_prefix` as a top-level setting. This setting could default to `ncov` for backwards-compatibility (defined in `defaults/parameters.yaml`), but users could also override the setting in their custom profiles (e.g., `auspice_json_prefix: zika-columbia`). This setting would allow users to automatically produce JSONs that could be hosted from their GitHub repositories.

huddlej:

Adds support for custom Auspice JSON prefixes other than the default `ncov`. This is a fork of @ignasiialemany's original PR #536 that includes some updates for the latest version of the workflow including documentation in the configuration reference (and support for authorized Docker pulls in CI, etc.).

## Related issue(s)

Fixes #488
Related to #536

## Testing

 - [x] Tested by CI

## Release checklist

### DIFF
--- a/Snakefile
+++ b/Snakefile
@@ -122,8 +122,8 @@ localrules: download_metadata, download_sequences, clean
 # Create a standard ncov build for auspice, by default.
 rule all:
     input:
-        auspice_json = expand("auspice/ncov_{build_name}.json", build_name=BUILD_NAMES),
-        tip_frequency_json = expand("auspice/ncov_{build_name}_tip-frequencies.json", build_name=BUILD_NAMES)
+        auspice_json = expand("auspice/{ext}_{build_name}.json", build_name=BUILD_NAMES,ext=config["auspice_json_prefix"]),
+        tip_frequency_json = expand("auspice/{ext}_{build_name}_tip-frequencies.json", build_name=BUILD_NAMES,ext=config["auspice_json_prefix"])
 
 rule clean:
     message: "Removing directories: {params}"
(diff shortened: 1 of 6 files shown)


tsibley (approved):
The final diff looks great, and this config knob will help address a need in #628. :raised_hands:

I would gently suggest that the 5 commits which make up this PR be rewritten into one commit as together they make one conceptual change and none of the commits really stand alone. One way I think about this is by asking myself if the repo/code would be in an inconsistent state at any point between commits on a branch. The principle is that the repo/code should be coherent at any given commit. There are times where it's appropriate to violate this principle, but IMO it's not often warranted.

jameshadfield (approved):
Let's get this merged!

PR #656, opened 2021-06-08
Use subsampled metadata for per-build outputs

### LINKED ISSUE #654: Speed up workflow by using subsampled metadata
Right now rules like `colors` will result in a call that looks like:

```
python3 scripts/assign-colors.py
  --ordering defaults/color_ordering.tsv
  --color-schemes defaults/color_schemes.tsv
  --output results/global/colors.tsv
  --metadata results/sanitized_metadata_gisaid.tsv.xz
```

where the entire ~1.7M line metadata file is passed into the script. This causes this rule to take 8 minutes when running locally. However, this rule could be called just as well as:

```
python3 scripts/assign-colors.py
  --ordering defaults/color_ordering.tsv
  --color-schemes defaults/color_schemes.tsv
  --output results/global/colors.tsv
  --metadata results/global/global_subsampled_metadata.tsv.xz
```

where just the ~4k line metadata file is passed in. This runs in 1.6 seconds (!) instead.

We have the same story with the `export` rule that currently is called as:

```
augur export v2
 --metadata results/sanitized_metadata_gisaid.tsv.xz
...
```

and takes 55 seconds. It could instead be called as:

```
augur export v2
 --metadata results/global/global_subsampled_metadata.tsv.xz
...
```

and take 9 seconds.

Similarly, correcting the `recency` rule goes from 66 seconds to <1 second.

The issue is that `_get_metadata_by_wildcards` is not actually returning the build-specific subsampled metadata. Instead it's defaulting to the broader `results/sanitized_metadata_gisaid.tsv.xz`. 

Besides `colors` and `export` there are a number of rules that rely on metadata. I'm assuming that correcting the input will speed these up as well. But even just fixing `colors`, `recency` and `export` would save ~60 minutes (!) of single-core compute in the daily rebuilds.

I'm not immediately sure how to correct this. @huddlej, would you be able to take a look? I'm assuming this bit of the workflow is something you're quite familiar with.

huddlej:

Replaces calls to unified metadata (or metadata-by-wildcards) with an explicit path to the subsampled metadata from the final subsampling filter step. The subsampled metadata should be substantially smaller than the full unified metadata, so the rules modified in this commit should all run correspondingly faster.

The use of the full metadata was a holdover from earlier versions of the workflow that did not produce a subsampled metadata. The use of the full metadata in the rules modified here was an oversight of mine when I added the subsampled metadata output.

## Related issue(s)

Fixes #654

## Testing

 - [x] CI build
 - [x] Nextstrain builds on SLURM cluster ("europe", "global", and "africa")
 - [x] [Trial Nextstrain builds on AWS](https://github.com/nextstrain/ncov/runs/2787879453?check_suite_focus=true)

## Release checklist

This is a bug fix.

### DIFF
--- a/workflow/snakemake_rules/export_for_nextstrain.smk
+++ b/workflow/snakemake_rules/export_for_nextstrain.smk
@@ -46,7 +46,7 @@ rule clean_export_regions:
 rule extract_meta:
     input:
         alignment = rules.build_align.output.alignment,
-        metadata = _get_metadata_by_wildcards
+        metadata="results/{build_name}/metadata_adjusted.tsv.xz",
     output:
         metadata = "results/{build_name}/extracted_metadata.tsv"
     run:
@@ -65,8 +65,7 @@ rule export_all_regions:
     input:
         auspice_json = expand("results/{build_name}/ncov_with_accessions.json", build_name=BUILD_NAMES),
         lat_longs = config["files"]["lat_longs"],
-        metadata = [_get_metadata_by_build_name(build_name).format(build_name=build_name)
-                    for build_name in BUILD_NAMES],
+        metadata=expand("results/{build_name}/metadata_adjusted.tsv.xz", build_name=BUILD_NAMES),
         colors = expand("results/{build_name}/colors.tsv", build_name=BUILD_NAMES),
     benchmark:
         "benchmarks/export_all_regions.txt"
--- a/workflow/snakemake_rules/main_workflow.smk
+++ b/workflow/snakemake_rules/main_workflow.smk
@@ -637,7 +637,7 @@ rule adjust_metadata_regions:
         Adjusting metadata for build '{wildcards.build_name}'
         """
     input:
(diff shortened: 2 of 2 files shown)


rneher:
this looks good to me, but why did you use the pattern `{build_name}/{build_name}_metadata...` here:

https://github.com/nextstrain/ncov/blob/f7af2f271ac519c4e4f29aebd6d4fc452f108930/workflow/snakemake_rules/main_workflow.smk#L524-L526

huddlej:
Thank you for checking this out, @rneher! That pattern is from #592 where we started uploading the subsampled sequences and metadata to S3 as a starting point for other folks in the lab to use for their builds. The pattern produces files like `global_subsampled_sequences.fasta` and `europe_subsampled_sequences.fasta` that can be distinguished in a flat S3 bucket.

We could probably upload these to S3 paths like `global/subsampled_sequences.fasta` though and make the filenames simpler.

PR #659, opened 2021-06-08
Dynamically calculate min_date for frequencies

trvrb:

By default, use 1 year before present for `min_date` value for frequencies estimates. This serves to emphasize more recent dynamics. If desired, some other `min_date` can be explicitly set in the build config.

## Testing

Testing has been performed locally. Resulting build output can be seen at: https://nextstrain.org/staging/restrict-frequencies/ncov/global

## Release checklist

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -227,7 +227,7 @@ Builds support any named attributes that can be referenced by subsampling scheme
 ### min_date
 * type: float or string
 * description: Earliest date to estimate frequencies for. Dates can be numeric floating point values (e.g., `2019.74`) or ISO 8601-style strings (e.g., `2019-10-01`).
-* default: `2020.0`
+* default: without value supplied, defaults to 1 year before present
 
 ### pivot_interval
 * type: integer
--- a/defaults/parameters.yaml
+++ b/defaults/parameters.yaml
@@ -115,7 +115,9 @@ ancestral:
 
 # Frequencies settings
 frequencies:
-  min_date: 2020.0
+
+  # min_date is set by default to 1 year before present
+  # but can be explicitly set if desired
 
   # Number of months between pivots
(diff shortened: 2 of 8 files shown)


docs/configuration.md
```diff
@@ -227,7 +227,7 @@ Builds support any named attributes that can be referenced by subsampling scheme
 ### min_date
 * type: float or string
 * description: Earliest date to estimate frequencies for. Dates can be numeric floating point values (e.g., `2019.74`) or ISO 8601-style strings (e.g., `2019-10-01`).
-* default: `2020.0`
+* default: with value supplied, defaults to 1 year before present
```
jameshadfield, line 230: Do you mean _without_ value supplied?

trvrb: Yes! I'll fix.

trvrb: Done in dedf9456eb442a7babb9c9640cc034ca226adfc5 and force pushed to branch.

rneher:
looks good to me.

huddlej (approved):
When I ran this with the getting started data, I initially got a frequencies panel like this:

image

Given that the getting started data only extends to mid-May 2020, this result made sense to me, but I could see it being surprising to users who run any analysis on data from the past that doesn't extend to today's date. I wonder if a better long-term solution would be for `augur frequencies` to support date offsets for the min/max date in [the same way we've proposed offsets for `augur filter`](https://github.com/nextstrain/augur/issues/721). Then we could explicitly store the min date in the config as:

```yaml
frequencies:
  min_date: 1Y # Or use negative offset values, if that's clearer like "-1Y".
```

This default would work more consistently for any dataset.

For the short-term, I've added reasonable min/max dates to the frequencies config for the getting started profile in this PR.

PR #676, opened 2021-06-30
fix: clade definition of 21E is too restrictive

rneher:
the definition of clade 21E often only included a single sequence because of a mutation with patchy distribution. This PR deletes that position from the definition.

### DIFF
--- a/defaults/clades.tsv
+++ b/defaults/clades.tsv
@@ -118,3 +118,4 @@ clade	gene	site	alt
 21H	nuc	17491	T
 21H	nuc	21995	A
 21H	nuc	21993	C
+21H	nuc	22599	A
--- a/defaults/clades.tsv
+++ b/defaults/clades.tsv
@@ -98,7 +98,6 @@ clade	gene	site	alt
 21D (Eta)	nuc	24748	T
 
 21E (Theta)	nuc	12049	T
-21E (Theta)	nuc	22356	G
 21E (Theta)	nuc	23341	C
 21E (Theta)	nuc	23604	A
 21E (Theta)	nuc	24187	A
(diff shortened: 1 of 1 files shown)


emmahodcroft:
I would also suggest we move back `21H` one node - it doesn't seem to me there's any reason to draw the line where the `19035 C` mutation is? I'd suggest replacing this with `22599 A` :

https://nextstrain.org/ncov/south-america?c=gt-nuc_22599
image

rneher:
It actually moves it back two branches as ties it to the 144/145 mutation which seems pretty distinctive.

emmahodcroft:
Great, thanks! I'll use this to generate defining mutations for CoVariants

trvrb:
I was attempting to place 21H to correspond directly to PANGO B.1.621. However, it didn't quite take as I wanted it too. However, I would just move back one branch rather than two to keep correspondence with PANGO:

21H-pango

Although it's of course possible that PANGO definition of B.1.621 would get updated as well.

emmahodcroft:
Notably, that one sequence that sits inside/outside is not great quality (2116 N's and only 9 mutations for a sequence dated April 2021) so that break may be slightly fictitious anyway...

rneher:
spike positions 95 and 950 are quite homoplasic. 346 is ok, but also shows up sometimes. Y144S/Y145N is not seen anywhere else. My hope was that this is a pretty robust marker and it labels the same node in most trees.

rneher:
We can move the definition to the 95/346 branch for example by adding one of the, Currently, the label would site here the color changes here:
image

which included the dubious sequence that not B.1.621. I like including 144/145 since I have seen this positions deleted but not mutated much, so I hope this might be a robust label.

PR #677, opened 2021-07-02
Feat/diagnostics

rneher:
allow more flexible (and restrictive) filtering in the diagnostic script.


emmahodcroft:
Thanks Richard, this should allow us to put in place more strict QC for older sequences. I'd like to touch on how best to get that implemented at the meeting tomorrow.

PR #683, opened 2021-07-07
Infer compression for downloaded sequence data

jameshadfield:
Downloading of sequence data ("sequences", "masked",
"aligned" and "filtered" remote inputs) used an inconsistent approach
which caused errors in certain situations.

(1) "sequences" were downloaded by `rule download_sequences` with no
attempt to decompress the data, but the resulting output always ended in
`.gz`. This file was read by `scripts/sanitize_sequences.py` which
uses `xopen` [1] to read the file. Xopen uses the file suffix to infer
the compression algorithm, and thus failed when the file used xz
compression [2], or no compression at all, with errors similar to
`OSError: Not a gzipped file (b'\xfd7')`.

Here we change the rule to save downloaded "sequences" inputs with only
a `.fasta` suffix, regardless of whether compression is used. This
forces xopen to infer the compression used from the file contents, thus
allowing various compression algorithms to be used (as well as
uncompressed remote inputs). Note that our open (GenBank) sequences
input (via `ncov-ingest`) is XZ compressed.

(2) When "masked", "aligned" and "filtered" remote inputs were
specified, these were decompressed during download, and saved with a
".fasta" suffix. All the scripts / augur commands which read this data
use xopen internally, and so we can avoid decompressing these files.
This results in large space savings: currently the open (GenBank)
"aligned" input is 35MB xz compressed and 24GB decompressed! We still
decompress remote metadata inputs as these are read by
`scripts/sanitize_metadata.py` which uses `pd.read_csv` which does not
infer compression from file contents.

The tests have been expanded to cover uncompressed inputs, and the
documentation has been updated accordingly.

[1] https://github.com/pycompression/xopen

[2] Note that this error did not occur on MacOS (where xopen would read
a `.gz` file which used xz compression). The error was observed in our 
AWS jobs (Linux).

## Testing

Successful AWS run using `nextstrain-open` profile. 

Tests pass for me (MacOS). It would be good for a linux user to run them as well (`cram --preserve-env tests/different-inputs.t`)

## Release checklist

### DIFF
--- a/workflow/snakemake_rules/download.smk
+++ b/workflow/snakemake_rules/download.smk
@@ -33,7 +33,7 @@ def _infer_decompression(input):
 rule download_sequences:
     message: "Downloading sequences from {params.address} -> {output.sequences}"
     output:
-        sequences = "data/downloaded_{origin}.fasta.gz"
+        sequences = "data/downloaded_{origin}.fasta"
     conda: config["conda_environment"]
     params:
         address = lambda w: config["inputs"][w.origin]["sequences"],
@@ -62,10 +62,9 @@ rule download_aligned:
     conda: config["conda_environment"]
     params:
         address = lambda w: config["inputs"][w.origin]["aligned"],
-        deflate = lambda w: _infer_decompression(config["inputs"][w.origin]["aligned"])
     shell:
         """
-        aws s3 cp {params.address} - | {params.deflate} > {output.sequences:q}
+        aws s3 cp {params.address} {output.sequences:q}
         """
 
 
@@ -76,10 +75,9 @@ rule download_masked:
     conda: config["conda_environment"]
     params:
         address = lambda w: config["inputs"][w.origin]["masked"],
-        deflate = lambda w: _infer_decompression(config["inputs"][w.origin]["masked"])
     shell:
         """
-        aws s3 cp {params.address} - | {params.deflate} > {output.sequences:q}
+        aws s3 cp {params.address} {output.sequences:q}
         """
 
 
@@ -90,8 +88,7 @@ rule download_filtered:
     conda: config["conda_environment"]
     params:
         address = lambda w: config["inputs"][w.origin]["filtered"],
-        deflate = lambda w: _infer_decompression(config["inputs"][w.origin]["filtered"])
     shell:
         """
-        aws s3 cp {params.address} - | {params.deflate} > {output.sequences:q}
+        aws s3 cp {params.address} {output.sequences:q}
         """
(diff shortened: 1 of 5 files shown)


workflow/snakemake_rules/download.smk
```diff
@@ -33,7 +33,7 @@ def _infer_decompression(input):
 rule download_sequences:
     message: "Downloading sequences from {params.address} -> {output.sequences}"
     output:
-        sequences = "data/downloaded_{origin}.fasta.gz"
+        sequences = "data/downloaded_{origin}.fasta"
```
rneher, line 36: I don't understand this... we are downloading an `xz` file but save it without suffix?

jameshadfield: Yes. `xopen` will then read the contents to infer the compression. This allows us to support `gz` and uncompressed FASTA inputs for sequences (as well as anything else `xopen` can read). 

I don't know how controversial this approach is, but there is precedent (e.g. the Auspice JSONs on `nextstrain-data` are `gz` compressed, but the files end in `.json`).

emmahodcroft: It is a shame if there's no way to flag this. I can 100% see this throwing me for an absolute loop - I wouldn't expect a file to be compressed without an ending, so I'd assume it was getting corrupted somehow....

jameshadfield: There are alternative approaches (e.g. #565). 

Currently the master branch is working (via https://github.com/nextstrain/ncov/commit/750112e2aa2dabbcb74a2a45933459f80abcda23) so there is no longer urgency for this PR.

emmahodcroft: I definitely like how neat this - it's a nice solution. My only hesitation is that I can completely see myself getting stumped by this in about a month (the absolute limit of my memory for all intents and purposes) - and I'm lucky as I could post on slack and have one of you guys remind me 😅 , but others might also get caught up in this (for the open files). However, I'm happy to defer to what works and what's organized.

rneher: I can't say I love it... the jsons on `nextstrain-data` have the appropriate content encoding so the browsers handle this correctly. 

But with our 'one workflow does everything' approach, this either requires even more plumbing or automagic solutions like this.

huddlej: I'm also not fully on-board with this approach. The precedent of compressed JSONs without standard compressed file extensions has often caused me problems and I can see this similarly confusing us in the future.

Even more generally, I'm worried about depending on `xopen`-specific functionality when we're planning to move away from `xopen` eventually. If we do go this route, we'll need to make sure we implement content sniffing in our own compression handlers, etc.

Even though it's a pain to update extensions throughout the Snakefile, I like @rneher's fix in 750112e.

PR #690, opened 2021-07-22
Set global recursion limit for shell commands

### LINKED ISSUE #689: Large trees produce "maximum recursion limit" errors at multiple steps of the workflow
**Current Behavior**
When users try to build a tree with more than our standard number of sequences (>~4000-5000), the workflow can crash at several different points with a "maximum recursion limit" error. This error is caused by BioPython's implementation of tree traversals and Python's default recursion limit. We work around the issue by allowing users to set a higher maximum recursion limit with the environment variable `AUGUR_RECURSION_LIMIT`.

Currently, we only set this variable in the export rule or in one of our custom profiles.

**Expected behavior**
We should always set the recursion limit to a reasonably high number, so users never have to encounter this error.

**Possible solution**

Use [Snakemake's `shell.prefix` function](https://snakemake.readthedocs.io/en/stable/project_info/faq.html#i-want-to-configure-the-behavior-of-my-shell-for-all-rules-how-can-that-be-achieved-with-snakemake) to set the maximum recursion limit to a reasonably high number at the beginning of the workflow (in the Snakefile) and remove all other references to this variable from the workflow. Specifically the solution looks like:

```python
shell.prefix("export AUGUR_RECURSION_LIMIT=10000; ")
```

**Additional context**

Thank you to @StephLunn for reporting this issue!

huddlej:

Instead of setting the recursion limit for individual rules as issues with maximum recursion occur, we set the recursion limit as an environment variable prefixed to all shell commands in the workflow. This approach allows us to control the limit with a single variable and affect the entire workflow.

## Related issue(s)

Fixes #689

## Testing

 - [x] Tested by CI

## Release checklist

This is a minor change.

### DIFF
--- a/Snakefile
+++ b/Snakefile
@@ -12,6 +12,10 @@ from collections import OrderedDict
 import textwrap
 import time
 
+# Set the maximum recursion limit globally for all shell commands, to avoid
+# issues with large trees crashing the workflow.
+shell.prefix("export AUGUR_RECURSION_LIMIT=10000; ")
+
 # Store the user's configuration prior to loading defaults, so we can check for
 # reused subsampling scheme names in the user's config. We need to make a deep
 # copy because Snakemake will deep merge the subsampling dictionary later,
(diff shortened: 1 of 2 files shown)


jameshadfield (approved):
Tested using augur 12.0.0

PR #691, opened 2021-07-26
Make submission date optional

jameshadfield:
Metadata typically includes the 'submission_date' field, which was used
in two rules: `diagnostic` and `recency`. The latter rule proceeded
without error if the metadata field was missing (albeit producing
a non-useful output) but the former required the field to exist. [This
has prevented people from running the pipeline](https://discussion.nextstrain.org/t/error-in-main-workflow-smk-file-in-latency-wait/616).

This commit allows the diagnostic step to proceed without the
'submission_date' field, printing a warning message instead.


huddlej (approved):
I just experienced this issue, too! This looks good to me.

PR #696, opened 2021-07-30
Add new tweet generation script

MoiraZuber:

With the new monthly tweeting style, the new tweeting script now needs to:
* Work on both `gisaid` and `open` data separately
* Take no diff files as input. Instead, the script traverses the entire metadata file and collects submitting, originating lab and authors from all sequences submitted within a specified month
* Search the excel spread sheet for matching twitter handles, and flag unknown labs for manual handle search

Additionally, if requested by the user:
* Traverse the entire metadata file to collect _all_  submitting, originating lab and authors from _before_ the designated month
* Search the excel spread sheet for matching twitter handles (ignoring unknown labs for simplicity reasons)
* Compare the list of all-time-labs with the list of new labs and produce a tweet on all newly appeared twitter handles
* Detect new countries that have not submitted before this month and include in the tweet

The script uses the following command line inputs:
`-data` specify `gisaid` or `open` (default: `gisaid`)
`-date` specify a month (e.g. 2021-07) (default: current month)
`-tweet` By adding this flag, the script will not only collect labs and handles for the past month, but for all time, and compare the two sets to produce a tweet

## Release checklist

The script should be usable for the upcoming tweet (covering new sequences from July) for `gisaid` sequences. The following points might need further discussion:

 - [ ] Solve `open` lab issues: Currently, only authors are provided in genbank metadata, which further complicates twitter handle search.
 - [ ] Whether to use originating lab and authors. For simplicity reason, the script currently only uses submitting lab for the handle search, which might cause originating labs in small countries not to be acknowledged if submitted by a larger, international lab.
 - [ ] Is more flexible date designation required? Currently, the script is designed to always cover an entire "calendar month" of sequences. Due to only needing to compare strings now instead of datetime variables, this creates a massive speed increase compared to the old script (which needed to filter for sequences younger than 30 days).
- [ ] Decide on final tweet template: The current formulations are only a draft and can be changed if a different style is desired
- [ ] Remove day specification in line 79 (which is only needed because half of July was already tweeted)
- [ ] Currently, new countries are treated separately from the twitter handles. So a new country could be mentioned in the tweet, while the lab is not tagged if it has been tagged previously (e.g. with internationally acting labs like `@CVFCarrington & @UWI_StAugustine`)
- [ ] Several small adjustments needed for `open` tweets (links, tagging of Genbank?)

### DIFF
--- a/scripts/developer_scripts/tweet_generation.py
+++ b/scripts/developer_scripts/tweet_generation.py
@@ -1,0 +1,314 @@
+import sys
+import datetime
+import pandas as pd
+
+def bold(s):
+    return('\033[1m' + s + '\033[0m')
+
+
+# 0. Read excel sheet, read stored list of new labs ("new labs" list)
+def read_excel_lab_file(table_file_name):
+    excel_table = pd.read_excel(table_file_name, index_col=0, skiprows=1)
+    excel_table = excel_table.fillna("empty?")
+    lab_dictionary = {}
+    for country, row in excel_table.iterrows():
+        description = row["Who"]
+        handle = row["Who to tag"]
+        if country not in lab_dictionary:
+            lab_dictionary[country] = {}
+        if description in lab_dictionary[country]:
+            print("Warning: lab description is found two times in excel table in same country (" + str(country) + ", " + str(description) + ")")
+        lab_dictionary[country][description.lower()] = handle
+    return lab_dictionary
+
+### Optional: only if requested
+# 1. Read metadata and collect submitting lab, originating lab and authors of all sequences within the specified month
+#   - Also make a collection of all countries all time in order to detect newly appeared countries this month
+#   - Return all labs of specified month & new countries
+def read_metadata(filename, date_g, tweet):
+
+    year_g = date_g[:4]
+    month_g = date_g[5:7]
+
+    if month_g == "12":
+        year_gplus = str(int(year_g) + 1)
+        month_gplus = "01"
+    else:
+        year_gplus = year_g
+        month_gplus = str(int(month_g) + 1)
+        if len(month_gplus) == 1:
+            month_gplus = "0" + month_gplus
+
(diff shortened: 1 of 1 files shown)


eharkins:
> Solve open lab issues: Currently, only authors are provided in genbank metadata, which further complicates twitter handle search.

Seems like a separate issue is needed for this, it may already exist: https://github.com/nextstrain/ncov-ingest/issues/133 ?

> Whether to use originating lab and authors. For simplicity reason, the script currently only uses submitting lab for the handle search, which might cause originating labs in small countries not to be acknowledged if submitted by a larger, international lab.

Could we include originating labs if we have handles for them in the spreadsheet like we were doing previously?

> Is more flexible date designation required? Currently, the script is designed to always cover an entire "calendar month" of sequences. Due to only needing to compare strings now instead of datetime variables, this creates a massive speed increase compared to the old script (which needed to filter for sequences younger than 30 days).

>  Decide on final tweet template: The current formulations are only a draft and can be changed if a different style is desired

Both seem ok for now, we can adjust as needed.


> Remove day specification in line 79 (which is only needed because half of July was already tweeted)

Will do.

> Currently, new countries are treated separately from the twitter handles. So a new country could be mentioned in the tweet, while the lab is not tagged if it has been tagged previously (e.g. with internationally acting labs like @CVFCarrington & @UWI_StAugustine)

This separation makes sense to me. I think we only need to tag people who are new submitters regardless of whether they are submitting data for a new country, since we mention new countries separately.

> Several small adjustments needed for open tweets (links, tagging of Genbank?)

Will probably go with something like this for the open tweet this time around:
```
Thanks to #opendata sharing via @NCBI, we've updated nextstrain.org/ncov/open with 56421 new #COVID19 #SARSCoV2 sequences during the last month!
  
1/4


We have received our first sequences from Guinea. Check them out on nextstrain.org/ncov/gisaid/africa!

2/4
```

eharkins (approved):

Issue #698, opened 2021-07-30
Left join multiple metadata files for the same named input


huddlej:
**Context**
A common use case for the ncov workflow is for users to download GISAID data for a specific analysis (e.g., a state-focused build) and combine other existing metadata they have about those state-level data (based on GISAID accession or strain name) with the GISAID data. This joining of public GISAID data and public/private local data allows users to create trees that can be colored, filtered, etc. by custom metadata annotations.

This type of joining requires programming and bioinformatics expertise that can be a burden or a complete roadblock to users trying to accomplish this task. The benefit of a merged metadata file in this case over a drag-and-drop metadata file in Auspice is the persistent state in the build JSONs that can be more easily shared with other users.

**Description**
Add support for left joining standard GISAID/GenBank metadata with custom metadata files on strain name and/or standard accession ids. This merging should be supported per named input for a given `builds.yaml` file, so users can define input-specific custom metadata or reuse the same custom metadata across multiple inputs as they choose.

**Possible solution**
One possible solution is to allow users to define multiple metadata paths for the same input like so:

```yaml
inputs:
  - name: mydata
     sequences: data/sequences.fasta
     metadata: ["data/metadata.tsv", "data/local_metadata.tsv"]
```

We can modify the sanitize metadata script to accept one or more metadata inputs (nargs="+"). When multiple inputs are provided, we can merge these files on predefined columns (strain name, accession id, etc.) and stream the resulting data frame as the final output.

If we use this approach, where should we define the columns to join on? There is already [a configuration section in the builds YAML for the sanitize metadata step](https://nextstrain.github.io/ncov/configuration.html#sanitizemetadata), so we could add a list of columns to join on by default that the user could override as needed.

jameshadfield:
This would be a big improvement. The sanitize metadata script [already has code](https://github.com/nextstrain/ncov/blob/master/scripts/sanitize_metadata.py#L235-L249) for inferring the column representing "strain", would requiring this to be present in all files be too onerous?  If we want to specify a column name to merge on using `config["sanitize_metadata"]` are we ok with this being the same for all inputs (not a big limitation)? We'll probably need an extra step to resolve duplicates for this column as well.

PR #702, opened 2021-08-04
Find clusters of closely related strains

huddlej:

This PR is based on work from @josephfauver who has been manually curating cluster annotations for SARS-CoV-2 in Nebraska. This PR attempts to automate the initial discovery of potential clusters by identifying polytomies in a phylogeny such that each polytomy has at least N tips (N=3 by default) and at least N of those tips in the polytomy belong to the same group based on strain-specific values from a predefined metadata column.

By default, strains are grouped by values in the `division` column. Users can change this behavior by changing the parameter `group_by` parameter of the new `cluster` section of the workflow configuration (see docs update in this PR).

The workflow annotates clusters as a separate tab-separated file that includes strain, cluster id, and the value of the group column used to assign strains to the same group. Users can drag-and-drop this file onto their trees in Auspice to view the clusters. We expect users will want to manually curate clusters (e.g., merge or delete) based on their epidemiological insights, expertise, or PII data that isn't available to the clustering algorithm. Users can then merge these curated clusters into their main workflow metadata.

Note that the work proposed here would be much simpler for users if [the workflow could join multiple metadata files from a single named input](https://github.com/nextstrain/ncov/issues/698). For example, most users will likely want to:

1. curate GISAID metadata and sequences to build their tree
2. generate a tree and clusters from these data
3. review and curate clusters in the tree
4. re-run their analysis with the clusters joined into the original GISAID metadata on strain name

## TODO

 - [x] emit clusters in a TSV file instead of a node data JSON file
 - [x] allow clusters to include strains from other groups as long as the original group meets the minimum thresholds (i.e., allow clusters to grow in future builds without disappearing)
 - [x] compare clusters with [manually annotated clusters for the same tree](https://nextstrain.org/community/josephfauver/ncovNE@main/Build-13?c=Cluster_ID&label=clade:B.1.617.2)
 - [x] name clusters deterministically
 - ~generate clusters as a default output of the workflow~ (not a feature we want, after all)

## Testing

 - [x] CI
 - [x] Custom South Africa build
 - [x] [Custom Nebraska build](https://nextstrain.org/staging/ncov/nebraska)

## Additional context

Thank you to @BryanTegomoh for requesting this feature!

### DIFF
--- a/workflow/snakemake_rules/main_workflow.smk
+++ b/workflow/snakemake_rules/main_workflow.smk
@@ -1231,6 +1231,31 @@ rule calculate_epiweeks:
             --output-node-data {output.node_data} 2>&1 | tee {log}
         """
 
+rule find_clusters:
+    input:
+        tree="results/{build_name}/tree_raw.nwk",
+        metadata="results/{build_name}/metadata_adjusted.tsv.xz",
+        mutations="results/{build_name}/nt_muts.json",
+    output:
+        node_data="results/{build_name}/clusters.json",
+    benchmark:
+        "benchmarks/find_clusters_{build_name}.txt",
+    log:
+        "logs/find_clusters_{build_name}.txt",
+    params:
+        group_by=config["cluster"]["group_by"],
+    resources:
+        mem_mb=12000,
+    shell:
+       """
+       python3 scripts/find_clusters.py \
+           --tree {input.tree} \
+           --metadata {input.metadata} \
+           --mutations {input.mutations} \
+           --group-by {params.group_by} \
+           --output-node-data {output.node_data}
+       """
+
 def export_title(wildcards):
     # TODO: maybe we could replace this with a config entry for full/human-readable build name?
     location_name = wildcards.build_name
@@ -1266,7 +1291,8 @@ def _get_node_data_by_wildcards(wildcards):
         rules.logistic_growth.output.node_data,
         rules.mutational_fitness.output.node_data,
         rules.distances.output.node_data,
-        rules.calculate_epiweeks.output.node_data
+        rules.calculate_epiweeks.output.node_data,
+        rules.find_clusters.output.node_data,
     ]
 
     if "run_pangolin" in config and config["run_pangolin"]:
--- a/workflow/snakemake_rules/main_workflow.smk
+++ b/workflow/snakemake_rules/main_workflow.smk
@@ -1237,7 +1237,7 @@ rule find_clusters:
         metadata="results/{build_name}/metadata_adjusted.tsv.xz",
         mutations="results/{build_name}/nt_muts.json",
     output:
-        node_data="results/{build_name}/clusters.json",
+        clusters="results/{build_name}/clusters.tsv",
     benchmark:
         "benchmarks/find_clusters_{build_name}.txt",
     conda:
@@ -1255,7 +1255,7 @@ rule find_clusters:
            --metadata {input.metadata} \
            --mutations {input.mutations} \
            --group-by {params.group_by} \
-           --output-node-data {output.node_data}
+           --output {output.clusters}
        """
 
 def export_title(wildcards):
@@ -1294,7 +1294,6 @@ def _get_node_data_by_wildcards(wildcards):
         rules.mutational_fitness.output.node_data,
         rules.distances.output.node_data,
         rules.calculate_epiweeks.output.node_data,
-        rules.find_clusters.output.node_data,
     ]
 
     if "run_pangolin" in config and config["run_pangolin"]:
--- a/workflow/snakemake_rules/main_workflow.smk
+++ b/workflow/snakemake_rules/main_workflow.smk
@@ -1240,6 +1240,8 @@ rule find_clusters:
         node_data="results/{build_name}/clusters.json",
     benchmark:
         "benchmarks/find_clusters_{build_name}.txt",
+    conda:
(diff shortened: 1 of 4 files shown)


huddlej:
As an example of what this looks like, I ran the "find_clusters" script on a custom South Africa build and looked for all clusters in Delta.

image

Zooming in, I found clusters like this one in Zambia:

image

However, because I'm using the adjusted metadata, I also found false positives like this one where two strains are from Spain and one is from France but all three have the same adjusted division of "Europe":

image

In addition to considering whether we should use the adjusted metadata or the original metadata (probably the latter), the current cluster script requires all strains in a polytomy to belong to the same group. This means it currently ignores clusters where 3 strains belong to the same division but a fourth strain is from a different division. These mixed clusters still seem valuable, so maybe we should support these.

huddlej:
I confirmed this approach works with a Nebraska-focused build based on the open dataset. Here is a view of the whole tree (rooting here is strange because it is using TreeTime's `best` rooting right now; I forgot to force include the reference sequences).

image

Zooming into Delta, we can see a clear cluster of three identical strains from Nebraska along with several other pairs of identical Nebraskan strains that didn't pass the 3-tip minimum threshold:

image

I will post the build config and link to the tree on the staging server after it finishes rebuilding with correct rooting.

workflow/snakemake_rules/main_workflow.smk
```diff
@@ -1124,6 +1124,31 @@ rule logistic_growth:
...
 
+rule find_clusters:
+    input:
+        tree="results/{build_name}/tree_raw.nwk",
+        metadata="results/{build_name}/metadata_adjusted.tsv.xz",
+        mutations="results/{build_name}/nt_muts.json",
+    output:
+        node_data="results/{build_name}/clusters.json",
+    benchmark:
+        "benchmarks/find_clusters_{build_name}.txt",
```
huddlej, line 1135: Reminder to self that I need to add `conda` directive here

huddlej:
The [Nebraska open build with clusters is available on the staging server](https://nextstrain.org/staging/ncov/nebraska?c=cluster_id&f_clade_membership=21A%20%28Delta%29&f_division=Nebraska). The build YAML for the Nebraska build is:

```yaml
# Define inputs
inputs:
  - name: open
    metadata: s3://nextstrain-data/files/ncov/open/metadata.tsv.gz
    filtered: s3://nextstrain-data/files/ncov/open/filtered.fasta.xz
  - name: references
    metadata: data/references_metadata.tsv
    sequences: data/references_sequences.fasta

# Define builds.
builds:
  nebraska:
    region: North America
    country: USA
    division: Nebraska
    subsampling_scheme: state-level

# Define subsampling schemes.
subsampling:
  state-level:
    focal:
      group_by: year month
      max_sequences: 2000
      query: --query "(country == '{country}') & (division == '{division}')"
    country:
      group_by: division year month
      max_sequences: 800
      query: --query "(country == '{country}') & (division != '{division}')"
      priorities:
        type: proximity
        focus: focal
    global:
      group_by: region year month
      max_sequences: 200
```

huddlej:
Talking with @jameshadfield about how to proceed with algorithmic cluster assignment of ncov builds:

  * Accept previous clusters as input
  * Append to previous clusters, if provided, to keep strains in the same cluster
  * Annotate the reason why something was included (or just whether a strain was just added to a cluster: “Y/N”?)
  * Set some reasonable limits on how far down the tree we walk (current limits seem reasonable)
  * Consider annotating maybe when a cluster was added or a strain was added (akin to recency annotations in ncov), allowing Joseph, et al. to filter the tree to the most recent clusters
  * Syntax to exclude strains from clusters

  * Don’t put the script in the workflow, but document how to use it as a standalone tool. This is a manual process, so we don’t want to imply that it is automated.
  * To test out the above ideas, first add functionality to append to previous inputs and then test the script out on a couple of regional builds to get a feel for what the user will experience when editing the output.

huddlej:
@jameshadfield After chatting with Bryan again this morning at office hours, it's clear that this would still be a very helpful feature of the workflow. The current implementation is good enough to identify clusters within groups and name them (mostly) deterministically by hashing the nucleotide sequence of the internal node of the polytomy. This deterministic cluster naming should cover many cases where we had wanted to accept previously annotated clusters.

I'd like to release this as an undocumented and manually-executed feature of the workflow for Bryan and others to test out. This way we don't start including the clusters into all builds by default, but we allow people to manually generate the clusters TSV and then drag it onto their Auspice tree. What do you think?

jameshadfield:
I'm all for releasing beta versions of things like this (with a disclaimer attached). 

> The current implementation is good enough to identify clusters within groups and name them (mostly) deterministically by hashing the nucleotide sequence of the internal node of the polytomy.

I'd be (pleasantly) surprised if this is the case, as only one change will result in a different hash entirely. But let's find out!

huddlej:
> I'd be (pleasantly) surprised if this is the case, as only one change will result in a different hash entirely. 

Yeah, I'm not 100% sure how this will play out, although I suspect that many hashes will be stable given that we're using polytomies with 3 or more tips. That density of identical nucleotide sequences should make most internal nodes stable for a while. I don't know how the inferred sequences will change as more descendants appear from the internal node. The other, perhaps better, solution would be to use the nucleotide sequence of one of the tips in the polytomy. I'll merge this now, though, and let Bryan know he can start trying it out.

PR #703, opened 2021-08-05
Annotate epiweek as color and filter option

huddlej:

Although we plan to eventually support custom start and end date values through Auspice's filter interface, this commit some similar information that is also relevant to epidemiologists in the form of CDC-style epiweeks. We add a new script to calculate epiweeks using the Python library of the same name and produce a node data JSON representation of those values.

Note that while this commit updates the Conda environment file to represent the new dependency on the "epiweeks" package, we would also need to include this library in the Docker base image, if we end up merging this code.

We would also need to update all auspice configs to include "epiweek" as a color-by and filter. This commit updates only the default auspice config.

The resulting annotations (for the Nextstrain CI build) look like this:

5CA37AF1-4810-4707-ADC1-952BF1303EC6

I think we need to use the fancy new Auspice legends syntax for this type of annotation that is both numerical and not a floating point value.

## Testing

 - [x] CI

## Release checklist

 - [x] Determine the version number for the new release by incrementing [the most recent release](https://github.com/nextstrain/ncov/releases) -> "v8".

## Additional context

Thank you to @BryanTegomoh for requesting this feature and @josephfauver for [implementing the original version in his builds](https://nextstrain.org/community/josephfauver/ncovNE@main/Build-13?c=epiweek).

### DIFF
--- a/defaults/auspice_config.json
+++ b/defaults/auspice_config.json
@@ -112,6 +112,11 @@
       "key": "region_exposure",
       "title": "Region of exposure",
       "type": "categorical"
+    },
+    {
+      "key": "epiweek",
+      "title": "Epiweek (CDC)",
+      "type": "continuous"
     }
   ],
   "geo_resolutions": [
@@ -136,7 +141,7 @@
     "division",
     "location",
     "host",
-    "author"
+    "epiweek"
   ],
   "panels": [
     "tree",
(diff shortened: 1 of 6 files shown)


jameshadfield:
Thanks @huddlej - this looks good. I expanded the `rule finalize` to dynamically create a legend using epiweeks which span the entire observed range, which results in auspice generating a colorscale which maps evenly to time:

image

cassiawag:
I think this looks fantastic with @jameshadfield's edits! I also agree that reusing color after 30 weeks is no problem. You could potentially choose an even fewer number of colors (20 or so) as is color's are distinguishable, but barely, and I'm not sure how much this would be improved by 10 fewer colors.

huddlej:
Thank you for the improved color scale, @jameshadfield and for the review, @cassiawag! The annotations look much better now.

I'm going to merge this and release it as a major version, since it does require users to update their software environment to include the epiweeks package.

PR #705, opened 2021-08-10
remove previously masked sites (which were specific to Belgium, proba…

rneher:
…bly no longer an issue) and add nucleotide corresponding to S:142


rneher:
The previously masked sites behave normally and shouldn't be masked anymore. Masking S:142 hopefully makes Delta more stable. But there was one minor issue: if we mask a site that is deleted, it is subsequently interpreted as an unknown nucleotide. I adjusted the masking script to only mask sites that aren't `-`.

PR #706, opened 2021-08-10
Dated freq file from hcov-19-d file

emmahodcroft:
Date the copy of tip_frequencies that has been processed to add `hCoV-19`, rather than the unadjusted copy. Otherwise tips don't match & auspice can't display.

Frequencies broken for all dated builds since we switched to `gisaid` seem broken. 
https://nextstrain.org/ncov/gisaid/global/2021-08-04

Seems to be because tip's don't match - we are currently dating the non-`hCoV-19` adjusted file. This tries to fix that.

## Testing
 Untested!!!

### DIFF
--- a/nextstrain_profiles/nextstrain-ci/builds.yaml
+++ b/nextstrain_profiles/nextstrain-ci/builds.yaml
@@ -33,6 +33,12 @@ subsampling:
         type: "proximity"
         focus: "region"
 
+# Override default frequency settings, so we can estimate frequencies from older
+# data with a fixed time range.
+frequencies:
+  min_date: 2020-01-01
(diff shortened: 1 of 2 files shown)


huddlej (approved):
This fixes the issue with strain name mismatches between the dated tree and tip frequency JSONs.

Our CI dataset is so old that we end up not estimating any frequencies for those data between Aug 2020 and now, so I didn't have a positive control for the change in this PR (the frequencies panel was still blank, but for a different reason). I just pushed a change to the frequency params, so we estimate frequencies between Jan 1, 2020 and May 10, 2020 which corresponds to the date range of our example data.

PR #715, opened 2021-09-01
Update parse_metadata to curate_metadata

MoiraZuber:

This pull request adds an updated version `curate_metadata.py` of our current curation script `parse_metadata.py`, featuring the following changes:
- Removal of all functions that went obsolete when geoLocations were introduced (e.g. variants, wrong_regions)
- More stable parsing of gisaid and genbank metadata (don't load entire file into memory anymore)
- Split manual_adjustments.txt into two files:
   - `geoLocationRules.txt`: Contains rules that ought to be copied over into the official `gisaid_geoLocationRules.tsv` file
   - `manualAnnotationRules.txt`: Use this file to produce annotations that will be added to `gisaid_annotations.tsv` or `genbank_annotations.tsv`. Useful if geoLocationRules cannot be used (e.g. if delimiter '/' is contained in location names)
- New duplicate detection: Instead of keeping a file containing all known duplicates, all duplicates are 're-detected' during every running of the script. This allows for more flexibility and adjustment of existing duplicates and removes the need for an additional input file.
- Consistent duplicate specifiers: The new duplicate detection now also enforces a very strict style in duplicate specification:
   - If a duplicate location is the only one within its country, add the country abbreviation (only exception is USA)
   - If a duplicate location appears several times within the same country, use '(division)' as unique identifier
   - Duplicate divisions are now also covered (by adding the country abbreviation)
- More readable arrangement of suggested rules for e.g. misspelled locations (e.g. separate similarly spelled locations from suggested 'County' additions)
- Adjustment of how lat_longs and color_ordering files are produced:
   - New lat_longs are now automatically sorted into the file, the manual confirmation for each entry was removed
   - Locations and divisions that are missing in color_ordering are not flagged anymore, but directly added to the file instead. While divisions are only added if lat_longs are present, locations are **always** added. This means it's in the responsibility of the user to use the produced `color_ordering.tsv` only when all locations were cleaned up properly, otherwise inconsistencies the metadata will be reflected in color_ordering.
- The automatic annotation production and sorting that was broken is now fixed. Annotations are processed in the following manner:
   - All strains affected by `geoLocationRules`are checked for conflicting existing annotations and potential conflicts are adjusted in the output files
   - All strains affected by `manualAnnotationRules` receive all necessary annotations additionally to checking for conflicts
- Some additional quality-of-life improvements (better readability, easier presentation of rules to copy&paste, print warnings when files are altered...)

## Release checklist

 - [x] This PR introduces new input and output folders for the curation script. Once this is merged, the old folders `config_files_parse_metadata` and `output_files_parse_metadata` can be deleted, including all files within. EDIT: Rename `developer_scripts` to `curate_metadata`, delete `config_files_parse_metadata`, delete `parse_metadata.py`
 - [x] This PR removed a lot of functions of the old script that I considered unnecessary, including outputs like `new_latlongs` or `new_annotations` (instead favoring to directly sort new entries into these files, then copying and replacing the old file). If @eharkins (or anybody else using the script) was still using them, they can easily be re-added!
 - [x] NEW: Check all remaining scripts for missing input/output folders and outdated file paths
 - [x]  Add special strain adjustments from parse_metadata.py (Mink standardization and Lu'An special char removal)
 - [x] Decide whether to re-introduce special country checks (like Belgium)
 - [x] Implement clade date check and auto-exclude
 - [x] Decide on automatic removal of redundant lat_longs


emmahodcroft:
I haven't run these scripts in so long that I'm not sure I'm qualified to really assess this - Eli will be better at that! However from reading the description this seems like some great changes to tidy things up and improve functionality!! Thank you for the work Moira! If Eli is happy with it - then I am too!

emmahodcroft (approved):

MoiraZuber:
> These directories are not tracked in git so I'm assuming you just mean to feel free to delete them locally? 

True! Although the input folder should be tracked since there are tracked files inside, or is this not linked that way?

> That reminds me that we should update the .gitignore (https://github.com/nextstrain/ncov/blob/master/.gitignore#L58) as part of this PR to reflect the input/output directory name changes, so that we continue to ignore those directories from git.

You're right, that would be a good idea. Since we're already at it, we could add files like the new `geoLocationRules.txt` and `manualAnnotationRules.txt` to .gitignore, since there's no need to track them and they are constantly changed locally.

> The directory input_curate_metadata seems like it is replacing config_files_parse_metadata. Would it be more accurate to call it config_curate_metadata?

Agreed.

> While we are renaming things, I think curate_metadata would be a more appropriate name for the directory now called developer_scripts.

Agreed as well.

eharkins:
Thanks for making those changes @MoiraZuber!

> Although the input folder should be tracked since there are tracked files inside, or is this not linked that way

Which input folder do you mean?

> Should I also delete parse_new_sequences.py since it was replaced with tweet_generation.py recently?

I thought the same thing, just wasn't sure if we are intending on starting to use it again in the future for wrong dated sequences etc.

MoiraZuber:
> Which input folder do you mean?

The one that's now called `config_curate_metadata`. I added it to git since it contains a few files that need tracking (`abbreviations.txt`,  `acceptedExposureAdditions.txt`, `internationalExceptions.txt` and the `country_ordering` folder), but added two files that are only used locally to exclude (`geoLocationRules.txt` and `manualAnnotationRules.txt`)

> I thought the same thing, just wasn't sure if we are intending on starting to use it again in the future for wrong dated sequences etc.

True, we might want to re-introduce these checks Is this something we could do in `curate_metadata.py`? This way we could do regular checks on the dates of all sequences, and flag all sequences too early for their clade which are not yet in `exclude.txt`, without needing the diff files that are still broken.

One additional point we need to decide on is whether we want to keep the special checks we had running for Belgium and Slovakia. They were very helpful to deal with large numbers of false divisions by automatically sorting them as locations into the proper division. However, these checks were always rather fragile, and I haven't encountered large numbers of false divisions from these countries recently, so I think we could omit them for the time being.

eharkins (approved):
Ah thanks for the clarification about including/excluding files, makes sense now and I agree with how you've done it!

> we might want to re-introduce these checks Is this something we could do in curate_metadata.py?

That seems like a great idea for all the reasons you point out.

> One additional point we need to decide on is whether we want to keep the special checks we had running for Belgium and Slovakia. They were very helpful to deal with large numbers of false divisions by automatically sorting them as locations into the proper division. However, these checks were always rather fragile, and I haven't encountered large numbers of false divisions from these countries recently, so I think we could omit them for the time being.

I sometimes encounter new variations or genuinely new Slovakia locations and find the dedicated file helpful, not necessarily in terms of automated rule generation but more just in terms of confirming that a location is in the right division and spelled correctly. But Wikipedia is pretty much just as good, so I would be totally ok with that (or something else) as our source of truth for Slovakia, etc.

MoiraZuber:
> That seems like a great idea for all the reasons you point out
I added it to the release checklist.

> But Wikipedia is pretty much just as good, so I would be totally ok with that
Then I'd say we leave any special checks out for now but keep the country_ordering folder in case we'll need it again someday

Another issue that needs discussion: By removing the supervised lat_longs sort (where it would show you the closest entries and ask for confirmation to add the new coordinates), we once again loose a lot of control over the `lat_longs.tsv` file. One point that was never sitting quite right with me is the fact that old, unused lat_longs are not removed from the file. This can cause inconsistencies (e.g. we had a case where a division with an arbitrary name like "Central" was renamed for one country, but then appeared in a different country, causing it to not be flagged since there were lat_longs present for it, and ended up being placed in a different continent than the country it was supposed to be in). My suggestion would be to delete all entries from `lat_longs.tsv` that are not being used, with the potential disadvantage of having to do some work twice if some sequences are removed and re-added to the metadata later.

eharkins:
> I added it to the release checklist. EDIT: I pushed a commit adding this feature, also see discussion in https://bedfordlab.slack.com/archives/CTZKJC7PZ/p1631023012078900)

Ah ok. I was assuming quality control flagging would end up in a separate PR so we could merge in the new metadata curation script ASAP. Should we continue using it to curate metadata as it's in development?

> Then I'd say we leave any special checks out for now but keep the country_ordering folder in case we'll need it again someday

Sounds good!

> My suggestion would be to delete all entries from lat_longs.tsv that are not being used, with the potential disadvantage of having to do some work twice if some sequences are removed and re-added to the metadata later.

Makes sense to me to do it this way, the same way we already do for `color_ordering.tsv` if I'm not mistaken.

MoiraZuber:
> Ah ok. I was assuming quality control flagging would end up in a separate PR so we could merge in the new metadata curation script ASAP. Should we continue using it to curate metadata as it's in development?

Yes, I hadn't realized this would be a bigger point of discussion. I think we could still merge this and sort the details out later.

> Makes sense to me to do it this way, the same way we already do for color_ordering.tsv if I'm not mistaken.

Exactly. I think we could also do this in another PR in case you'd want to go ahead with merging.

eharkins:
@MoiraZuber I'd say let's merge this now and follow it up with two separate PRs:

1. Removing unused lat longs
2. Automatically excluding sequences with dates before their clade dates (and printing these to a list to check later if we want to include any)

I'll let you click merge here in case you prefer to coordinate it with your curation work on monday.

PR #717, opened 2021-09-02
change: use `mamba` instead of `conda` consistently

corneliusroemer:

Use `mamba` instead of `conda` consistently in the ncov tutorial here: https://docs.nextstrain.org/projects/ncov/en/latest/analysis/setup.html#update-the-nextstrain-environment

Right now, bash code for installation uses `mamba` but the update is still done through `conda`. This PR improves consistency by also using `mamba` for updates.


huddlej (approved):
Thank you, @corneliusroemer!

PR #720, opened 2021-09-07
Add 26456T to Beta defining muts

emmahodcroft:
The clade-defining mutations for Beta also define Mu, with the exception of the first position (1059T).  However, in some circumstances (particularly Mu-focused builds), if any sequences in Mu have 1059T, then this can then be designated as "Beta" while "real Beta" is then unlabelled on the tree.

To minimise the risk of this, add another mutation to the definition of Beta so that it is unlikely to be called in the wrong place (particularly inside of Mu).


trvrb:
I've confirmed this works as expected. Sorry for the delayed review.

PR #723, opened 2021-09-20
Remove travel exposure from pipeline

### LINKED ISSUE #646: Remove travel history (exposure)
[Fifteen months ago](https://github.com/nextstrain/ncov/commit/d0c75376bb87996e5d25542a510e5a8a87f1cc3b) we implemented an ad-hoc approach to dealing with samples which were travel associated (e.g. were sequenced in country B but had recently arrived from country A where the infection was believed to have been acquired). At this time we had few sequences, and were actively tracking the travel history for all of them. This implementation required new metadata fields being introduced, e.g. `country_exposure`, and the DTA was performed on this data. We modified the tree such that terminal nodes actually became 2 nodes: an internal node with the country of exposure, and a single child node with the country of sampling (ditto for other geographic resolutions); a solution which was less than ideal for Auspice (e.g. mutations went on _one_ of those branches, so tips often appeared to have no mutations!). 

The situation now is very different. We have exposure information for a vanishingly small minority of sequences, and for some data sources (genbank) we don't have this at all. We also avoid DTA reconstructions for most builds. The time is right to remove this from our workflow. This issue represents a (potentially incomplete) checklist of what needs to be done.

- [ ] Remove the `incorporate_travel_history` rule, remove the script it calls (`./scripts/modify-tree-according-to-exposure.py`, remove the function `_get_exposure_trait_for_wildcards`, and modify the input to the rule `finalize` so that it now uses the auspice JSON from an upstream rule.
- [ ] Update `adjust_regional_meta.py` to no longer rely on or modify `{trait}_exposure`. This may also necessitate further changes. 
- [ ] Add helpful error messages for the main snakemake file (or an imported `deprecations.smk` or similar) which,
  - [ ] If the config file sets `skip_travel_history_adjustment=true` the notify that this setting is now unneeded 
  - [ ] If auspice config JSONs define colourings for exposure traits,notify that these should be updated. The workflow should still complete if these aren't updated, but the colouring will either be unwanted or will be dropped by `augur export` if it no longer appears in the metadata.
  - [ ] If a `builds.yaml` has a top-level key `exposure` then warn that this is no longer used & should be removed
  - [ ] If a `builds.yaml` defines `traits` which use `{trait}_exposure` then either (i) warn that these are no longer used & modify the config data on the fly to use `{trait} instead` or (ii) print an error & exit.
  - [ ] If any subsampling schemes use `{trait}_exposure` in their exclude rules / group-by rules (etc) then we should print an error & exit. 
- [ ] Modify the ingest scripts / workflows to remove the `{trait}_exposure` fields from the metadata. (Not essential to roll out these changes, but it should simplify things for us.)

jameshadfield:
This PR changes the pipeline to not use exposure metadata information,
thus making all builds behave as if
`config["skip_travel_history_adjustment"]` was True.

The Snakemake file contains checks and warns users if their config file specifies
unused parameters. Requesting a trait reconstruction of an exposure
value is a fatal error to ensure users update their configs.

## Related issue(s)

See GitHub issue #646 for more discussion on the rational for this
change.

This commit closes #646

## Testing
- Local tests pass
- example profile builds successfully 
- Using the GitHub actions here to trigger GISAID and GenBank builds - https://github.com/nextstrain/ncov/actions/runs/1252513680

## Release checklist


jameshadfield:
Test builds passed and this is ready for review.

One change is that `_exposure`s no longer appear as color-bys, meaning [the following view](https://nextstrain.org/ncov/gisaid/global?c=region_exposure&r=region) is no longer available. I don't think this is an issue - the data was very sparse and possibly missing many values - but wanted to flag it here. 

image

emmahodcroft:
Sorry, I'm still trying to catch up - but we actually used `exposure` very recently for some sequences in Europe and a new lineage to identify some patterns - so some pipelines may still want to use this exposure info, even if it's not being curated so extensively anymore. How could someone using `ncov` as a base to run things (like me), continue to include this?

emmahodcroft:
Ok, I've caught up a bit. My main reason for changing my thinking on this a bit is that a) we recently found this useful for a lineage @corneliusroemer was looking into and b) I had forgotten how much has to be adjusted in the tree to make exposure work - so it's not something that's so easily added back in or taken out.

@rneher & @corneliusroemer do you have any thoughts on this, given that it proved useful recently? On the other hand I certainly hadn't used it at all for months before now...

trvrb:
Can we remove from the `augur trait` reconstruction? And just leave as an optional colorby? The way that we can color by `pango_lineage`, etc... just based on what's in the metadata TSV? I'm assuming this would have worked for your use case, while still allowing us to clean up the DAG.

jameshadfield:
> so some pipelines may still want to use this exposure info, even if it's not being curated so extensively anymore. How could someone using ncov as a base to run things (like me), continue to include this?

If you add (or keep) `country_exposure` (etc) in your `auspice-config.json`, then this information will still be exported as a colouring (no changes here); however I've removed these for our core builds. If you try to define these as traits to reconstruct, the snakemake pipeline will throw an error, however I can switch this to a warning instead if you'd still like to reconstruct this? What has changed is the ability to run DTA on the exposure values, and then split up terminal branches into two when the tip was sampled in a different location. This was always a hack and I think it's better to no longer do this.

Snakefile
```diff
@@ -98,6 +98,23 @@ if "inputs" not in config:
...
+    else:
+        logger.error("Your config specifies 'skip_travel_history_adjustment=False' however travel history (exposure) has been removed from this pipeline!")
+        sys.exit(1)
+for name,trait_info in config.get("traits", {}).items():
+    fatal = False
+    if len([col for col in trait_info.get("columns", []) if "_exposure" in col])>0:
+        logger.error(f"Build {name} is asking to reconstruct ancestral traits for exposure metadata, however the pipeline no longer uses exposure values.")
+        fatal = True
+    if fatal:
+        sys.exit(1)
```
huddlej, line 116: This very considerate! Do you think we could make this a major (breaking) release, document these changes well, and broadcast the changes on social media instead of keeping these checks in here?

Or, can we at least plan to remove these checks at the next major release after this one?

jameshadfield: ✔️ We can make this a major release. What is the general objection to keeping these checks around? (We can move them to a special `.smk` if we want to keep the main snakefile clean.) My gut tells me that a sizeable fraction of users won't read changelogs, no matter how well we write them!

emmahodcroft: I agree with James here, I think most people don't read changelogs :) We could keep these checks in for a release or a few, then let them go :)

emmahodcroft:
>If you add (or keep) country_exposure (etc) in your auspice-config.json, then this information will still be exported as a colouring (no changes here)

Ah - fantastic! And sorry I didn't catch this. This should be perfect, thank you! And allows others to do the same. While seeing exposure traits can still be useful in some situations, I agree that the reconstruction very likely is not (there are simply too few, too far away), and that given this was always a bit hacky, best to retire this.

Give how complex the implementation is, I would be happy with it being an error - warnings are really easy to miss with our pipeline because there's so much output, and I think it would be best for people not to try to reconstruct unless they really know what they're trying to do, probably.

Snakefile
huddlej: It's definitely a balance between user-friendliness and code bloat. My main arguments for avoiding this kind of code are 1) major releases are designed to allow us to make backward-incompatible (breaking) changes to the workflow to avoid maintaining legacy code or notifications about legacy code and 2) if we commit to supporting warnings for deprecated features (under the assumption that some users never updated their workflow configs), we will never be able to remove this code even when it isn't relevant a year from now.

On the one hand, the kindest move would be to include the warnings in this major release and then remove them in the next major release. On the other hand, this is a task we'd need to keep on our stack and remember for the next release. I'd bet that most people don't regularly update their workflow much less read the change logs. But those people that do update regularly, probably will check the change log.

If people don't read the change log or update their workflow, how can we improve awareness of these changes? I don't think we tweet from @nextstrain when we release new software (this workflow, auspice, augur, nextclade), but maybe we should. I would love for others to be as excited to read our change logs to find the cool new features we offer as I am to read Snakemake's and other tools' change logs. 🤓

jameshadfield:
I'm going to merge this tomorrow (with #730) if there are no objections. @huddlej - we can remove the warnings at some time in the future (e.g. next major release, or the one after that 😉 )

PR #728, opened 2021-09-22
Sanitize metadata in chunks

huddlej:

Instead of loading all metadata into memory to perform transformation
and filtering associated with sanitizing logic, process metadata in
chunks that easily fit in memory.

To support this low-memory implementation, this commit changes the
approach to resolving duplicate records. We now make a first pass
through the metadata to store strain ids and corresponding database ids
for each record in a temporary file on disk. We make a second pass to
filter duplicate records, transform the remaining records, and stream
the results to disk.

This commit adds cram-based functional tests that were used for
test-driven development of the new low-memory functionality and to
confirm that the user interface remains stable.

One minor interface change (improvement) from this refactoring is the
creation of a separate file containing the list of all duplicate strains
when the user has requested an error on duplicates. This change allows
the user to bioinformatically address duplicates after the fact in a way
that printing the list of strains to stderr did not easily allow.

Note that one side effect of the two-pass approach to sanitizing
metadata is that reading metadata from tarballs no longer works. This
functionality will be fixed in a subsequent commit in this branch that
will likely change how we handle data in tarballs.

## Testing

Run doctests:

```
python3 -m doctest scripts/sanitize_metadata.py
```

Run functional tests:

```
cram --shell=/bin/bash tests/sanitize-metadata.t
```

## Release checklist


scripts/sanitize_metadata.py
```diff
@@ -157,6 +148,8 @@ def resolve_duplicates(metadata, strain_field, error_on_duplicates=False):
         formatter_class=argparse.ArgumentDefaultsHelpFormatter,
     )
     parser.add_argument("--metadata", required=True, help="metadata to be sanitized")
+    parser.add_argument("--metadata-id-columns", default=["Virus name", "strain", "name"], nargs="+", help="names of valid metadata columns containing identifier information like 'strain' or 'name'")
+    parser.add_argument("--database-id-columns", default=[], nargs="+", help="names of metadata columns that store external database ids for each record (e.g., GISAID, GenBank, etc.) that can be used to deduplicate metadata records with the same strain names.")
```
tsibley: If `--metadata-id-columns` defaults to the previous hardcoded value (modulo an ordering change¹), then is seems that `--database-id-columns` should too?

¹ Order does matter in the code, so probably should keep it the same as the previous hardcoded value if we're going to provide it as default?

defaults/parameters.yaml
```diff
@@ -11,6 +11,13 @@ strip_strain_prefixes:
   - SARS-CoV-2/
 
 sanitize_metadata:
+  metadata_id_columns:
+    - Virus name
+    - strain
+    - name
```
tsibley, line 17: Should probably keep the previous order of this default, since order is significant in the code.

scripts/sanitize_metadata.py
tsibley: Ah! I see that it gains a default in a subsequent commit. Seems like the change got misplaced in the wrong commit.

defaults/parameters.yaml
```diff
@@ -16,6 +16,7 @@ sanitize_metadata:
     - strain
     - name
   database_id_columns:
+    - "Accession ID"
```
tsibley, line 19: IIUC, this addition is necessary because the database id columns are now used for deduplication _before_ renaming (where `Accession ID=gisaid_epi_isl` below) but previously deduplicating was done _after_ renaming. It's not a behaviour change, as long as the original default renaming rule was left in place; otherwise it could be.

Would it be just as simple (or nearly so) to keep renaming as the first step (or very close to first) and thus leave the semantics of `database_id_columns` and `rename_fields` unchanged?

Either way, it'd be good to document in config and arg help text the order of operations.

scripts/utils.py
```diff
@@ -16,28 +17,34 @@ def extract_tar_file_contents(filename, filetype):
...
-                # corresponding library.
-                internal_file = lzma.open(internal_file, "rt")
-            elif extension == ".fasta":
-                # For sequence data, handle decoding of the binary stream prior
-                # to passing the data back to the caller.
-                internal_file = (line.decode("utf-8") for line in internal_file)
+                if ".xz" in suffixes:
+                    # Check for LZMA-compressed data and open these with the
+                    # corresponding library.
+                    internal_file = lzma.open(internal_file, "rt")
```
tsibley, line 32: I realize the mode here is historic, but why do `.xz` files default to decoding as text? Esp. since later any text streams are re-encoded (presumably back) to UTF-8?

scripts/utils.py
```diff
@@ -16,28 +17,34 @@ def extract_tar_file_contents(filename, filetype):
...
+        # Write the internal file out to an uncompressed temporary file. This
+        # approach allows downstream processes to re-read the file in multiple
+        # passes instead of making a single pass through a stream.
+        with NamedTemporaryFile(delete=False) as temporary_file:
+            temporary_file_path = temporary_file.name
+            for line in internal_file:
+                if not isinstance(line, bytes):
+                    line = line.encode("utf-8")
+
+                temporary_file.write(line)
```
tsibley, line 48: I don't understand why this iterates over `internal_file` writing line-by-ilne (from what it expects might be binary handles, even) to `temporary_file`. I'd expect to see `shutil.copyfileobj()` or `tarfile.extract()` instead. Can you explain it to me? If the reason is only for the conditional UTF-8 encoding, then see my comment about lzma handling above.

defaults/parameters.yaml
huddlej: Good catch! I've fixed this in 79bc714

defaults/parameters.yaml
huddlej: I found the script easier to understand with filtering of duplicates as early as possible. I'm not sure what my logic was in placing that step near the end originally...

But yes, good point about documenting these details. I've updated the script's help and the configuration guide in f4dc7c2 to clarify the order of operations.

scripts/utils.py
huddlej: Good eye on this one and the comment below, @tsibley! The text stream was because the calling code used to consume the returned object from this function directly with BioPython and BioPython expected a text stream. There is no reason to do this extra work now.

As you pointed out below, there's also no reason to write out the contents of these files line by line! I rewrote this logic in 79497a1 to use `extractall` to extract the files into a temporary directory (with their original names) that the calling code can clean up (and that Python will automatically clean up when the temp dir objects get destroyed). These changes make the tar extraction way simpler and also keeps the files in compressed format on disk, etc. I think this is much nicer, but I'm curious what you think about this revision.

tsibley: Awesome. I like this simpler way of doing it.

Since `extract_tar_file_contents()` no longer transparently decompresses `.xz` files, I assume all callers/downstream uses can handle compressed files? IIRC, it was only BioPython which couldn't directly accept an `.xz` file?

huddlej: Yeah, support for compressed files was added a while back in augur and the ncov workflow. The reason not to use compressed files previously was that we were streaming the contents from the tar file.

PR #731, opened 2021-09-23
Update ncov/docs/src/analysis/running.md

philipshirk:

- specify that the user changes the paths to data files in their builds.yaml instead of config.yaml
- change a few typos ("which" to "that")


huddlej:
Thank you, @philipshirk!

PR #733, opened 2021-09-25
stop curating locations

eharkins:
Per discussion in slack: https://bedfordlab.slack.com/archives/C026WS0FJBW/p1632350089038300.

Still need to test my assumption that this will add the uncurated locations from North America to the color ordering as is currently the case for all non North American locations when this script gets run and we copy the result over the default color ordering.

### DIFF
--- a/scripts/curate_metadata/curate_metadata.py
+++ b/scripts/curate_metadata/curate_metadata.py
@@ -489,7 +489,7 @@ def print_missing_places(missing_latlongs):
 # For all missing place names, search the data for similarly spelled names
 # Always search only within the same ordering level as the missing name, as well as below the same higher-level place
 # (e.g. for a missing location, compare only to locations within the same division)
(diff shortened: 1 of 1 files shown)


MoiraZuber:
I am a bit on the fence about deleting location duplicate detection. It would for sure be nice to drop the location level entirely during our curation. However, this would mean we'll get duplicate locations that appear in several countries and/or divisions that are not designated by the country (or division) abbreviation. Would this disrupt the coloring somehow? Also, where and how many times should we add this location then in color_ordering.tsv? 

Apart from that and a small other issue I corrected in my commit (`search_similar_names()` wasn't called anymore although we still need it for division etc.), this looks good to me!

eharkins:
Thanks @MoiraZuber! Decided to tackle this along with the sorting of divisions by occurrences together in #735! Left duplicate location checks in. Would you have a look at that when you get the chance?

PR #735, opened 2021-09-28
drop location and some division curation;

eharkins:
stop checking for missing locations

stop checking for missing divisions
with less than 20 sequences.

Does not stop checking for duplicate locations since this is a more complicated issue (see #733).

### DIFF
--- a/scripts/curate_metadata/curate_metadata.py
+++ b/scripts/curate_metadata/curate_metadata.py
@@ -2,6 +2,7 @@ import os
 from difflib import SequenceMatcher
 from pathlib import Path
 from geopy.geocoders import Nominatim
+from collections import Counter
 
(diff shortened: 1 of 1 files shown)


MoiraZuber:
This solution looks perfect to me! I'll give it a try during curation on Wednesday. The only remark I have is that we decided to curate `open` without any threshold. One simple, dirty solution I can think of would be to update (or initiate, if that's possible) the Counter not with 1, but with 20 if we're reading `open` metadata, so any division there would be above the threshold.

eharkins:
Thanks @MoiraZuber! That makes sense to me as a simple way to handle it for now. I think I'll merge and test it out today and we can adjust as necessary post-merge.

PR #737, opened 2021-09-28
auto parse additional info

eharkins:
Adds `--auto` option to `scripts/curate_metadata/parse_additional_info.py` which doesn't ask for user input and automatically takes all known patterns and converts them to annotations and ignores others.

We still will need a solution to more easily aggregate additional info files over the course of many days so that this can be easily run for an arbitrary number of days' worth of additional info.


MoiraZuber:
This does a good job for me, should be ready to merge

MoiraZuber (approved):

PR #738, opened 2021-09-28
add submitters to footer filters;

eharkins:
adds `author`, `originating_lab`, `submitting_lab`
to footer filters for main nextstrain builds
so that we can link to this instead of trying
to find the submitters' handles on twitter.

### DIFF
--- a/nextstrain_profiles/nextstrain-gisaid/africa_auspice_config.json
+++ b/nextstrain_profiles/nextstrain-gisaid/africa_auspice_config.json
@@ -146,7 +146,10 @@
     "division",
     "location",
     "host",
-    "pango_lineage"
+    "pango_lineage",
+    "author",
+    "originating_lab",
+    "submitting_lab"
   ],
   "panels": [
     "tree",
--- a/nextstrain_profiles/nextstrain-gisaid/asia_auspice_config.json
+++ b/nextstrain_profiles/nextstrain-gisaid/asia_auspice_config.json
@@ -146,7 +146,10 @@
     "division",
     "location",
     "host",
-    "pango_lineage"
+    "pango_lineage",
+    "author",
+    "originating_lab",
+    "submitting_lab"
   ],
   "panels": [
     "tree",
--- a/nextstrain_profiles/nextstrain-gisaid/europe_auspice_config.json
+++ b/nextstrain_profiles/nextstrain-gisaid/europe_auspice_config.json
@@ -146,7 +146,10 @@
     "division",
     "location",
     "host",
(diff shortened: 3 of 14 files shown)


emmahodcroft:
Sorry, I think I'm behind on the history of this. I don't have anything against this as an idea - to point to instead of Twitter tagging - however, I'm concerned this will add a lot of length to the page - further burying the download link and the rest of our very bottom-of-page, behind huge blue-link walls of text. Can we maybe test this so we can see the damage this does in terms of page length & display?

eharkins:
Yes, I was worried about that @emmahodcroft. Just ran a trial build to see how it looks: https://github.com/nextstrain/ncov/runs/3758690775?check_suite_focus=true

emmahodcroft:
Thanks Eli! Can you let me know where I look to see the output?  I tried the URL in the build: https://nextstrain.org/staging/ncov/gisaid/trial/submitters-in-footer/ But couldn't seem to find anything. 
Sorry for not knowing how to get to it!

eharkins:
@emmahodcroft its here https://nextstrain.org/staging/ncov/open/trial/submitters-in-footer/global, that's my fault for not posting the link to begin with (I didn't know exactly where it would show up either)! Here is the format for finding these trial builds in the future: https://github.com/nextstrain/ncov/blob/a94f0cea349ce8e0c23eade4a4ff65cb081bf0e1/docs/dev_docs.md#triggering-trial-builds.

emmahodcroft:
So, after looking, I'm not sure this is a great solution, unfortunately, though I completely get why it would be nice to link to for credit. But, it's very long - about half the page length (for me), which really further buries the footer - including the important Download Data link and the citation.

Also, because we don't (and can't) control redundancy and repetition, it looks pretty bad in some places:
image
image

I'm not sure this is the best solution for not tagging people. Would another option be to just direct link to the author information TSV file, or is this created by Auspice? Could we generate it (or a version of it) automagically and just link to that, instead? It's not a lot better in terms of content, but at least we wouldn't be cluttering up the build page with it...

eharkins:
Yeah that isn't the best-looking, I agree @emmahodcroft. One thing that would make it better is just being able to have that section collapsed by default (and expandable on click) as @jameshadfield has suggested. Otherwise your suggestion of linking to a separate file seems like a workable alternative.

trvrb:
Collapsable filters would be a nice solution here and shouldn't be too difficult. Would pay dividends in a variety of circumstances.

trvrb:
This has now existed for a while. Closing the PR.

PR #739, opened 2021-09-29
Revert action through github interface

eharkins:

Creates a github action that allows those with permissions to run GH actions on this repository to revert either gisaid or open builds (all regions) to a given date-stamped set of builds on s3. Must be run through the github interface similar to how trial builds workflows work in this repo.

One thing I am not sure will work is running `aws s3` cli, since i'm not sure whether this gets installed as a dependency of `nextstrain-cli` in the setup step. We could use `nextstrain-cli` to perform the overwriting of datasets but I couldn't find a good analog for `aws s3 cp` in https://docs.nextstrain.org/projects/cli/en/stable/commands/; I guess we could do it by `nextstrain remote download`ing the desired date-stamped datasets and `nextstrain remote upload`ing them over the live datasets.

## Related issue(s)
This will be helpful once https://github.com/nextstrain/ncov-ingest/pull/212 and #730 are in place in case we don't catch problematic data in quality controls steps in the ncov workflow or need to revert to an old set of datasets on nextstrain.org for any reason.

## Testing

Untested. Could test running the script that the workflow runs but the workflow cannot be triggered until merged, unless we want to try triggering it by `push` and hardcoding some of the `input` values.

### DIFF
--- a/.github/workflows/revert.yml
+++ b/.github/workflows/revert.yml
@@ -1,0 +1,32 @@
+name: Revert nextstrain.org/ncov/gisaid or nextstrain.org/ncov/gisaid
+
+on:
+  # Manually triggered using GitHub's UI
+  workflow_dispatch:
+    inputs:
+      build_name:
+        description: Which set of datasets on nextstrain.org/ncov to reset. "gisaid" or "open"
+        required: true
+      date:
+        description: Date to revert to. A corresponding set of date-stamped datasets must exist on the s3 bucket. Format is YYYY-MM-DD.
+        required: true
+
+env:
+  BUILD_NAME: ${{ github.event.inputs.build_name }}
+  DATE: ${{ github.event.inputs.date }}
+
+jobs:
+  revert:
+    runs-on: ubuntu-20.04
+    steps:
+    - uses: actions/checkout@v2
+
+    - name: Setup
+      run: ./scripts/setup-github-workflow
+
+    - name: Revert build
+      run: |
+        ./scripts/revert "$BUILD_NAME" "$DATE"
+      env:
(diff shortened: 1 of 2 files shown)


eharkins:
Thanks James! James and I talked about this PR in a phone call yesterday, and the answer to this question
>  My biggest comment is the splitting of the job into a separate bash script vs inlining it into the action. I know ncov-ingest has gone for the former, however my preference would be to avoid the extra complexity this introduces. I can see that there is a benefit to having bash scripts, as this allows users to call them locally - was this behind your decision here?

was that being able to run locally (important in general to test future changes to the script, since github actions can't always be triggered in certain ways on a branch) was my main reason for doing this in a separate script. James said it's fine to leave it as is for now, since it's just the one bash script and not overly complex.

Only other thing preventing this from getting merged is the [above question](https://github.com/nextstrain/ncov/pull/739#discussion_r723739827) of `aws s3` vs `nextstrain remote`.

emmahodcroft:
I can't comment on one method through the other, but I'm ashamed to say I never converted to triggering things via command line vs web interface. However, sounds like this might be useful if that's the route this goes. I'd be super keen for a demo on how to revert builds once this is finished, if someone could spare a few minutes! As I figure it will be good for multiple of us to know how to do it, in case we happen to spot something wrong-looking!

tsibley:
@eharkins [wrote](https://github.com/nextstrain/ncov/pull/739#discussion_r723704989):

> Makes sense. Although it will not allow us to continue in the loop with other regions for which the datasets may exist. But maybe rare that dated datasets exist inconsistently across regions for a particular date.

Instead of `exit 1` inside the loop, you could do something like this instead:

```bash
missing_count=0

if …
else
    …
    ((missing_count++))
fi

exit $missing_count
```

PR #741, opened 2021-10-01
Update GISAID guide URL

ms609:
https://nextstrain.github.io/ncov/data-prep now redirects to https://docs.nextstrain.org/projects/ncov/en/latest/analysis/data-prep.html


huddlej:
Thanks, @ms609!

PR #745, opened 2021-10-06
Rebuild workflow

jameshadfield:
WIP

### DIFF
--- a/.github/workflows/preprocess-gisaid.yml
+++ b/.github/workflows/preprocess-gisaid.yml
@@ -1,0 +1,60 @@
+name: preprocess-gisaid
+
+on:
+  ## This workflow is intended to be run via a repository dispatch event sent from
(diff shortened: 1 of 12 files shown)


jameshadfield:
Thanks @joverlee521 - will tidy up the errors you found!

The number of actions are indeed complicated (and spread over multiple repos, some of which run "in" GitHub actions, some of which detach...). It would, of course, be a good idea to document this somewhere. In lieu of that, here's the summary:

* `fetch-and-ingest` jobs run as (ncov-ingest) actions, and the action monitors the AWS job's progress.
* If ☝️ was running on the master branch, it triggers a preprocess action (ncov repo). This happens from the bash script (which is running on AWS).
* The preprocess workflow runs, but does not monitor the AWS job as it may take > 6 hours. At the end, if successful and on master branch, it triggers a rebuild action (ncov repo). This is done via the snakemake rule `trigger_phylogenetic_rebuild` because the GitHub action has long ago finished.
* The rebuild action runs, and the GitHub action stays "attached" to the AWS job.

Issue #748, opened 2021-10-07
Rule upload succeeds even if the actual uploads fail


jameshadfield:
**Current Behavior**

The internal-to-nextstrain rule `upload` will succeed if the underlying bash script it calls exits with a non-zero code.

**Expected behavior**

If the bash script called by `upload` rule exits with a non-zero code (in this case, "./scripts/upload-to-s3") the rule should fail.

**How this was detected:**

A nextstrain profile ran, which invoked the `upload rule`. The logs are [here](https://console.aws.amazon.com/cloudwatch/home?region=us-east-1#logEventViewer:group=%2Faws%2Fbatch%2Fjob;stream=nextstrain-job%2Fdefault%2F5d60f90e993e4f08bd9c5d9cbb8b772f), but these may not last forever. The rule correctly inferred _which_ files to upload, and to where, for instance:
```
Uploading results/aligned_open.fasta.xz → s3://nextstrain-data/files/ncov/open/aligned.fasta.xz
```
However, because the AWS keys used to perform this upload did not have permissions to do this, the upload failed:
```
upload failed: results/aligned_open.fasta.xz to s3://nextstrain-data/files/ncov/open/aligned.fasta.xz
An error occurred (AccessDenied) when calling the CreateMultipartUpload operation: Access Denied
```

The rule succeeded, the workflow continued to run, and the AWS job itself duly "SUCCEEDED". 

**Possible solution**

The problem is our piping of the bash script's output to `tee`, which means we only see the error code of `tee`, not the bash script we actually care about:
```
    run:
        for remote, local in input.items():
            shell("./scripts/upload-to-s3 {local:q} s3://{params.s3_bucket:q}/{remote:q} | tee -a {log:q}")

```

Note that 

```
$( exit 1 ) | tee
echo $?
0
```

Testing indicates that this 🐛 is fixed by changing the invocation to the following:
```
    run:
        for remote, local in input.items():
            shell("set -o pipefail; ./scripts/upload-to-s3 {local:q} s3://{params.s3_bucket:q}/{remote:q} | tee -a {log:q}")
```
Which surprised me, as the [snakemake docs claim that](https://snakemake.readthedocs.io/en/stable/snakefiles/rules.html)  "shell commands in Snakemake use the bash shell in strict mode by default."

~I'll make a PR to fix this now.~ We should also survey all our snakemake invocations to see if this pattern is used elsewhere. _Update: it's used everywhere!!!_

PR #749, opened 2021-10-07
Upload rule fails if the underlying script fails

jameshadfield:
If the underlying script (upload-to-s3) failed, this exit code was
not being passed onto snakemake, instead we were getting the exit
code of `tee` (i.e. 0). This resulted in workflows appearing to succeed
but when looking deeper, the actual uploading failed.

This behaviour surprised me as the docs state that "shell commands in
Snakemake use the bash shell in strict mode by default." [1,2]

[1] https://snakemake.readthedocs.io/en/stable/project_info/faq.html?highlight=strict%20mode#id11
[2] https://snakemake.readthedocs.io/en/stable/snakefiles/rules.html

This is in response to #748, however the problem may be more widespread than just this rule (snakemake rules use the shell throughout our workflows), so this PR may not close that issue.

### DIFF
--- a/workflow/snakemake_rules/export_for_nextstrain.smk
+++ b/workflow/snakemake_rules/export_for_nextstrain.smk
@@ -195,7 +195,7 @@ rule upload:
         "benchmarks/upload.txt"
     run:
         for remote, local in input.items():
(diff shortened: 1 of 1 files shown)


tsibley:
Closed in favor of #751.

PR #755, opened 2021-10-08
Speed up mutation summary

jameshadfield:
Converts the main algorithm of the script to numpy arrays.

Output was tested against the former code to ensure the results were identical.

On my machine, analysing a set of 50k sequences was reduced from 3.45min to 0.42min (8-fold increase).

Running on AWS now on the entire GISAID dataset - will update this message with the results.

### DIFF
--- a/scripts/mutation_summary.py
+++ b/scripts/mutation_summary.py
@@ -5,8 +5,6 @@ from Bio.SeqIO.FastaIO import SimpleFastaParser
 import numpy as np
 import pandas as pd
 
-
-
 def read_reference(fname, genemap):
     try:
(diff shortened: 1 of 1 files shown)


jameshadfield:
AWS runs have successfully completed, and I tested against the old algorithm (GISAID data) to ensure the results are identical (`diff`ing was a bit tricky as the column order isn't stable). 

Speed increases do indeed hold up. Current AWS runs are over 8x faster (for this step), currently taking
- GISAID: Alignment 5.25 hrs, mutation summary 40min
- GenBank: Alignment 2.5 hrs, mutation summary 20min

PR #756, opened 2021-10-08
Only notify slack on non-trial preprocessing actions

jameshadfield:
GitHub actions should not notify slack for trial runs where
the results won't update the "official" data locations.

### DIFF
--- a/.github/workflows/preprocess-gisaid.yml
+++ b/.github/workflows/preprocess-gisaid.yml
@@ -40,7 +40,7 @@ jobs:
             --profile nextstrain_profiles/nextstrain-gisaid-preprocess \
             --config \
               S3_DST_BUCKET=${S3_DST_BUCKET} \
-              slack_token=${SLACK_TOKEN} \
+              ${{ github.event.inputs.trial_name && format('slack_token={0}', env.SLACK_TOKEN) || '' }} \
         |& tee build-launch.log
       env:
         AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
--- a/.github/workflows/preprocess-open.yml
+++ b/.github/workflows/preprocess-open.yml
@@ -40,7 +40,7 @@ jobs:
             --profile nextstrain_profiles/nextstrain-open-preprocess \
             --config \
               S3_DST_BUCKET=${S3_DST_BUCKET} \
-              slack_token=${SLACK_TOKEN} \
+              ${{ github.event.inputs.trial_name && format('slack_token={0}', env.SLACK_TOKEN) || '' }} \
         |& tee build-launch.log
       env:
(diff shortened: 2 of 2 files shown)


tsibley:
> When you don't specify a `slack_token` I guess snakemake will handle that well and just not notify slack vs failing?

Yeah, e.g.

https://github.com/nextstrain/ncov/blob/aebcc274fea1070bd59d5469ecbe198e6e918ea7/workflow/snakemake_rules/export_for_nextstrain.smk#L147

https://github.com/nextstrain/ncov/blob/aebcc274fea1070bd59d5469ecbe198e6e918ea7/workflow/snakemake_rules/export_for_nextstrain.smk#L203

tsibley:
…which, come to think of it, suggests that unsetting the default `slack_channel` might be a better way, e.g.

```diff
diff --git a/.github/workflows/preprocess-open.yml b/.github/workflows/preprocess-open.yml
index 772b5bf8..67c8c4ee 100644
--- a/.github/workflows/preprocess-open.yml
+++ b/.github/workflows/preprocess-open.yml
@@ -41,6 +41,7 @@ jobs:
             --config \
               S3_DST_BUCKET=${S3_DST_BUCKET} \
               slack_token=${SLACK_TOKEN} \
+              ${{ github.event.inputs.trial_name && 'slack_channel="''''"' || '' }} \
         |& tee build-launch.log
       env:
         AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
```

This esp. makes sense if you consider that we could instead want to pipe trial build Slack messages into a separate channel (e.g. `#ncov-genbank-trials`).

tsibley:
Why all the quotes?

```
'slack_channel="''''"'
```

[GitHub Actions expressions](https://docs.github.com/en/actions/learn-github-actions/expressions) only support single-quote string literals, so that's the outermost pair. To write single-quotes inside a single-quoted string, you escape with another single-quote, or double the single-quote. So the innermost set `''''` is the escaped form of `''`, which is a valid YAML representation of the empty string. Snakemake doesn't allow `--config` values of the empty string itself, so we have to encode it as YAML, which Snakemake decodes for us. The intervening double quotes are to preserve the YAML empty string `''` when the workflow expression is interpolated into the shell invocation (instead of the shell treating it as its own string literal).

jameshadfield:
Thanks for the 👀 @eharkins and @tsibley. I've rewritten this to use an inline config setter for the token:
```
${{ github.event.inputs.trial_name && format('slack_token={0}', env.SLACK_TOKEN) || '' }}
```
This will be easily updatable to a custom `slack_channel` in the future if we wish to do this (and, in that case, add an extra line to always use the token).

jameshadfield:
Whoops, that conditional should be flipped. Fixed in 39051183.

PR #772, opened 2021-10-30
fix: sorting prior to applying the crowing penalty inverted

rneher:
for each focal set, we sort the sequences that picked it as closest neighbor by priority (after shuffling to break spurious patterns in the input data) and then spread these priorities using the crowding penalties such close matches to different parts of the focal set are included in the builds (i.e. have highest priorities). The problem, however, was that the initial sort was in reverse order such that the highest priority sequence ended up with the highest crowding penalty. This nullified the proximity priorities in large data sets.

### DIFF
--- a/scripts/priorities.py
+++ b/scripts/priorities.py
@@ -31,7 +31,7 @@ if __name__ == '__main__':
         tmp["priority"] = -tmp.distance - tmp.N*args.Nweight
         name_prior = [(name, d.priority) for name, d in tmp.iterrows()]
         shuffle(name_prior)
-        candidates[focal_seq] = sorted(name_prior, key=lambda x:-x[1])
+        candidates[focal_seq] = sorted(name_prior, key=lambda x:x[1], reverse=True)
 
     # export priorities
     crowding = args.crowding_penalty
--- a/scripts/priorities.py
+++ b/scripts/priorities.py
@@ -31,13 +31,13 @@ if __name__ == '__main__':
         tmp["priority"] = -tmp.distance - tmp.N*args.Nweight
         name_prior = [(name, d.priority) for name, d in tmp.iterrows()]
         shuffle(name_prior)
-        candidates[focal_seq] = sorted(name_prior, key=lambda x:x[1])
+        candidates[focal_seq] = sorted(name_prior, key=lambda x:-x[1])
 
     # export priorities
     crowding = args.crowding_penalty
     with open(args.output, 'w') as fh:
         # loop over lists of sequences that are closest to particular focal sequences
         for cs in candidates.values():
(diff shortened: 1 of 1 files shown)


emmahodcroft:
We uncovered this while I was trying to do some more specific cluster runs. I think the current algorithm works well enough if you've grabbed most of what you expect to be there already - and often it's just hard to check as you don't know what should be there.

I'm re-running some clusters with the new algorithm, but initially I think the default parameters might need adjusting - I'm pulling in a few groups of identicals that we'd probably want to reduce with the crowding penalty. On the other hand I seem to be picking up sequences which are genuinely part of one of the clusters for the first time, though I suppose it's always hard to tell this apart from stochasticity?

rneher:
It doesn't seem to change to behavior much for the regional runs:
https://nextstrain.org/staging/ncov/gisaid/trial/fix-prios/europe?f_region=Asia,Africa

(which would explain why we didn't notice this...)

emmahodcroft:
I'm not surprised that on larger scales we wouldn't see much impact given the number of sequences now. I'm re-running some much more focal clusters today with crowding penalty set to 0.1 (rather than default 0.01) and I'll see how this looks. I do think the change in this PR is probably good - I'm pulling in much closer sequences than initially, but the number of identical clusters near my focal clusters makes me think the crowding might still be skewing things a bit.

emmahodcroft:
I've done 0.1 and it seems good. I thought I'd run at 0.5 just to compare and see if it makes a noticeable difference.

scripts/priorities.py
```diff
@@ -31,13 +31,13 @@
         tmp["priority"] = -tmp.distance - tmp.N*args.Nweight
         name_prior = [(name, d.priority) for name, d in tmp.iterrows()]
         shuffle(name_prior)
-        candidates[focal_seq] = sorted(name_prior, key=lambda x:x[1])
+        candidates[focal_seq] = sorted(name_prior, key=lambda x:-x[1])
```
huddlej, line 34: This is a mostly stylistic point, but we could use `sorted(name_prior, key=lambda x: x[1], reverse=True)` here instead of negating the priority value. It's a more verbose solution to get the same result, but it would clearly highlight that we want to sort in descending order. By comparison, it's easy to overlook the negative sign or its function here.

huddlej (commented):
This looks good to me. My only suggestion below is to be more explicit that we want to sort the priorities in descending order with `reverse=True`.

scripts/priorities.py
rneher: yeah. good point. I had thought that since the thing we sort is negative, the `-` might be sensible. but it is still that the highest value should end up first...

PR #775, opened 2021-11-06
Feat/masking

rneher:
this PR uses the `augur tree --exclude-sites` argument to ignore homoplasic sites during tree reconstruction. In addition, it adds another messy site to the "hard exclude" pattern.


AngieHinrichs:
So from defaults/sites_to_mask.txt it appears that by default, S:N501Y, S:E484K/Q, and S:P681R (the famed "Nelly, Eek and Phil") will be masked?  Does that mean that they won't appear when labeling by amino acid changes, or in the pop-up lists of mutations for samples?  If so, that might surprise a lot of users who are used to seeing them in VoC clades.

Several of the sites are used in [defaults/clades.tsv](https://github.com/nextstrain/ncov/blob/9833274e1595c5eaa9f14ca42776fae2f59bf07b/defaults/clades.tsv).  If the sites disappear from the tree, then it seems like some of the clades might not be labeled?
```
20F     nuc     22992   A
20H (Beta, V2)  nuc     23063   T
20H (Beta, V2)  nuc     23012   A
20I (Alpha, V1) nuc     23063   T
21B (Kappa)     nuc     23012   C
21E (Theta)     nuc     23604   A
21F (Iota)      nuc     21575   T
```

jameshadfield:
@rneher should we remove `rule mask` and use the alignment as input to `rule filter`? (We will also need to remove [the masked output](https://github.com/nextstrain/ncov/blob/master/workflow/snakemake_rules/common.smk#L145) from the upload step.) This should allow the full (unmasked) alignment to be used for state reconstruction. 

>  Does that mean that they won't appear when labeling by amino acid changes, or in the pop-up lists of mutations for samples?

Hey @AngieHinrichs - this only changes the information we use to infer the tree structure. The mutations we report (colourings and on the pop-ups) should be unchanged here. We're actually trying to improve this by reporting mutations we may have previously masked.

rneher:
yes, these are only masked to obtain the tree topology. these sites are then used again to calculate branch length, assign clades, and do ancestral inference. 

@jameshadfield, there is PR https://github.com/nextstrain/ncov/pull/776 on top of this one that does what you propose... (I think/hope)

AngieHinrichs:
Ah that's great @jameshadfield and @rneher, that's something I can't do so easily with the usher tree.  :)

huddlej:
Just noting that I reviewed #776 which seems to be the PR we want to merge instead of this one.

jameshadfield:
Yes, I agree - let's close this and merge #776 instead.

PR #776, opened 2021-11-06
Feat/no global mask

rneher:
There is no need to mask 4M sequences. This moves the mask after the build align step and removes the upload of the masked.fasta.gz file. 

(this PR sits on top of #775)


defaults/parameters.yaml
```diff
@@ -68,6 +68,7 @@ files:
   clades: "defaults/clades.tsv"
   emerging_lineages: "defaults/emerging_lineages.tsv"
   mutational_fitness_distance_map: "defaults/mutational_fitness_distance_map.json"
+  sites_to_mask: "defaults/sites_to_mask.txt"
```
huddlej, line 71: Echoing my comment on Slack (and Angie's question on #775), I think this parameter should be called something like `sites_to_mask_for_tree_building` or `sites_to_mask_for_tree_inference`, etc. so we all know when in the workflow masking happens.

Whatever we end up calling it, we should also have a brief description in [the "files" section of the configuration guide](https://github.com/nextstrain/ncov/blob/master/docs/src/reference/configuration.md#files), to make the feature more visible to users. I'm happy to add this, once we've settled on a name.

defaults/sites_to_mask.txt
```diff
@@ -0,0 +1,12 @@
+1820
```
huddlej, line 1: Do we have an external reference for this list of sites that we could mention as a comment in this file or in the commit message? Or are these sites that we found as problematic? Either way, if we support comments in this file, that could help future us.

huddlej (commented):
Thank you for making this happen, @rneher! I pushed a couple of commits to remove references to the masked output. I don't know if ncov-ingest also needs to be updated in some way to reflect this change.

My biggest question was whether you wanted the inputs to the `aa_muts_explicit` and `ancestral` rules to be the unmasked alignment (as in the PR currrently) or if you'd prefer these to be masked prior to analysis. I only ask because the current workflow runs these rules on masked alignments, so the change to unmasked inputs in this PR could potentially change behavior in existing builds. I realize that the `aa_muts_explicit` rule is complicated by depending on translated outputs from the alignment that will not have been masked, so maybe the current way is the only sane way.

I made a couple of minor comments inline below. I will work on updating augur to work with compressed masked alignments, although the workaround here seems perfect.

defaults/parameters.yaml
rneher: I am happy to have this renamed -- you suggestion sounds good to me.

defaults/sites_to_mask.txt
rneher: I don't think we support comments here, though I agree it would be nice to give masking reasons and at least the corresponding amino acid changes. This is the augur function that parses it:
https://github.com/nextstrain/augur/blob/5ad157485015623883c6b637d247459f906b63cb/augur/utils.py#L700

rneher: The sites mutated convergently in many important clades and include things like S:501Y.

rneher:
My rational was to show that data in as many places as possible, but avoid distortions in the tree. This means 
 * list of sites are excluded for topology, but used for branch lengths
 * the hard masked sites are also excluded from refine (branch lengths) since they are artifacts, not recurrent muts
 * ancestral reconstructions can use full translations without distorting the tree. and we want to show patterns of convergent evolution. 

This PR is a superset of #775 which only implements the masking without skipping the global masking step. if we are sure we want both, we can just focus on this one

jameshadfield (commented):
In principle this is great - I have summarised below, as much for my understanding, but also to check that this was the intention. 

Our annotation of mutations (rules ancestral, translate) will use the (unmasked) build alignment. No sites will be excluded, and so we will report mutations at sites such as 21987 (prone to sequencing error etc); this is the right decision in my opinion. Relatedly, we are using `--infer-ambiguous` which I think we should change.

Tree topology will not consider positions at `Config → mask → mask_sites` nor `Config → files → sites_to_mask`. 

Augur refine won't use sites at `Config → mask → mask_sites` (which aren't in the masked alignment) for branch length inference. Mutations at `Config → files → sites_to_mask`, however, will be used.

rneher:
Yes, your summary reflects the intention. `--ignore-ambiguous` could be dropped for nucleotides, but results in unacceptably noisy views for amino acids as mutations to/from `X` aren't ignored in auspice.

rneher:
I tried to test this as a trial run, but trial runs are still broken....

jameshadfield:
> I tried to test this as a trial run, but trial runs are still broken....

Should be fixed by rebasing to grab [these changes from 4 days ago](https://github.com/nextstrain/ncov/commit/9833274e1595c5eaa9f14ca42776fae2f59bf07b). I just triggered a trial run successfully (to test #783).

rneher:
thanks, James. I hadn't realized that. here is the build info:
```

--> When completed, the following will be available:
build files: s3://nextstrain-staging/files/ncov/open/trial/no-global-mask/
nextstrain URLs: https://nextstrain.org/staging/ncov/open/trial/no-global-mask/REGION_NAME

--> You can attach to this AWS job via:
nextstrain build --aws-batch --attach 91f0c802-0a37-4389-ad86-79fc4bd94b8f /home/runner/work/ncov/ncov
```

rneher:
looks like this is working fine.

rneher:
I made the list of ignored sites for topology slightly more conservative. Ignoring S95 increases the polytomy of Delta. But otherwise the deep topology should be more robust. 

I also renamed the file as John suggested.

PR #777, opened 2021-11-06
remove priority sampling from regional builds in gisaid profile

rneher:
Remove priority sampling from gisaid builds (there are no priorities in open builds)


jameshadfield (approved):
Working well.

PR #778, opened 2021-11-08
Upgrade Python and the toolchain in GitHub Actions workflows

tsibley:
Rationale the same as the corresponding change for Travis addressing the
same issue: "CI: Upgrade Python environment" (6a167de).

For good measure, also always ensure that the Python toolchain (pip,
setuptools, and wheel) are up-to-date before installing packages.


tsibley:
As a spot check, I started a [rebuild of the open phylogenetic outputs](https://github.com/nextstrain/ncov/runs/4142831510?check_suite_focus=true) and it successfully installed nextstrain-cli and started job submission (at which point I cancelled it, satisfied with the change in this PR).

PR #780, opened 2021-11-11
Rebuilds are no longer watched by actions

jameshadfield:
We were running into the 6 hour timelimit for GitHub actions.
We now use the action to trigger the AWS job, and provide information
on how to attach to that / view it in the AWS console. This is the
same as the preprocessing jobs, which also exceed the time limits.

Here are the (GISAID) actions for the last two days which have exceeded 6 hours and thus been killed:
https://github.com/nextstrain/ncov/actions/runs/1440638013
https://github.com/nextstrain/ncov/actions/runs/1445459246

GenBank builds are only taking ~4 hours, but I think it's simpler if the behavior of each is the same.

## Testing

I'll trigger a GitHub action for GISAID now, which (if successful) will rebuild the canonical datasets.
Update: [this action](https://github.com/nextstrain/ncov/actions/runs/1447258764) completed successfully, and farmed the job out to AWS - [AWS console link](https://console.aws.amazon.com/batch/home?region=us-east-1#jobs/detail/4d33e4a3-79fe-4bb2-812c-74168bed16b1).

## Release checklist

Nothing needed

---

@joverlee521 / @victorlin would you take a look at this if possible - it should fix we observed on today's call. Thanks 🙏

### DIFF
--- a/.github/workflows/rebuild-gisaid.yml
+++ b/.github/workflows/rebuild-gisaid.yml
@@ -50,6 +50,7 @@ jobs:
 
         nextstrain build \
           --aws-batch \
+          --detach \
           --cpus 36 \
           --memory 70GiB \
           . \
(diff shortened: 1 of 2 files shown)


victorlin:
Nice! As discussed on the call, this will need to be replaced by a more reliable means of monitoring – either on the AWS side or another GitHub actions workflow. I'm leaning towards exploring options on the AWS side, as monitoring batch jobs is probably a common use case with existing solutions.

victorlin (approved):

jameshadfield:
Confirmed that the AWS job worked as expected (7h36min, for the record!)

PR #782, opened 2021-11-12
Notify slack using threads

jameshadfield:
**Rationale**
The slack messages were becoming problematic due to both the number and
the long-running nature of the job, which made it hard to link messages
to the pipeline which triggered them.

**Solution**
We now use one slack thread per pipeline invocation. Failure messages
are broadcast to the channel as well, so that they are harder to miss.
Message content has also been improved.

**Testing**
Tested running locally and posting to slack channel #scratch

This will be harder to test using GitHub actions, as [we only send slack messages if it's not a trial build](https://github.com/nextstrain/ncov/blob/master/.github/workflows/rebuild-open.yml#L42-L50), but we can change the workflow temporarily to test if desired.

**Implementation details**
Sending slack messages as a thread response requires knowing the
timestamp of the parent message. As snakemake (potentially) runs each
rule in a different process, we use pytools' PersistentDict to do this,
[as recommended by the Snakemake docs](https://snakemake.readthedocs.io/en/stable/project_info/faq.html#i-want-to-pass-variables-between-rules-is-that-possible). To avoid a new dependency
(pytools) I chose to copy PersistentDict's [source code](https://github.com/inducer/pytools/blob/a0d9fdc33f459802b54a79c00a617fbc410e9d4d/pytools/persistent_dict.py) to this repo.
Note that contrary to the snakemake docs, the PersistentDict persists upon future runs
of the pipeline, so we clear it when the pipeline ends. (As we typically
run this on fresh EC2 instances this shouldn't be an issue.)

**Screenshots**

image
image


workflow/lib/persistent_dict.py
```diff
@@ -0,0 +1,891 @@
+"""Generic persistent, concurrent dictionary-like facility."""
+
+"""
+Copied into nextstrain/ncov workflow from pytools version 2021.2.9 unchanged (except for this message).
+https://github.com/inducer/pytools/blob/a0d9fdc33f459802b54a79c00a617fbc410e9d4d/pytools/persistent_dict.py.
+See https://snakemake.readthedocs.io/en/stable/project_info/faq.html#i-want-to-pass-variables-between-rules-is-that-possible
```
victorlin, line 6: I haven't used Snakemake and I'm sure there is good reason for this, but why does the file have to be copied here when the example uses `from pytools.persistent_dict import PersistentDict`?

victorlin: Sorry, didn't see your reasoning in the PR comment. That makes sense, closing this conversation.

victorlin (approved):

workflow/lib/persistent_dict.py
jameshadfield: No worries - for a bit of background here: workflow specific packages are a bit of a pain at the moment. For example, we recently ran into when we added the `epiweeks` package to this pipeline, and...
* users have to upgrade their dependencies. Pretty normal, but problematic as there's so many ways to install nextstrain tools.
* it's not clear what set of packages should be installed -- e.g. "augur" environment vs "ncov" environment. Again, a range of different approaches here.
* the same nextstrain docker image is used to run all our builds, so the package needs to be added there

I know Tom & John have been thinking about this a lot (buildpacks etc), but I don't think changes are imminent.

scripts/upload-to-s3
```diff
@@ -33,15 +33,6 @@ main() {
     if [[ $src_hash != "$dst_hash" ]]; then
         echo "Uploading $src → $dst"
         aws s3 cp --no-progress "$src" "$dst" --metadata sha256sum="$src_hash" "$(content-type "$dst")"
-
-        if [[ $quiet == 1 ]]; then
-            echo "Quiet mode. No Slack notification sent."
-            exit 0
-        fi
-
-        if ! "$bin"/notify-slack "Updated $dst available."; then
```
joverlee521, line 42: Looks like this is the only instance where `notify-slack` is being used. We can delete the script if it's no longer needed or we may replacing it with the new `send_slack_message`. Definitely not blocking this PR.

joverlee521 (approved):
Thanks for diving into this James! It's nice to see how much cleaner the Slack channel can be with the testing in #scratch. 

I made a couple comments below, but nothing blocking the merge of this PR.

workflow/snakemake_rules/export_for_nextstrain.smk
```diff
@@ -227,34 +228,72 @@ rule trigger_phylogenetic_rebuild:
...
+            data["reply_broadcast"]=True
+    except NoSuchEntryError:
+        pass
+
+    response = requests.post("https://slack.com/api/chat.postMessage", headers=headers, data=json.dumps(data))
+    response.raise_for_status()
+    try:
+        storage.fetch("slack_thread_ts")
+    except NoSuchEntryError:
+        storage.store("slack_thread_ts", response.json()["ts"])
```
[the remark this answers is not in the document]
jameshadfield: Nice! Switched to using this in 5d1f7ee

scripts/upload-to-s3
jameshadfield: :+1: deleted in 5d1f7ee

workflow/snakemake_rules/export_for_nextstrain.smk
```diff
@@ -227,34 +228,72 @@ rule trigger_phylogenetic_rebuild:
...
+    # if slack_thread_ts has been stored, then there is a parent message, so we thread this messaege
+    try:
+        data["thread_ts"]=str(storage.fetch("slack_thread_ts"))
+        if broadcast:
+            data["reply_broadcast"]=True
+    except NoSuchEntryError:
+        pass
+
+    response = requests.post("https://slack.com/api/chat.postMessage", headers=headers, data=json.dumps(data))
+    response.raise_for_status()
```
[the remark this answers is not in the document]
jameshadfield: Yeah. Our daily builds are not critical -- they can fail for a few days without any major side-effects, and I'm gambling that slack will sort itself out in less time than matters for our purposes. As you outline, if we don't explicitly set a slack channel & token these steps are skipped, which would allow us to deploy builds from our local machines.

workflow/snakemake_rules/export_for_nextstrain.smk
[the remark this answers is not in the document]
jameshadfield: (And also re-ran a test build using the `#scratch` channel to double check.)

PR #786, opened 2021-11-17
fix: gff parser was using old non-standard gff convention

rneher:
The GFF in defaults was non-standard prior to the merge of the translation PR. The official GFF parser in biopython is broken atm. 
https://github.com/chapmanb/bcbb/issues/134

This should fix the parsing in mutation_summary.py -- but this is brittle and hacky


corneliusroemer:
Apparently fixed in new release 0.6.9
https://github.com/chapmanb/bcbb/issues/134

PR #790, opened 2021-11-24
export_for_nextstrain: Use a config "build_date"

### LINKED ISSUE #788: Timestamped builds should represent the day the build was launched
**Current Behavior**
`rule dated_json` produces timestamped build assets by adding the YYYY-MM-DD date to the files. Because this date is generated each time the rule runs, it can produce different dates if the build starts on one day and finishes on the next.

**Expected behavior**
The same date should be used for the entire pipeline, and this should be the day it was launched.

**How to reproduce**

Observed in [this AWS run](https://console.aws.amazon.com/batch/home?region=us-east-1#jobs/detail/9e715ae1-aa5d-4d42-97e6-fdde4a757b56), logs [here](https://console.aws.amazon.com/cloudwatch/home?region=us-east-1#logsV2:log-groups/log-group/$252Faws$252Fbatch$252Fjob/log-events/nextstrain-job$252Fdefault$252Fdc17a0e0de8240c194b6073c7ae5870f). 

~This run produced files such as:~ (see below)

**Possible solution**

As snakemake runs each rule in a separate process, and initialises (?) the snakefile each time, we can't store the output of `get_todays_date()` as a global variable. We could use a `PersistentDict` as I did in #782 - e.g. [code here](https://github.com/nextstrain/ncov/blob/master/workflow/snakemake_rules/export_for_nextstrain.smk#L224).

joverlee521:
~Use a config "build_date" instead of running a function within the
expand function to get the current date. This puts more burden on the
user to declare a build date but it ensures that the build date is
stable for the entire run of the build. This shouldn't be too much of a
burden on our normal operations since we add the "build_date" config to
the rebuild GitHub workflows.~

Use a config "build_date" if it's available, with the default being the
old behavior of running a function within the expand function to get the
current date. This allows users to define build date that is stable for
the entire run of the build. We add the "build_date" config to
the rebuild GitHub workflows to ensure that we don't run into missing
input file errors if a build runs over multiple days.

Fixes #788

### DIFF
--- a/.github/workflows/rebuild-gisaid.yml
+++ b/.github/workflows/rebuild-gisaid.yml
@@ -38,6 +38,7 @@ jobs:
         set -x
 
         declare -a config
+        config+=(build_date=$(date +'%Y-%m-%d'))
         if [[ "$TRIAL_NAME" ]]; then
           config+=(
             S3_DST_BUCKET=nextstrain-ncov-private/trial/"$TRIAL_NAME"
--- a/.github/workflows/rebuild-open.yml
+++ b/.github/workflows/rebuild-open.yml
@@ -39,6 +39,7 @@ jobs:
         set -x
 
         declare -a config
+        config+=(build_date=$(date +'%Y-%m-%d'))
         if [[ "$TRIAL_NAME" ]]; then
           config+=(
             S3_DST_BUCKET=nextstrain-staging/files/ncov/open/trial/"$TRIAL_NAME"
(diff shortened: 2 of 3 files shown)


tsibley (commented):
Thanks for doing this. Since `config["build_date"]` doesn't have a default, will its unconditional use cause our "preprocess" profiles to throw a KeyError? Those profiles appear to load `export_for_nextstrain.smk` too.

joverlee521:
> Thanks for doing this. Since `config["build_date"]` doesn't have a default, will its unconditional use cause our "preprocess" profiles to throw a KeyError? Those profiles appear to load `export_for_nextstrain.smk` too.

Yup, realized this was happening when the Travis run failed. Fixed this in the latest push where the default is the previous behavior of using the `get_todays_date()`.

jameshadfield (approved):
Code looks good to me 💯  - let's merge this as it's failing rather regularly these days.

emmahodcroft:
Thank you so much for the quick fix @joverlee521 ! 🚀

PR #793, opened 2021-11-25
fix rebuild workflows: quote build_date string

joverlee521:
The date string is being interpreted as integers when passed through
the `nextstrain build` command to snakemake. Use double quotes and
escaped single quotes to ensure that the date string gets passed as a
proper string to the snakemake config.

### DIFF
--- a/.github/workflows/rebuild-gisaid.yml
+++ b/.github/workflows/rebuild-gisaid.yml
@@ -38,7 +38,7 @@ jobs:
         set -x
 
         declare -a config
-        config+=(build_date=$(date +'%Y-%m-%d'))
+        config+=(build_date=\'$(date +'%Y-%m-%d')\')
         if [[ "$TRIAL_NAME" ]]; then
           config+=(
             S3_DST_BUCKET=nextstrain-ncov-private/trial/"$TRIAL_NAME"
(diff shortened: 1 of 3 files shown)


albasit363:
config-build-date

tsibley:
@joverlee521 Weird! I'd like to better understand what was happening here so my mental model of Snakemake's `--config` handling ([e.g.](https://github.com/tsibley/blab-standup/blob/master/2021-11-04.md#4-november-2021)) is more complete. Can you point me to more details about the errors/misbehaviour that was happening?

tsibley:
Nevermind, answered this myself with a little more digging!

This is Snakemake behaviour that changed (for the better!) with https://github.com/snakemake/snakemake/commit/3675412e87765929bbed7d7795ad0f3119a3429f. In versions earlier than 5.28.0, `eval()` was used (:scream:), so `2021-11-29` for example failed `int()` and `float()` and then got interpreted as 2021 minus 11 minus 29 = 1981. In versions 5.28.0 and onwards, this quoting isn't strictly necessary, although won't hurt as the YAML loader will deserialize the unquoted `2021-11-29` as a datetime object (which fortunately stringifies back to the same string, so it'd work out).

PR #794, opened 2021-11-26
chore: update color_ordering.tsv

corneliusroemer:
Used the utility script https://github.com/cov-lineages/pango-designation/blob/master/utils/sort_lineages.py to create a sorted list of lineages


emmahodcroft:
Really nice Cornelius - thank you!

PR #800, opened 2021-12-02
Drop metadata records with too many missing fields

### LINKED ISSUE #799: Newlines in GISAID metadata for Omicron break sanitize metadata
## Current Behavior

The sanitize metadata script tries to identify duplicate strain names and resolve these duplicates with GISAID and/or GenBank accession ids. Newlines in GISAID metadata (specifically in the "originating lab" field of the "Input data for Augur pipeline" metadata) cause originating labs to appear in the first column where strain names appear in valid records. This breaks an internal validation check by sanitize metadata and produces the following exception:

```python
  File "scripts/sanitize_metadata.py", line 394, in <module>
    metadata = filter_duplicates(
  File "scripts/sanitize_metadata.py", line 313, in filter_duplicates
    metadata = metadata.reset_index().merge(
  File "/Users/jlhudd/miniconda3/envs/nextstrain/lib/python3.8/site-packages/pandas/core/frame.py", line 9190, in merge
    return merge(
  File "/Users/jlhudd/miniconda3/envs/nextstrain/lib/python3.8/site-packages/pandas/core/reshape/merge.py", line 106, in merge
    op = _MergeOperation(
  File "/Users/jlhudd/miniconda3/envs/nextstrain/lib/python3.8/site-packages/pandas/core/reshape/merge.py", line 709, in __init__
    self._validate(validate)
  File "/Users/jlhudd/miniconda3/envs/nextstrain/lib/python3.8/site-packages/pandas/core/reshape/merge.py", line 1425, in _validate
    raise MergeError(
pandas.errors.MergeError: Merge keys are not unique in left dataset; not a one-to-one merge
```

## Expected behavior

Of course, this error should not happen. The ideal behavior would be to correct the invalid newline in the metadata. This approach requires us to assume that we know what strain names should look like, so we can identify rows that begin with non-strain-like strings and append their content to the preceding row.

For example, @corneliusroemer uses the following Snakemake rule to perform this fix:

```python
rule remove_false_meta_linebreaks:
    input:
        "data/metadata_raw.tsv",
    output:
        "data/metadata.tsv",
    run:
        import re
        with open(input[0], "r") as f:
            string_in = f.read()
        string_out = re.sub(r"\n(?!(hCoV-19/|MN908947))"," ",string_in)
        with open(output[0], "w") as f: 
            f.write(str(string_out))
```

Another less ideal solution would be to drop the metadata following the spurious newlines. This solution leads to data loss, but it does not require any assumptions about the format of the input data. Since spurious newlines could appear in other metadata source (e.g., the user's own local metadata, INSDC records, etc.), a solution that does not depend on strain name formats might be better.

## How to reproduce

1. Go to GISAID's EpiCoV site.
2. Select the "Search" tab
3. Filter results by variant of concern to "Omicron".
4. Download results as "Input for Augur pipeline" 
5. Either extract the metadata from the tarball or run the tarball right into sanitize metadata like so:

```
python3 scripts/sanitize_metadata.py \
  --metadata gisaid_auspice_input_hcov-19_2021_12_01_00.tar \
  --output test.tsv
```

huddlej:

Handles corrupted metadata (e.g., rows containing unquoted newlines in text fields) by filtering out records with more than half of columns missing data. This approach attempts to handle the issue in #799 without using a data-specific solution. This solution may not catch all similar issues in the future, but it does allow us to omit problematic records from analyses at the outset.

## Related issue(s)

Fixes #799

## Testing

 - [x] Tested by CI
 - [x] Tested locally with Omicron metadata from GISAID
 - [x] Tested locally with full GISAID metadata from Nextstrain
 - [x] Tested locally with full open metadata from Nextstrain

### DIFF
--- a/scripts/sanitize_metadata.py
+++ b/scripts/sanitize_metadata.py
@@ -313,7 +313,6 @@ def filter_duplicates(metadata, database_ids_by_strain):
         metadata = metadata.reset_index().merge(
             records_to_keep,
             on=database_id_columns,
-            validate="1:1",
         ).set_index(strain_field)
 
     return metadata
@@ -395,6 +394,24 @@ if __name__ == '__main__':
                     metadata,
                     database_ids_by_strain,
                 )
+
+            # Drop rows with too many empty fields. These are likely problematic
+            # records that should not be included in the downstream analyses
+            # (e.g., records with spurious newlines).
+            max_empty_fields = int(len(metadata.columns) / 2)
+            metadata["_empty_fields"] = pd.isnull(metadata).sum(axis=1)
+
+            if metadata[metadata["_empty_fields"] >= max_empty_fields].shape[0] > 0:
+                print(
+                    "WARNING: Dropping the following strains whose records have too many empty fields.",
+                    file=sys.stderr,
+                )
+                for strain in metadata[metadata["_empty_fields"] >= max_empty_fields].index.values:
+                    print(strain, file=sys.stderr)
+
+                metadata = metadata[metadata["_empty_fields"] < max_empty_fields].copy()
+
+            metadata = metadata.drop(columns=["_empty_fields"])
 
             # Reset the data frame index, to make the "strain" column available
             # for transformation.


huddlej:
Closing in favor of the more appropriate and useful solution in #801

PR #805, opened 2021-12-08
feat: more stable 21A/I/J clades

corneliusroemer:

We've had issues with the stability of 21A/I/J clades over the past few months.

This proposal changes the clade defining mutations so that:
1. 21I/J will always be children of 21A by being a superset of 21A
2. Defining mutations were chosen through covSpectrum to make sure they are present in as many of the sequences as possible, even in potential lower quality ones. The reason 21A was nested below 21J in the past was because it contained defining mutations that were often enough backfilled that trees including them would put the clade label in the wrong place.

21A will now be defined through:
```
21A (Delta)	nuc	21618	G
21A (Delta)	nuc	23403	G
21A (Delta)	nuc	26767	C
21A (Delta)	nuc	28461	G
```

21J contains these and two more on top (these are very stable according to covSpectrum)
```
21J (Delta)	nuc	11332	G
21J (Delta)	nuc	19220	T
```

21I contains 21A muts plus these two (which were already part of old definition)
```
21I (Delta)	nuc	5184	T
21I (Delta)	nuc	9891	T
```


emmahodcroft:
Haven't 'tried' this but have looked at these on the tree and they seem good to me. One note about `21A (Delta)	nuc	23403	G` - it seems to be very ancestral on the tree. Of course, that doesn't hurt - but it probably isn't doing anything for the definition. Just wanted to double check this was the mutation intended?

corneliusroemer:
Fair point @emmahodcroft, this was somewhat on purpose since I didn't want to add mutations that show reversions.

I guess we could just take that mut out, should still be stable.

emmahodcroft:
Yea, I don't think it matter if we take it out or leave it - I just wanted to mostly double-check there hadn't been a transcription error somewhere along the way and this wasn't the intended base! If all is good, then this seems good to me. Let's try to keep an eye on it the next few runs, if we can remember - not quite the same as backwards running through a set of trees, but... we do what we can :)

rneher (approved):

emmahodcroft:
If there's no objection I would vote we merge this in. Our Delta clades keep messing up :/

emmahodcroft (approved):

corneliusroemer:
Richard has approved too, shall we just merge then?

emmahodcroft:
We won't have another build until tomorrow, so I was thinking giving the night in case anyone had a comment was fair - I've got a reminder for first thing tomorrow, as long as we merge it before the build, it should take effect!

PR #817, opened 2021-12-14
Create clade hierarchy file

corneliusroemer:

This PR adds a machine and human readable `clade_hierarchy` file that lists for each clade the parent and also the WHO name (if any).

### DIFF
--- a/defaults/clade_hierarchy.tsv
+++ b/defaults/clade_hierarchy.tsv
@@ -20,6 +20,6 @@ clade	parent	WHO
 21H	20A	Mu
 21I	21A	Delta
 21J	21A	Delta
-21K	20B	Omicron
-21L	20B	Omicron
-
+21K	21M	Omicron
+21L	21M	Omicron
+21M 20B Omicron
--- a/defaults/clade_hierarchy.tsv
+++ b/defaults/clade_hierarchy.tsv
@@ -1,0 +1,25 @@
+clade	parent	WHO
+19B	19A	
+20A	19A	
+20B	20A	
+20C	20A	
+20D	20B	
+20E	20A	
+20F	20B	
+20G	20C	
+20H	20C	Beta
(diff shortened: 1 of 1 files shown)


emmahodcroft:
I think this is a very nice simple file. The main question for me would be if `defaults` is the right place for it - this is mostly a place where we keep files used in the default configuration for runs, and this file isn't "used" as such. However, it may be there's not really a better place, but would be interested if anyone had alternative suggestions for location.

jameshadfield:
:+1: thanks @corneliusroemer. Also cc @victorlin @j23414 as it would be great to have this displayed in the "resources" section of the documentation.

emmahodcroft:
I think from the comments above we're good Cornelius!

emmahodcroft (approved):

j23414:
I feel like there should be a way to automatically check for broken links, maybe like this [gh-action/broken-link-check](https://github.com/marketplace/actions/broken-link-check)? Or some manual command?

For linking to docs and after we decide `defaults` or `source-data`, I assume we're updating the [orientation-files.md](https://github.com/nextstrain/ncov/blob/master/docs/src/analysis/orientation-files.md), under "Workflow configuration files...".

j23414 (approved):

defaults/clade_hierarchy.tsv
```diff
@@ -0,0 +1,25 @@
...
+21D	20A	Eta
+21E	20B	Theta
+21F	20C	Iota
+21G	20D	Lambda
+21H	20A	Mu
+21I	21A	Delta
+21J	21A	Delta
+21K	21M	Omicron
+21L	21M	Omicron
+21M 20B Omicron
```
victorlin, line 25:
```suggestion
21M	20B	Omicron
```

emmahodcroft:
Shall we merge this in?

PR #822, opened 2021-12-15
Add 21M (Omicron) clade, add (Omicron) to 21L

corneliusroemer:

Adds `21M (Omicron)` clade - corresponding to pango lineage B.1.1.529

Adds `(Omicron)` to `21L` in line with WHO/pango convention.

The clade defining mutations are those that are shared between 21K and 21L which means the clades are guaranteed to nest correctly.

The defining mutations do not occur in the Nextclade reference tree other than in Omicron, so specificity is ensured. Sensitivity is also great, since the set of defining mutations is a subset of 21K and 21L.

```
21M (Omicron)	nuc	23525	T
21M (Omicron)	nuc	23599	G
21M (Omicron)	nuc	27259	C
```

image

### DIFF
--- a/defaults/clades.tsv
+++ b/defaults/clades.tsv
@@ -141,9 +141,13 @@ clade	gene	site	alt
 21K (Omicron)	nuc	24424	T
 21K (Omicron)	nuc	27259	C
 
-21L	nuc	21618	T
-21L	nuc	22775	A
-21L	nuc	23525	T
-21L	nuc	23599	G
-21L	nuc	24424	T
-21L	nuc	27259	C
+21L (Omicron)	nuc	21618	T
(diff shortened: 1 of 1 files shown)


rneher (approved):

emmahodcroft:
Is it worth running this as a trial build? However, not sure if we're getting any of the other BA lineages (except BA.1) through into our builds? If not, how does this label look? (In fact, maybe that's worth running a trial build on `global` to see?)

corneliusroemer:
Don't see what a trial build would help us with.

By inspection this will work with very high probability.

If it doesn't, it doesn't crash anything.

After merger, I'll do a new Nextclade data release which is a test then.

PR #825, opened 2021-12-16
fix: 21I (Delta) must contain S:222V

corneliusroemer:
Adds nuc mutation C22227T (S:222V) back into 21I (Delta)

I'm using this branch for the newest nextclade build, should work as test


corneliusroemer:
Works on nextclade build, so can be merged
image

Issue #811, opened 2021-12-09
Use latest Nextclade for post-subsampling alignment instead of Nextalign


huddlej:
**Context**
After we move filtering to follow subsampling, we will be applying sequence diagnostics with a smaller dataset than we do currently. Most of our diagnostics assume that the input data were previously processed by Nextclade (i.e., the diagnostics look for Nextclade columns in the metadata), but this is not likely to be the case for most users.

Nextclade uses Nextalign internally already to do clade annotations and QC. The latest version of Nextclade more clearly surfaces the alignment output.

**Description**
We should use Nextclade instead of Nextalign for our post-subsampling `align` step of the workflow, to get the benefits of the alignment and diagnostics that users will need for the filter step.

fanninpm:
It's important to consider that the `nextstrain/base` Docker image currently only has Nextalign, not Nextclade.

huddlej:
Assigning @sidneymbell as an indication that her team at CZI has volunteered to work on this in an upcoming sprint.

PR #827, opened 2021-12-20
increase clock-filter iqd to 8, effectively disabling it. we prefilte…

rneher:
…r in diagnostic, so this shouldnt matter that much


huddlej (approved):
This makes sense to me (not having tested with Omicron data). Another good reason to [switch to Nextclade for post-subsampling alignment](https://github.com/nextstrain/ncov/issues/811), though, since everyone outside of our team (without Nextclade annotations) will lose clock filtering with this change.

PR #828, opened 2021-12-20
Surface crowding penalty

### LINKED ISSUE #708: Expose crowding penalty from priorities script as a workflow parameter
**Context**
[Based on conversations on the discussion site](https://discussion.nextstrain.org/t/proximity-confusion/646), we learned that [the default crowding penalty](https://github.com/nextstrain/ncov/blob/29c09377addbd7fdca5c94ebd8edd2426f56b92f/scripts/priorities.py#L18) prevents users from subsampling many similar contextual strains for variant-specific builds. The crowding penalty technically does its job as expected in these cases, by penalizing too many similar strains in the same group. However, when this behavior is not desired, users have no recourse to change it other than editing the priorities script.

**Description**
Define the crowding penalty in `defaults/parameters.yaml`, [pass this value to the priorities script as an explicit argument](https://github.com/nextstrain/ncov/blob/e7724efa467812294e1fca6ba7993010631d89f7/workflow/snakemake_rules/main_workflow.smk#L482), and document the new parameter in the workflow configuration guide (`docs/configuration.md`).

sidneymbell:
### N.B. -- @danrlu did all the actual work here, I just nudged it over the finish line and happen to have git permissions :) 

Fixes #708 by surfacing the `crowding_penalty` as a config parameter. 
`config['priorities']['crowding_penalty']`

Noted in changelog and commented in the params yml. @huddlej -- was there somewhere else in the docs you'd like this noted? Happy to add a note wherever. For now, given that we don't actually list out all the parameters in the docs anywhere afaik, I settled for a note in the changelog and a comment in the default params yml itself. 

## Testing

Runs with the example data on my box.

## Release checklist

 - [X] Update `docs/src/reference/change_log.md` in this pull request to document these changes by the date they were added.


huddlej (changes requested):
Thank you, @sidneymbell and @danrlu for making this happen! I actually prefer the original implementation you've used in [the related aspen PR](https://github.com/chanzuckerberg/aspen/pull/882); a top-level `priorities` key is less likely to get confused for a subsampling scheme definition. The subsampling schemes (like `builds` and `inputs` keys) have slightly unusual behavior compared to other top-level parameters like filter, mask, etc. Switching back to this top-level key in the default params should also fix [the CI error from the last commit](https://app.travis-ci.com/github/nextstrain/ncov/builds/243780871#L495).

We have a separate [reference guide for configuration parameters like this](https://github.com/nextstrain/ncov/blob/master/docs/src/reference/configuration.md). The content from the change log would be a perfect addition to a top-level `priorities` section of that file (right after the [partition_sequences param](https://docs.nextstrain.org/projects/ncov/en/latest/reference/configuration.html#partition-sequences)). I can update this file, if you prefer, though. Let me know either way.

sidneymbell:
Thanks for the review, @huddlej ! Just pushed a reversion to the initial implementation with `priorities` as top-level key and added a note in the config docs.

huddlej (approved):

PR #833, opened 2022-01-06
Reduce filter resources requests

huddlej:

Reduces memory requests for augur filter commands from 12GB to 4GB to
account for lower requirements of the latest augur filter and to allow
more subsample jobs to run simultaneously on AWS Batch jobs for
Nextstrain builds.

## Related issue(s)

This issue came up after we merged #814 and saw unexpected lack of parallelism for jobs on AWS Batch.

## Testing

 - [ ] Run open rebuild action and confirm parallelism

### DIFF
--- a/workflow/snakemake_rules/main_workflow.smk
+++ b/workflow/snakemake_rules/main_workflow.smk
@@ -296,8 +296,9 @@ rule subsample:
         max_date = _get_specific_subsampling_setting("max_date", optional=True),
         priority_argument = get_priority_argument
     resources:
-        # Memory use scales primarily with the size of the metadata file.
-        mem_mb=12000
+        # Memory use scales with the number of sequences per group * number of groups.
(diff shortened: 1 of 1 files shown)


huddlej:
The test build ran as expected with open data. The parallelism of the run also improved quite a bit. We can see from the screenshot of the stats view below that 17 subsample rules were able to start at the same time (right after sanitizing metadata) while compared to 5 subsample rules with a recent GISAID run (the difference in datasets not contributing in any way to the number of jobs that could start at once).

Stats for GISAID build with original subsample resource requests:
image

Stats for open build with reduced subsample resource requests:
image

These changes should allow the GISAID builds to run faster, too, even though each subsample command takes longer than the corresponding command with the open data, because more subsample jobs will be able to run at once.

jameshadfield (approved):

PR #836, opened 2022-01-13
Add new seq to exclude

emmahodcroft:
The reason this is a PR is because the `exclude` file is not showing up the same in the GitWebEditor as it is on Github. So I'm concerned that something is wrong with the editor, and don't want to mess up master.


emmahodcroft:
Yea, this has a huge conflict because the Git Web Editor wasn't loading the entire exclude file. Don't know what's going on. 

Anyway, resolved conflict by just adding to the end.... will merge in.

This sequences is Omicron but lacks 20B so can mess up where the 20B label is placed (detected on Europe Builds).

PR #841, opened 2022-01-22
Feat/reversion contamination filter

rneher:
filter sequences based on their number of reversions and putative contaminants. the relevant QC fields should appear in the metadata next week.

### DIFF
--- a/scripts/diagnostic.py
+++ b/scripts/diagnostic.py
@@ -36,8 +36,8 @@ if __name__ == '__main__':
     parser.add_argument("--clock-filter-recent", type=float, default=20, help="maximal allowed deviation from the molecular clock")
     parser.add_argument("--clock-filter", type=float, default=15, help="maximal allowed deviation from the molecular clock")
     parser.add_argument("--snp-clusters", type=int, default=1, help="maximal allowed SNP clusters (as defined by nextclade)")
-    parser.add_argument("--rare-mutations", type=int, default=15, help="maximal allowed rare mutations as defined by nextclade")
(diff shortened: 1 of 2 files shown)


scripts/diagnostic.py
```diff
@@ -36,8 +36,8 @@ def earliest_clade_date(Nextstrain_clade, clade_emergence_dates_filename, window
     parser.add_argument("--clock-filter-recent", type=float, default=20, help="maximal allowed deviation from the molecular clock")
     parser.add_argument("--clock-filter", type=float, default=15, help="maximal allowed deviation from the molecular clock")
     parser.add_argument("--snp-clusters", type=int, default=1, help="maximal allowed SNP clusters (as defined by nextclade)")
-    parser.add_argument("--rare-mutations", type=int, default=15, help="maximal allowed rare mutations as defined by nextclade")
```
corneliusroemer, line 39: We'll be throwing out very diverged sequences with unlabeled mutations, but that's probably fine since the remaining clock-filter will catch them.

scripts/diagnostic.py
```diff
@@ -36,8 +36,8 @@ def earliest_clade_date(Nextstrain_clade, clade_emergence_dates_filename, window
...
-    parser.add_argument("--clock-plus-rare", type=int, default=25, help="maximal allowed rare mutations as defined by nextclade")
+    parser.add_argument("--reversion-mutations", type=int, default=5, help="maximal allowed reversion mutations as defined by nextclade")
```
corneliusroemer, line 39: This keeps the style of keeping all metrics independent.  Especially for the reversions/contamination rule, summing them could make sense, especially because they are independent

PR #842, opened 2022-01-25
Nextclade align

### LINKED ISSUE #811: Use latest Nextclade for post-subsampling alignment instead of Nextalign

sidneymbell:

- [x] Adds rule to download necessary reference data for nextclade
- [x] Replaces nextalign with nextclade in `rule: align`
- [x] Concatenates QC metrics from nextclade output -> metadata
- [x] Adds nextclade to docker -- see Jess's PR in `docker-base`

## Related issue(s)

Fixes #811 

## Testing

Runs successfully on the example dataset within a conda env: ` snakemake --profile ./my_profiles/example --use-conda --conda-frontend mamba`. Travis is currently broken, I suspect because the corresponding PR to add nextclade to the docker installation hasn't landed yet.

## Known limitations  
* Per Ivan's excellent suggestion, it would be ideal to have a separate script that automatically downloads the newest version of nextclade each time it's run. This is OOS for this issue / PR but would make a good future addition.
* Because we are using the `--input-dataset` for nextclade instead of enumerating files, if one of the constituent files goes missing but the directory is OK, snakemake won't know to download its replacement. 

## Docs

## Release
* Nothing backwards-incompatible here, but given that this enables the community to benefit from Richard's excellent PR #841 which significantly improves the quality of omicron trees, I'd perhaps suggest making this a release anyways.


sidneymbell:
I _think_ travis fails because Jess's PR to add nextclade to the docker image hasn't landed yet. 
https://github.com/nextstrain/docker-base/pull/38

Runs to completion on my box with a conda env.

sidneymbell:
@huddlej -- thanks for merging in Jess's docker PR. With that in place, the branch CI now passes. Can't get the PR CI to rerun, but there have been no commits on master since I last merged master into this branch (i.e., if the branch CI passes, the PR CI should be passing, too, but it's 18 hours old and won't rerun).

huddlej:
@sidneymbell No worries! I'm testing this PR out now. I'll probably end up pushing a couple of additional changes with a rebase onto master today.

huddlej:
@sidneymbell I made some minor aesthetic revisions and also ended up copying the script @ivan-aksamentov recommended from the ncov-ingest workflow as described in the commit message for c03336d.

I've tested these changes locally, but I'm going to run a full Nextstrain "open" build, to make sure there aren't any regressions I missed.

sidneymbell:
Awesome, thank you John!

huddlej:
The open builds ran as expected. The only issue that I'm hung up on is that this change will now reannotate our own Nextstrain metadata that has already been annotated by the ncov-ingest workflow. As @ivan-aksamentov points out, there is a better chance for the Nextclade version in this workflow to get out-of-date than in ncov-ingest. Two immediate solutions to this issue appear to be:

1. Don't care about reannotation. If the latest version of Nextclade is not in the Docker image/conda environment, that is a development maintenance issue to solve. There's no reason to add complexity to the workflow to account for what happens in ncov-ingest.

2. Check for existing Nextclade annotations in the metadata and skip reannotation when they exist. This solution is slightly complicated by the likely scenario that users could define one input that uses pre-annotated metadata (e.g., pre-subsampled metadata from the Nextstrain open builds) and one input that uses metadata from GISAID. Records in the second input from GISAID will have missing values in the Nextclade QC columns and will need to be populated from the Nextclade run in the workflow.

I have code that implements solution 2, but I'm actually leaning toward solution 1 as the simpler implementation. I'm curious for thoughts from others on the team though.

tsibley:
@huddlej 

> I have code that implements solution 2, but I'm actually leaning toward solution 1 as the simpler implementation. I'm curious for thoughts from others on the team though.

I'm probably missing something, but (2) seems like it would be a relatively straightforward and short addition to `scripts/join-metadata-and-clades.py` and only very marginal additional complexity to the whole workflow. Is it more complicated than that? (or could you share your code that implements (2)?)

emmahodcroft:
I would vote for 2 given the chance of being out of date. One way to get around partially-filled annotations might be to warn loudly if this is detected and provide a way to manually override? But may also be able to find a way to detect a partially-filled column - we'd never expect this if it's been through our own ingest, right?

huddlej:
Thank you, @emmahodcroft and @tsibley! I pushed the code to check for existing annotations in 11985b182e9cf059b81d8f2c3eec1780a1cc3c39 (for some reason, this commit doesn't appear here in the PR). As described in the commit message, the solution was slightly complicated by the potential for different column names with and without sanitized metadata. I tried to simplify the situation by aligning the name of the amino acid substitutions column with and without sanitized metadata to `aaSubstitutions`.

scripts/join-metadata-and-clades.py
```diff
@@ -82,7 +82,24 @@ def main():
...
+
+    # Check for existing annotations in the given metadata. Skip join with
+    # Nextclade QC file, if those annotations already exist and none of the
+    # columns have empty values. In the case where metadata were combined from
+    # different sources with and without annotations, the "clock_deviation"
+    # column will exist but some values will be missing. We handle this case as
+    # if the annotations do not exist at all and reannotate all columns. We
+    # cannot look for missing values across all expected columns as evidence of
+    # incomplete annotations, since a complete annotation by Nextclade will
+    # include missing values for some columns by design.
```
tsibley, line 95: I have ~no context on the columns produced by Nextclade, but in general reannotating all rows when there is a mix of annotated and unannotated rows seems good to me, as it means the resulting annotations will be consistently from a single Nextclade version. (OTOH, it probably adds a bit of not-strictly-necessary runtime.)

My thought had been to handle mixed annotated/unannotated rows by (1) always annotating everything and then (2) looking at the `*_original` columns to copy back/preserve the existing annotations in preference to the new annotations. But this solution of yours seems better!

scripts/join-metadata-and-clades.py
```diff
@@ -0,0 +1,146 @@
...
+            print(f"Metadata file '{args.first_file}' has already been annotated with Nextclade QC columns. Skipping re-annotation.")
+            metadata.to_csv(args.o, sep="\t")
+            return
+
+    # Read and rename clade column to be more descriptive
+    clades = pd.read_csv(args.second_file, index_col=NEXTCLADE_JOIN_COLUMN_NAME,
+                         sep='\t', low_memory=False, na_filter = False) \
+            .rename(columns=column_map)
+
+    clades = clades[list(column_map.values())]
```
sidneymbell, line 109:
```suggestion
    clades = clades[c for c in list(column_map.values()) if c in clades.columns.values]
```

PR #848, opened 2022-01-29
Update mutational fitness calculation

trvrb:
This uses the latest results from [Obermeyer et al](https://www.medrxiv.org/content/10.1101/2021.09.07.21263228v1) available at [github.com/broadinstitute/pyro-cov/blob/master/paper/mutations.tsv](https://github.com/broadinstitute/pyro-cov/blob/master/paper/mutations.tsv) last updated Jan 26. This update includes Omicron lineages BA.1, BA.2 and BA.3 resulting in "mutational fitness" estimates from sequence data becoming aligned with realized clade success. There is no longer the mismatch with Delta estimated to have higher "mutational fitness" than Omicron.

### Before update

Screen Shot 2022-01-29 at 10 06 33 AM

### After update

Screen Shot 2022-01-29 at 10 07 24 AM


trvrb:
I'm going to go ahead and merge this. It's a helpful update and doesn't have the potential to break anything.

PR #851, opened 2022-02-03
Fix ambiguous date filtering for epiweeks

### LINKED ISSUE #845: Epiweeks annotations crash on invalid date column values
**Current Behavior**
When running the epiweeks annotations with metadata for which the date column has invalid date/time values like `?`, the pandas to_datetime function call we use crashes.

**Expected behavior**
We should convert invalid dates to ambiguous values during metadata sanitization.

**How to reproduce**
I found this issue with GISAID data for "Europe / Poland" and for lineage "BA.3". The command I ran was:

```
python3 scripts/calculate_epiweek.py \
            --metadata results/poland/metadata_adjusted.tsv.xz  \
           --output-node-data results/poland/epiweeks.json
```

The offending record in the epiweeks execution was record 5 below:

```
                                     strain        date
0                  Poland/PZH-UMB-6423/2022  2022-01-04
1                  Poland/PZH-UMB-7193/2022  2022-01-03
2                  Poland/PZH-GUM-5932/2022  2022-01-06
3               Poland/81SNR22_wsserze/2022  2022-01-01
4                  Poland/PZH-GUM-6303/2022  2022-01-03
5   2._Diagtron_Laboratoria_Łukasz_Rąbalski           ?
```

**Possible solution**
Sanitize dates in `sanitize_metadata` step before processing downstream.

huddlej:

Replaces check for "X" ambiguous characters in dates with a more conservative definition of a valid date based on a regular expression applied to the string representation of the date column. This should filter most incomplete date values including the original "2020-XX-XX" type of date that we had previously filtered. These filters should avoid errors in datetime conversion with pandas or epiweek conversion.

## Related issue(s)

Fixes #845

## Testing

 - [x] Added incomplete functional test data entry in unsanitized metadata and tested locally
 - [x] Tested by CI

### DIFF
--- a/scripts/calculate_epiweek.py
+++ b/scripts/calculate_epiweek.py
@@ -26,7 +26,7 @@ if __name__ == '__main__':
 
     # Find records with unambiguous dates. These must be complete date-like
     # records in YYYY-MM-DD format.
-    date_pattern = re.compile(r"\d{4}-\d{2}-\d{2}")
+    date_pattern = re.compile(r"^\d{4}-\d{2}-\d{2}$")
     has_complete_date = metadata["date"].astype(str).apply(lambda date: date_pattern.match(date) is not None)
     metadata_with_dates = metadata.loc[has_complete_date, ["date"]].copy()
 
--- a/scripts/calculate_epiweek.py
+++ b/scripts/calculate_epiweek.py
@@ -1,8 +1,10 @@
 #!/usr/bin/env python3
 import argparse
+from augur.io import read_metadata
 from augur.utils import write_json
 import epiweeks
 import pandas as pd
+import re
 
 
 if __name__ == '__main__':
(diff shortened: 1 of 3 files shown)


scripts/calculate_epiweek.py
```diff
@@ -11,26 +13,22 @@
...
-        }
-    ).fillna("")
+        id_columns=args.metadata_id_columns,
+    )
 
-    # Find records with unambiguous dates.
-    metadata_with_dates = metadata.loc[~metadata["date"].str.contains("X"), ["strain", "date"]].copy()
+    # Find records with unambiguous dates. These must be complete date-like
+    # records in YYYY-MM-DD format.
+    date_pattern = re.compile(r"\d{4}-\d{2}-\d{2}")
```
victorlin, line 29: Add match for start/end, otherwise something like `2020-01-011` passes when it shouldn't.

```suggestion
    date_pattern = re.compile(r"^\d{4}-\d{2}-\d{2}$")
```

victorlin (changes requested):

scripts/calculate_epiweek.py
huddlej: Nice catch, @victorlin! I knew I was searching the start of the string with `re.match`, but I forgot about the end.

victorlin (approved):

PR #854, opened 2022-02-04
Reduce frequencies bandwidth slightly

trvrb:

Omicron's rise in frequency has been extraordinarily swift. The existing frequency estimates used an 18 day (0.05 year) bandwidth. This was wide enough that Delta samples bleed forward and look to be at higher frequency at present than is accurate.

This commit reduces bandwidth to 14 days (0.041 year). This was chosen to capture speed of variant dynamics in a few geographies, while trying to avoid overly noisy estimates due to sample count.

From testing, I believe a 15 day bandwidth is the better compromise than an 18 day bandwidth.

I had initially wanted to go with 12 day bandwidth to capture Omicron, but this created enough noise that I pushed it back up to 15 days.

## Testing

A few direct comparisons.

### "north-america" dataset, filtered to USA

Reducing bandwidth drops Delta's current frequency from 7% to 4% at Jan 22 timepoint. Using all sequences from the US, [we estimate that Delta frequency is 0.6% on this day](https://github.com/blab/rt-from-frequency-dynamics/blob/master/estimates/omicron-countries/omicron-countries_freq-combined-GARW.tsv#L7769). So 4% is more accurate.

#### Existing

Screen Shot 2022-02-03 at 9 22 28 PM

#### PR

usa-pr

### "europe" dataset, filtered to UK

Like the US example, takes Delta frequency at latest timepoint from 9% to 5%, which is an improvement.

#### Existing

Screen Shot 2022-02-03 at 9 33 45 PM

#### PR

uk-pr

### "south-america" dataset, filtered to Peru

Chosen because Peru has had interesting variant circulation in 2021. The narrower bandwidth definitely makes the plot more dynamic / noisy, but I don't think overly so.

#### Existing

Screen Shot 2022-02-03 at 9 25 53 PM

#### PR

peru-pr

### DIFF
--- a/defaults/parameters.yaml
+++ b/defaults/parameters.yaml
@@ -143,7 +143,8 @@ frequencies:
   pivot_interval_units: "weeks"
 
   # KDE bandwidths in proportion of a year to use per strain.
-  narrow_bandwidth: 0.05
+  # using 15 day bandwidth
+  narrow_bandwidth: 0.041
   proportion_wide: 0.0
 
   # Diffusion frequency settings


huddlej (approved):
This looks good to me!

trvrb:
Thanks for the review John. I'm going to go ahead and merge this.

PR #859, opened 2022-02-07
Add author-lab credit to auspice jsons

emmahodcroft:

This adds `originating_lab` and `submitting_lab` (in this order) to the auspice `json` files for GISAID builds, and `authors` to the same for the open builds.

This should allow us to stop tweeting credits from the builds, which is very time-consuming.

## Testing
This has not been tested.

### DIFF
--- a/nextstrain_profiles/nextstrain-gisaid/africa_auspice_config.json
+++ b/nextstrain_profiles/nextstrain-gisaid/africa_auspice_config.json
@@ -131,7 +131,9 @@
     "division",
     "location",
     "host",
-    "pango_lineage"
+    "pango_lineage",
+    "originating_lab",
+    "submitting_lab"
   ],
   "panels": [
     "tree",
--- a/nextstrain_profiles/nextstrain-gisaid/asia_auspice_config.json
+++ b/nextstrain_profiles/nextstrain-gisaid/asia_auspice_config.json
@@ -131,7 +131,9 @@
     "division",
     "location",
     "host",
-    "pango_lineage"
+    "pango_lineage",
+    "originating_lab",
+    "submitting_lab"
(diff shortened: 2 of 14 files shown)


joverlee521:
Both trial runs completed without build errors. 

The [gisaid trial build](https://nextstrain.org/staging/ncov/gisaid/trial/add_authors/global) looks as expected and the collapsed filters work great (>1000 originating labs!) There are definitely some weird values in there, but I don't think we want to dedicate time to cleaning those up. 

The [genbank trial build](https://nextstrain.org/staging/ncov/open/trial/add_authors/global) does not include any authors...I'll poke around in the trial run to see what went wrong.

One thing I do want to flag is that the dropdown for filters gets a little ugly because of the long lab names. (see [auspice issue](https://github.com/nextstrain/auspice/issues/1453))

emmahodcroft:
Ah, nice catch with `author` Jover - thanks!

I had completely forgotten about these showing up in the drop-down as well as in the collapsed menu on the page, and yea - that looks messed up...  Might want to hold merging this until we can come up with a solution as discussed in the other issue.

joverlee521:
The [second genbank trial build](https://nextstrain.org/staging/ncov/open/trial/add_authors_v2/global) looks as expected. The values for authors are definitely messy though.

emmahodcroft:
Thanks @joverlee521 ! Yes, it's a shame they're so messy - and particularly in such a weird way, with those letters after authors. But I think we probably don't want to invest the time/effort into curating this manually. At some point it's probably the onus of submitters to try and be consistent in how they list themselves, if they want this to be easily recognised by others.

emmahodcroft:
Thank you Jover!

Issue #283, opened 2020-03-28
Translate March 27 report to Portuguese


cassiawag:
Hello @EricaBarroso @TNNeto @7glaucio @andersonbrito !
 
The situation report from 2020-03-27 is ready to be translated to Portuguese. A branch exists as `sit-rep_pt`  where this can be done, and the file to translate is `ncov_sit-rep_pt_2020-03-27.md` under the `narratives` folder: https://github.com/nextstrain/ncov/blob/sit-rep_pt/narratives/ncov_sit-rep_pt_2020-03-27.md
 
You can translate directly in this file, but please remember to push often as other people may be working on the translation!
 
After pushing changes, you can preview them here: https://nextstrain.org/community/narratives/nextstrain/ncov@sit-rep_pt/sit-rep/pt/2020-03-27
 
A few new process things for this week:
- We’ve created a translations project board here:  https://github.com/nextstrain/ncov/projects/1 As you work on the translation, please adjust its location in the project board.  
- Please assign yourself to the translation issue if you’re working on it this week. This will help me keep track of which translations are in progress.
- When a translation is ready for publication -- at minimum meaning two people have worked on it and both agree that it’s ready -- please create a pull request to merge to `master` and tag me and the issue in it!
 
**Standard rules for translation:**
Note that the file to translate is in Markdown. In general, it is important _not_ to change any formatting or links. (Although it is okay to update Wikipedia or CDC links to the appropriate language version.) Comments to help with translation are noted in comment symbols like `<! -- Comment -->`. 
 
Please add your name under `translators:` with a tab, dash, and new line, formatted just like authors are. Please also add a link to your professional page (can use Twitter) under `translatorLinks:`.
 
After the report is translated and proofread, we’ll publish it to https://nextstrain.org/ and post on Twitter. Please share the report with your Portuguese-speaking colleagues, so we can get the information out there!
 
Note: You may know of other great resources available in Portuguese; please don’t add them to the situation report. We need to keep content the same across all translations, so we can be confident in the accuracy of what we publish. Thanks for understanding.
 
At minimum, I require two sets of eyes on every translation; one of those people should have a background in the field. Although if you have more, that’s fine too.
 
Please coordinate amongst yourselves to create and adjust a workflow that is suited for your translation team. You can trade off weeks; you can use a live, online editor like [hackmd](https://hackmd.io/) (sign in with GitHub) to translate together, etc. Whatever works best for you.
 
[...]

PR #861, opened 2022-02-07
Surface Nextclade QC metrics as color-bys

huddlej:

Updates Auspice config(s) to include colors for Nextclade QC information that now gets passed through metadata to augur export. When these columns are present in the metadata, users can color their trees by reversions, potential contaminants, and rare mutations. Users can now skip the diagnostic filters in the workflow, to keep as much data in the tree as possible, but still identify the strains that would be have been filtered by diagnostics.

**Note: This PR only updates the default Auspice config at the moment. It will update all configs once we agree on the fields to include.**

## Testing

 - [ ] Test by CI
 - [ ] Test with full Nextstrain builds

## Release checklist

### DIFF
--- a/defaults/auspice_config.json
+++ b/defaults/auspice_config.json
@@ -41,6 +41,21 @@
       "type": "continuous"
     },
     {
+      "key": "reversion_mutations",
+      "title": "Reversion mutations",
+      "type": "continuous"
+    },
+    {
+      "key": "potential_contaminants",
+      "title": "Potential contaminants",
+      "type": "continuous"
+    },
+    {
+      "key": "rare_mutations",
+      "title": "Rare mutations",
+      "type": "continuous"
+    },
+    {
       "key": "location",
       "title": "Location",
       "type": "categorical"


tsibley:
If these are all counts, should they be on some discrete color scale instead so scale values like `0.555` don't show up?

jameshadfield:
> If these are all counts, should they be on some discrete color scale instead so scale values like 0.555 don't show up?

We can add a `"legend"` item to the coloring which will bin them into groups (e.g. "<5", "5-10"). Since we don't know these values ahead of time, the way to do this would be by adding this to the dataset JSON once it's created ([similar to epiweeks](https://github.com/nextstrain/ncov/blob/92202ab92162ffcaed7436b92787aa3068993445/scripts/fix-colorings.py)). 

Here's an example of custom legend entries, albeit for percentages not counts:

image

huddlej:
I don't think the scale appearance is in the scope of this PR. [We had this discussion about S1 mutations a year ago](https://github.com/nextstrain/ncov/pull/620) and never agreed on an outcome. This isn't to say we shouldn't fix the appearance, but this PR is primarily about which columns to surface or not. A fix to the appearance of counts in ncov should be its own separate PR.

tsibley:
Ah, I see. Agreed not in scope here. I didn't realize this wasn't easily supported by Auspice just by flipping the `type` on the `colorings` entry.

corneliusroemer:
That's great! Good to surface this, could help spot regions where there's something going on, either not enough coverage on the Nextclade reference tree or some recombination/amplicon dropout etc.

I can tell you that the expected range of contamination is 0-5, very rarely >10, so making the color scale 0,1,2,3,4,... up to 10, then in steps of 5 or so would make sense.

Reversions appear a bit more often, so maybe 0,1,2,3,4,5-6,7-8,9-10,11-15,>15?

huddlej:
@corneliusroemer Maybe it would be helpful to see what the numbers are for a proper build? From there we could determine whether custom scale(s) would improve the UX...

Are there any other fields you'd think to include beside the three proposed here? Here's the list of what's added by the QC table:

```
divergence
missing_data
nonACGTN
substitutions
deletions
insertions
reversion_mutations
potential_contaminants
rare_mutations
frame_shifts
aaSubstitutions
QC_missing_data
QC_mixed_sites
QC_rare_mutations
QC_snp_clusters
QC_frame_shifts
QC_stop_codons
clock_deviation
```

huddlej:
After discussion in the group, we should pull in the overall QC columns from Nextclade's QC TSV in [the join script](https://github.com/nextstrain/ncov/blob/069d28e858178f93065dca7bdc39fdecab72f49d/scripts/join-metadata-and-clades.py#L19):

 - `qc.overallScore` -> `QC_overall_score`
 - `qc.overallStatus` -> `QC_overall_status`

These columns should go into the default Auspice config and the main Nextstrain builds. We don't need to include the reversion mutations, etc. in the main Nextstrain build config.

ivan-aksamentov:
@huddlej I made 2 symmetrical PRs in ingest and ncov to add the columns:

 - https://github.com/nextstrain/ncov-ingest/pull/283
 - https://github.com/nextstrain/ncov/pull/865

The order of operations:
 - [ ] review and merge PR in ingest [#283](https://github.com/nextstrain/ncov-ingest/pull/283)
 - [ ] run daily ingest to update the metadata
 - [ ] wait until metadata is updated, then merge PR in ncov [#865](https://github.com/nextstrain/ncov/pull/865)
 - [ ] use the new columns (I am not familiar with this part, so that's on augur/auspice gurus)

corneliusroemer:
@huddlej On an actual build the numbers will be pretty much what I outlined above. The problem of automatic scales is that they can be rendered very coarse by a single outlier. If most values are 0-5 but one value comes in at 50, it'll make all the 0-5 appear as one category, even though it'd be best to have something like 0,1,2,3,4,5,5-10,10+.

I also think that just including the overall QC score may not be hugely helpful in our builds for two reasons:
- we filter out most bad sequences in `diagnostic.py`
- There are a lot of potential reasons why something could be bad, I'd be more interested in how it breaks down than just whether it's good or bad

Did I understand it correctly yesterday that the reason for limiting extra color bys to 1 was that we didn't want to exacerbate the Auspice bug https://github.com/nextstrain/auspice/issues/1453 (overlapping items in filter dropdown)?

I can think of use cases for almost all of the fields. There's a general question whether we should use more coarse grained `status` (just good, mediocre, bad) or continuous quantitative scales. I'd probably be in favour of continuous quant scales.

Divergence is useful in time view, just as date is useful in divergence view.

Missing_data, nonACGTN, substitutions, deletions, insertions: not sure

reversion_mutations:  useful to investigate problematic parts of the tree and reason for bad qc
potential_contaminants: useful to investigate problematic parts of the tree and reason for bad qc
rare_mutations: useful to investigate problematic parts of the tree and reason for bad qc
frame_shifts: useful to investigate problematic parts of the tree and reason for bad qc

QC_missing_data, QC_mixed_sites, QC_rare_mutations, QC_snp_clusters, QC_frame_shifts, QC_stop_codons,
For these the question is: score (continuous) vs status (categorical) and there's some overlap with the raw counts above, but should be useful to find out what caused a bad QC. Definitely interesting in the tip-tooltip if one can see all of these at a glance.

clock_deviation: would help quickly spot if basal sequences fit the clock or could be there because of reference backfilling/reversion

huddlej:
Thanks, @ivan-aksamentov and @corneliusroemer!

@ivan-aksamentov, regarding:

> The order of operations:

[6d325f2](https://github.com/nextstrain/ncov/pull/861/commits/6d325f2a3d089044a60e2ae0ca596f87b1c1b732) in this PR adds the same columns to the join script. I'm not too worried about the order of operations, here. If expected Nextclade QC columns don't exist in the user's metadata already, the script will annotate the metadata with the newly generated QC columns from the workflow. Both the Docker image and the conda environment use the latest Nextclade, so these results should be the same as what we'd get from the ingest (unless I'm missing something).

Updating ncov-ingest makes sense, though. The actual use of these columns depends on the user's Auspice config JSON to define the columns as colors or filters (as in the default config in this PR).

@corneliusroemer:

> I also think that just including the overall QC score may not be hugely helpful in our builds

That's right. The purpose of surfacing these data is to help users with their own builds, especially in the case where users want to skip diagnostic filters to show as much data as possible. In this case, these data allow users to identify problematic sequences they allowed into their build. This is a common use case with US public health labs and our Africa CDC partners.

So, that said, we do not need to include these additional columns in our Auspice config JSONs. It should be sufficient to include them in the default and then people can choose to use them or not per build.

> Did I understand it correctly yesterday that the reason for limiting extra color bys to 1 was that we didn't want to exacerbate the Auspice bug

This is an orthogonal issue that @trvrb has been concerned about with the Auspice UI/UX more generally. The problem would definitely be exacerbated by adding all of these columns to the main Nextstrain builds, but I think we generally agree that our main Nextstrain builds don't need all of these columns. @rneher and @trvrb seemed to agree that the overall QC score and/or status might be good enough.

> I can think of use cases for almost all of the fields. There's a general question whether we should use more coarse grained status (just good, mediocre, bad) or continuous quantitative scales. I'd probably be in favour of continuous quant scales.

Yeah, the more we've discussed this, the more I think folks should be able to opt in to whichever fields they find most helpful. I can imagine the quantitative fields have little meaning for people independent of some qualitative thresholding (e.g., "Is 50 a good or bad QC score?").

We just demoed the overall QC, reversion, potential contaminants, and rare mutations fields at office hours and folks were excited about these. Based on this feedback, I think our plan could be:

1. Merge this PR as it is with only changes to the default Auspice config
2. Add overall QC score/status to main Nextstrain build Auspice configs once @ivan-aksamentov's PRs go in
3. Write up a separate PR with a how-to guide about surfacing ncov-specific metadata fields as colors or filters in Auspice.

ivan-aksamentov:
Alright, I merged the https://github.com/nextstrain/ncov-ingest/pull/283 and closed https://github.com/nextstrain/ncov/pull/865. The new columns should be in the metadata after tomorrow's ingest.

PR #864, opened 2022-02-09
Test for Emma: Is this PR based off of Master

emmahodcroft:
Adding a few sequences to exclude (Texas Omicron labelled Oct 2021) - mostly to check this is based off of master rather than being 400 commits behind.


victorlin:
You should be able to use GitHub's branches page to check this: https://github.com/nextstrain/ncov/branches

PR #866, opened 2022-02-10
Use datetime.date.today() in finding Max Freq Date

emmahodcroft:
I ran into this today while trying to copy this function to another build - seems like we need to use datetime.date.today() rather than just date.today()? At least, I couldn't get it to work otherwise - but maybe am missing something?


tsibley (approved):
Thanks, @emmahodcroft. Looks like that code path wasn't tested until now!

@fanninpm Indeed importing `date` would be an alternative fix. There's another place in this file already using `datetime.date.today()`, so good to stick to that here to be consistent.

Issue #863, opened 2022-02-09
Rewrite and reorganize ncov tutorial


huddlej:
**Context**
The ncov tutorial is now outdated and does not represent the primary use case for most users. Additionally, much of the tutorial's contents belong in separate how-to or reference guides.

**Description**
We plan to rewrite and reorganize the tutorial to reflect modern use cases. This issue serves to track progress toward this goal.

abattenhouse:
As an outside user who relies on your documentation, I have noticed how the code has changed but the documentation has not. For example, the description of how to install Nextstrain in a conda using mamba simply does not work. What I do to rebuild my Nextstrain conda from scratch (which sometimes happen when just updating them doesn't work), I use the workflow/envs/nextstrain.yaml file, and also add auspice.

jameshadfield:
> I have noticed how the code has changed but the documentation has not. 

This is a perpetual struggle -- we have to adapt the workflow as the world changes, and this can lead to documentation (especially text-heavy tutorials) falling out of sync. We are obviously trying to improve things here, and the rewrite should be released in a few weeks.

> For example, the description of how to install Nextstrain in a conda using mamba simply does not work.

Would you mind making this a separate issue? This approach _should_ work.

abattenhouse:
Thanks. I've added a new issue for the mamba install.

PR #867, opened 2022-02-10
add: reference build targets

j23414:

Create new `ncov/gisaid/reference` and `ncov/open/reference` build targets. Benefits include simplifying the `ncov-tutorial` writeup. Mainly changed two places in the `build.yaml` files. But let me know if I need to change `traits` or something else.

```
builds:
  reference:
    subsampling_scheme: nextstrain_clades
    auspice_config: nextstrain_profiles/nextstrain-open/clades_auspice_config.json
    title: Genomic epidemiology of SARS-CoV-2 with clade-focused subsampling

subsampling:
  # Custom subsampling logic for clades.
  nextstrain_clades:
    clades:
      group_by: "Nextstrain_clade"
      max_sequences: 300
      sampling_scheme: "--probabilistic-sampling"
```

## Related issue(s)

Related to #863

## Testing

```
git clone https://github.com/nextstrain/ncov.git
cd ncov
git checkout reference_targets

# Check DAG of pipeline
# Should see a "build_name: reference" node in both png files
snakemake \
  -pf all \
  --profile nextstrain_profiles/nextstrain-gisaid \
  -n -p --dag | \
  dot -Tpng > gisaid_dag.png

snakemake \
  -pf all \
  --profile nextstrain_profiles/nextstrain-open \
  -n -p --dag | \
  dot -Tpng > open_dag.png

# Check reference run specifically
snakemake \
  -pf auspice/ncov_open_reference.json \
  --profile nextstrain_profiles/nextstrain-open

# == move output files to get ready for other run (optional)
# mkdir open
# mv results auspice benchmarks logs open/.
#
# snakemake \
#  -pf auspice/ncov_gisaid_reference.json \
#  --profile nextstrain_profiles/nextstrain-gisaid

# Check output files
# View number of samples per clade, should be at least 10 samples each
xzcat results/reference/reference_subsampled_metadata.tsv.xz | \
  awk -F'\t' '{print $19}' |\
  sort | uniq -c | sort -rn | less

# Check tree (not sure if we want a better sampling scheme)
nextstrain view auspice
```


j23414:
Screen Shot 2022-02-10 at 10 32 54 AM

I assume I'll need to play around with subsampling. But merging this will allow us to have a test file for the `ncov-tutorial` refactor. Open to discussion.

j23414:
Hopefully to address:

> One thing I'd wonder is if it's worth trying to increase some early samples - 20B/20A etc

made a minor change to `builds.yaml`:

```
subsampling:
  # Custom subsampling logic for clades.
  nextstrain_clades:
    clades:
      group_by: "Nextstrain_clade"
      max_sequences: 300
      sampling_scheme: "--probabilistic-sampling"
    clades_month:
      group_by: "Nextstrain_clade year month"  #<== ensure at least one sample from the earliest month
      max_sequences: 200
      sampling_scheme: "--probabilistic-sampling"
```

which results in tree:

Screen Shot 2022-02-16 at 10 18 26 AM

Additionally, the full dag of the open build looks like:

```
snakemake \
  -pf all \
  --profile nextstrain_profiles/nextstrain-open \
  -n -p --dag | \
  dot -Tpng > open_dag.png
```

open_dag

Open to other suggestions, otherwise merging this will help `ncov-tutorial` development.

huddlej:
> Is there a way to deploy so we can see how it's looking after adjustment?

@emmahodcroft We can queue up an open and/or GISAID build from this branch with [the repo's GitHub Actions](https://github.com/nextstrain/ncov/actions). The builds get deployed to the staging server (e.g., https://nextstrain.org/staging/ncov/open/trial/add-tree-args-override/africa where I named the run `add-tree-args-override` after the branch I was testing). I just queued an open build from this branch which will put @j23414's tree here when the build completes:

https://nextstrain.org/staging/ncov/open/trial/reference_targets/reference

huddlej (changes requested):
This looks good, @j23414! There are a couple of lines and files I think we can safely delete, for our current use case.

Like @emmahodcroft mentioned, it would be great to see how these trees look when deployed. I've queued up an "open" build from this branch that should show us the tree we'll want for the tutorial. If the tree looks consistently good enough, we can merge.

nextstrain_profiles/nextstrain-gisaid/builds.yaml
```diff
@@ -30,6 +30,9 @@ inputs:
 # These subsampling schemes are defined at the bottom of this file.
 # (They override the defaults)
 builds:
+  reference:
+    subsampling_scheme: nextstrain_clades
+    auspice_config: "nextstrain_profiles/nextstrain-gisaid/clades_auspice_config.json"
```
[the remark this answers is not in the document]
j23414: Agree, the `clades_auspice_config.json` is really a copy of the `global_auspice_config.json` and was included to be consistent with other `builds`.  I didn't know if we needed to customize it so it was just a stub/placeholder.

I can replace it with the `global` one, or happy to drop line 35 all-together.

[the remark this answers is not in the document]
huddlej: I think a custom title for these builds should be enough. It looks like the GISAID build just needs the `title` added back and then we can drop the Auspice config files. Thank you!

nextstrain_profiles/nextstrain-gisaid/builds.yaml
```diff
@@ -215,6 +217,15 @@ subsampling:
...
       exclude: "--exclude-where 'region={region}'"
+  
+  # Custom subsampling logic for group by clade
+  nextstrain_clades:
+    clades:
+      group_by: "Nextstrain_clade"
+      max_sequences: 300
+    clades_month: # To ensure at least one sample from the earliest month
+      group_by: "Nextstrain_clade year month"
+      max_sequences: 200
```
victorlin, line 228: I'm a bit confused, is this supposed to be 2 subsamplings of 300 and 200 for a total of 500 sequences? There are only 318 in the staging dataset that John linked https://nextstrain.org/staging/ncov/open/trial/reference_targets/reference

nextstrain_profiles/nextstrain-gisaid/builds.yaml
```diff
@@ -215,6 +218,17 @@ subsampling:
       max_sequences: 500
       sampling_scheme: "--probabilistic-sampling"
       exclude: "--exclude-where 'region={region}'"
+  
+  # Custom subsampling logic for group by clade
+  nextstrain_clades:
+    clades:
+      group_by: "Nextstrain_clade"
+      max_sequences: 300
+      sampling_scheme: "--probabilistic-sampling"
```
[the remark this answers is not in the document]
victorlin: I can see why @j23414 included it – `--probabilistic-sampling` is explicit in all the other subsampling schemes in this file. Maybe better to add it for consistency?

victorlin (approved):

nextstrain_profiles/nextstrain-gisaid/builds.yaml
j23414: I could see how separate-sampling means there're 2 passes through the dataset (slight inefficiency). I'm open to ways to speed this up. 

* If I only do `clades_month`, then clades with a short duration will almost disappear in the reference tree. 
* If I only do `clades`, then I risk not capturing an early strain. 

Ideally, I'd be listing a set of strains (first instance in a clade) to force-include

victorlin: Ah yeah there are some limitations here, which can be solved with a database approach in augur.

### 1

> separate-sampling means there're 2 passes through the dataset

The bottleneck is loading all the data for every call to `augur filter`. The ideal state is to load the data once then perform queries on it.

### 2

> Ideally, I'd be listing a set of strains (first instance in a clade) to force-include

This isn't possible with the current interface for `augur filter`. In pandas terms, we would need something like:

```py
df.sort_values(['year', 'month'], ascending=True).groupby(['Nextstrain_clade', 'year', 'month']).first()
```

The closest interface available in `augur filter` is `--query` as an alias to [`pandas.DataFrame.query`](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.query.html), which doesn't support this.  A global `groupby` is actually out of the picture now that metadata is read in chunks.

In SQLite, this would be something like:

```sql
WITH grouped_rows AS (
    SELECT
        strain,
        ROW_NUMBER() OVER (
            PARTITION BY Nextstrain_clade, year, month
            ORDER BY year ASC, month ASC
        ) AS group_i
    FROM metadata  -- need to join with extracted year/month
)
SELECT strain
FROM grouped_rows
WHERE group_i = 1
```

victorlin: (No action needed in this PR, you've answered my question!)

jameshadfield: > My understanding was that the two samples were combined and deduplicated

Correct, but I'm surprised that there would be so many duplicates -- circa 182/200 of the "clades_month" strains are also in the "clades" if I'm understanding correctly.  The logs should report how many strains were picked for each filter ("sample") - can you double check these are (close to) 200 and 300?

j23414: Sure thing!

```
snakemake -pf auspice/ncov_open_reference.json \
  --profile nextstrain_profiles/nextstrain-open \
  &> output_capture.txt
```

Checking the stdout/stderr, I can see `clades` sampling:

```
augur filter \
  --metadata nextstrain-data/files/ncov/open/metadata.tsv.gz \
  --include defaults/include.txt \
  --exclude defaults/exclude.txt \
  --group-by Nextstrain_clade \
  --subsample-max-sequences 300 \
  --output-strains results/reference/sample-clades.txt \
  2>&1 | \
  tee logs/subsample_reference_clades.txt
```

and `clades_month` sampling:

```
augur filter \
  --metadata nextstrain-data/files/ncov/open/metadata.tsv.gz \
  --include defaults/include.txt \
  --exclude defaults/exclude.txt \
  --group-by Nextstrain_clade year month \
  --subsample-max-sequences 200 \
  --output-strains results/reference/sample-clades_month.txt \
  2>&1 | \
  tee logs/subsample_reference_clades_month.txt
```

The `clades` sampling returned 262 strains:

```
less logs/subsample_reference_clades.txt
Sampling at 10 per group.
3851953 strains were dropped during filtering
        562 of these were dropped because they were in defaults/exclude.txt
        2 strains were added back because they were in defaults/include.txt
        3851393 of these were dropped because of subsampling criteria
262 strains passed all filters
```

The `clades_month` sampling returned 202 strains:

```
WARNING: Asked to provide at most 200 sequences, but there are 422 groups.
Sampling probabilistically at 0.4761 sequences per group, meaning it is possible to have more than the requested maximum of 200 sequences after filtering.
3852013 strains were dropped during filtering
        562 of these were dropped because they were in defaults/exclude.txt
        707 were dropped during grouping due to ambiguous year information
        22270 were dropped during grouping due to ambiguous month information
        2 strains were added back because they were in defaults/include.txt
        3828476 of these were dropped because of subsampling criteria
202 strains passed all filters
```

Since there's a warning of `there are 422` groups, I'm also happy to increase the number of `clades_month` to 450. What do you think?

j23414: Although my local run seems to have more samples (459 unique strains), hmm...

```
cat results/reference/sample-clades_month.txt \
    results/reference/sample-clades.txt  |\
  head -n2
#> Wuhan/WH01/2019
#> Wuhan-Hu-1/2019
#> LR991346

cat results/reference/sample-clades_month.txt \
    results/reference/sample-clades.txt  |\
  sort | uniq | wc -l
#>      459
```

But then I looked at the fasta files which does eventually filter samples down to 348 strains.

```
ls -ltr results/reference/*.fasta
#> 13M Feb 23 11:11 results/reference/aligned.fasta
#> 13M Feb 23 11:11 results/reference/masked.fasta
#> 11M Feb 23 11:11 results/reference/filtered.fasta
#> 11M Feb 23 11:11 results/reference/masked_filtered.fasta

grep -c ">" results/reference/*.fasta
#> results/reference/aligned.fasta:439
#> results/reference/filtered.fasta:348
#> results/reference/masked.fasta:439
#> results/reference/masked_filtered.fasta:348   #<= this one
```

trvrb:
Thanks for this @j23414. Looking at https://nextstrain.org/staging/ncov/open/trial/reference_targets/reference, I think I'd recommend to switch to just:
```
    clades:
      group_by: "Nextstrain_clade"
      max_sequences: 300
      sampling_scheme: "--probabilistic-sampling"
```
and drop
```
    clades_month: # To ensure at least one sample from the earliest month
      group_by: "Nextstrain_clade year month"
      max_sequences: 200
      sampling_scheme: "--probabilistic-sampling"
```
This is because in the test build you can seen ancestral lineages persisting up through 2021 like so: https://nextstrain.org/staging/ncov/open/trial/reference_targets/reference?f_clade_membership=19A,19B,20A,20B,20C,20D,20E%20%28EU1%29,20F,20G

non-variant

This is not consistent with our understanding of SARS-CoV-2 circulation. Effectively the super rare persistent viruses have been highlighted.

nextstrain_profiles/nextstrain-gisaid/builds.yaml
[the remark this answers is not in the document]
trvrb: I think would make sense to just drop `--probabilistic-sampling` everywhere in this file if it's on by default. Less confusing this way.

nextstrain_profiles/nextstrain-gisaid/builds.yaml
victorlin: I found the `augur filter` command with `snakemake --dry-run`:

```sh
augur filter \
--sequences results/reference/masked.fasta \
--metadata results/reference/metadata_with_index.tsv \
--include defaults/include.txt \
--query "(_length >= 27000)" \
--max-date 2022-02-24 \
--min-date 2019.74 \
--exclude-ambiguous-dates-by any \
--exclude defaults/exclude.txt results/reference/excluded_by_diagnostics.txt \
--exclude-where division='USA'
--output results/reference/filtered.fasta \
--output-log results/reference/filtered_log.tsv
```

and this was the output:

```
113 strains were dropped during filtering
	65 of these were dropped because they were in results/reference/excluded_by_diagnostics.txt
	24 of these were dropped because of 'division=USA'
	17 of these were filtered out by the query: "(_length >= 27000)"
	7 of these were dropped because of their ambiguous date in any
	2 strains were added back because they were in defaults/include.txt
323 strains passed all filters
```

Input to that has 436 strains (1st row is header):

```sh
wc -l results/reference/metadata_with_index.tsv
# 437
```

So the end result is 436 - 113 = 323

j23414:
This new commit incorporates the following changes:

* Removed unnecessary `sampling_scheme: "--probabilistic-sampling"` across all sampling schemes.
* Reverted to the simpler case to avoid highlighting rare persistent clades

```
  # Custom subsampling logic for clades.
  nextstrain_clades:
    clades:
      group_by: "Nextstrain_clade"
      max_sequences: 300
```

Thanks all! And let me know if there're other suggestions

trvrb:
Thanks @j23414! Could you spin out a trial build from this latest commit (and paste link to PR thread)? It would be good to look directly at output.

j23414:
Sure thing! Spun out the trial build (Thanks @victorlin for walking me through the steps!) and tree can be viewed at:

* https://nextstrain.org/staging/ncov/open/trial/reference_targets/reference?c=clade_membership

huddlej (commented):
While testing out a build based on these reference data, I noticed several strains getting dropped by the augur filter steps that we could avoid including in the first place by modifying the reference data subsampling scheme. For example, here is the output of the filter step:

```
48 strains were dropped during filtering
	14 of these were dropped because they were in results/default-build/excluded_by_diagnostics.txt
	21 of these were dropped because of 'division=USA'
	10 of these were filtered out by the query: "(_length >= 27000)"
	3 of these were dropped because of their ambiguous date in any
	2 strains were added back because they were in defaults/include.txt
204 strains passed all filters
```

We can modify the subsampling scheme for the reference data to select strains that will have unambiguous dates and division != USA. The open metadata also have additional columns like `missing_data`, `rare_mutations`, and `potential_contaminants` that we can filter on from the subsampling scheme to avoid adding strains to the reference data that will just get filtered out again. We could even just filter for data with a Nextstrain QC quality status of “good”.  Based on [the quality scores docs for Nextclade](https://docs.nextstrain.org/projects/nextclade/en/stable/user/algorithm/07-quality-control.html#quality-scores), a sequence with 3000 or more missing bases will get marked as “bad” quality. But, since the full-length genome is ~29900 bp, some sequences missing fewer than 3000 bp could still have fewer than the 27,000 bp required by the ncov workflow’s filters. We could try to enforce the length requirement with a separate query entry based on genome length minus missing data.

For example, the following modified subsampling scheme could avoid including strains that will be filtered out by the default build:

```yaml
subsampling:
  nextstrain_clades:
    clades:
      group_by: "Nextstrain_clade"
      max_sequences: 300
      exclude_ambiguous_dates_by: any
      query: --query "(division != 'USA') & (QC_overall_status == 'good') & ((length - missing_data) >= 27000)"
```

I haven’t tested the calculated length in the last subclause of the query, so this may not work. It may be sufficient to just use the overall QC status with the simpler query like so:

```yaml
subsampling:
  nextstrain_clades:
    clades:
      group_by: "Nextstrain_clade"
      max_sequences: 300
      exclude_ambiguous_dates_by: any
      query: --query "(division != 'USA') & (QC_overall_status == 'good')"
```

trvrb:
@j23414: It looks like we need to update the Auspice config to set clade membership as default coloring. Compare:

- https://nextstrain.org/staging/ncov/open/trial/reference_targets/reference
- https://nextstrain.org/ncov/gisaid/global

This also needs the GISAID logo. 

There needs to be a similar file to https://github.com/nextstrain/ncov/blob/master/nextstrain_profiles/nextstrain-gisaid/global_auspice_config.json and https://github.com/nextstrain/ncov/blob/master/nextstrain_profiles/nextstrain-open/global_auspice_config.json specifically for the `reference` targets.

trvrb:
I'd prefer this, it's easy to have a target and the resulting compute is minuscule and if https://nextstrain.org/ncov/open/reference exists then I'd expect as a user of the website for https://nextstrain.org/ncov/gisaid/reference to exist as well.

PR #868, opened 2022-02-10
Convert setup page to reST, update with central installation guide and glossary terms

victorlin:

- Convert setup page to reStructuredText (reST).
    - There is more reST syntax that can be leveraged now (e.g. glossary term links), but I don't want to overload this PR.
- Redirect readers to the updated central installation guide, rather than having duplicate/differing guidance.
    - Users are no longer discouraged from using non-native runtimes.
    - For users using native runtime, added a note to install `epiweeks nextclade pangolin pangolearn` since I believe these are required for running the ncov workflow (though this is not in the current setup docs?).

## Related issue(s)

- #863

## Preview

- current: https://docs.nextstrain.org/projects/ncov/en/latest/analysis/setup.html
- PR: https://nextstrain--868.org.readthedocs.build/projects/ncov/en/868/analysis/setup.html


j23414 (approved):

docs/src/analysis/setup.rst
```diff
@@ -0,0 +1,94 @@
...
+   cd ncov-7/
+
+3. Run a basic analysis with example data
+-----------------------------------------
+
+Run a basic workflow with example data, to confirm that your Nextstrain environment is properly configured.
+
+.. code:: bash
+
+   nextstrain build . --cores 4 --use-conda \
```
huddlej, line 77: Can we drop the `--use-conda` bit here? We need to surface this use case somewhere, for people who run the workflow on HPCs, etc., but this isn't the best place for it.

victorlin: Done: https://github.com/nextstrain/ncov/pull/868/commits/4b78ddcfe38ddc3332cdc3e276f3c8b7d7d775b8

I also added a few more updates to the content.

PR #869, opened 2022-02-14
add: annotation updates from 2022 Feb 9 to 17th

j23414:

Update annotations up to Feb 10th. Let me know if I missed anything.

### DIFF
--- a/defaults/color_ordering.tsv
+++ b/defaults/color_ordering.tsv
@@ -594,6 +594,7 @@ location	Addagutta
 location	Adilabad
 location	Aghapura
 location	Ah Malakpet
+location	Aliyabad
 location	Ams
 location	Bagh Amberpet
 location	Balagunj
@@ -661,6 +662,7 @@ location	Kothagudem
 location	Krishna (Telangana)
 location	Kumrambeem
 location	Kumrambheem
+location	Kumuram Bheem Asifabad
 location	Kuttivellodi
 location	Kuttyvellodi
 location	Mahabubabad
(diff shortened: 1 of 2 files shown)


MoiraZuber:
This generally looks good to me! A few points that might need adjustment:
- Not all of the places you added to lat_longs seem to be real divisions. E.g. `division Kosice - Okolie` seems to be a smaller district, so I would have changed it to `division` Kosice, `location` Kosice-Okolie using the geoLocationRule

`Europe/Slovakia/Kosice - Okolie/	Europe/Slovakia/Kosice/Kosice-Okolie`

(Or maybe `Europe/Slovakia/Kosice - Okolie/*	Europe/Slovakia/Kosice/Kosice-Okolie` depending on whether there are locations downstream of this division)

Please note that all geoLocationRules you add to `scripts/curate_metadata/config_curate_metadata/geoLocationRules.txt` during your curation need to be added to `ncov-ingest/source-data/gisaid_geoLocationRules.tsv` when you finished. The file in `config_curate_metadata` is only a temporary holder for the rules to be applied by the curation script, while the one in `ncov-ingest` will be automatically applied in every ingest.


- I would not use divisions like `Upper Austria / Voecklabruck / Frankenmarkt	47.9855432	13.4191655`. Instead, I'd create a rule to change these to `Upper Austria, Frankenmarkt`. However, this specific case is extra nasty since the division contains the delimiter we use for our geoLocationRules. What I would do in that case is add a rule like this one

`Europe,Austria,Upper Austria / Voecklabruck / Frankenmarkt,	Europe,Austria,Upper Austria, Frankenmarkt`

to `scripts/curate_metadata/config_curate_metadata/manualAnnotationRules.txt`, because this file is able of taking `,` as delimiter in addition to `/`. The script will create annotations based on the rules in this file and add them to `gisaid_annotations.tsv`. Please mark that this file is not tracked in GitHub, and these rules will not be applied automatically from now on like with the normal geoLocationRules. This means that cases like these will have to be corrected every time we run the script.
This issue is quite inconvenient and it would be nice to automize this at some point, since it has been appearing ever more often.

PR #870, opened 2022-02-15
Update emerging lineages for Omicron

jameshadfield:
Currently clade 21M and 21L genomes get assigned 20B as the emerging lineage, as the emerging lineage definitions haven't been updated.

Current situation (branch labels show clade, not emerging lineage):

image

This updates adds the defining mutations for 3 nextstrain clades 21M, 21L and 21K to correspond to emerging lineages B.1.1.529, BA.1 and BA.2, respectively. While B.1.1.529 was there previously, the defining mutations are here updated.

## Testing

Trial build https://github.com/nextstrain/ncov/runs/5194049893, when completed, should produce test datasets at  https://nextstrain.org/staging/ncov/gisaid/trial/update-emerging-lineages-2022-02-15/global etc.


trvrb:
Looks great! Fully support making this change.

PR #871, opened 2022-02-16
Add parameter to enable overriding tree args

huddlej:

Adds a configuration parameter to the workflow, `override_default_args`,
in the `tree` section that allows users to toggle the corresponding new
flag in `augur tree` on or off. Sets the default value to `true`, since
the custom tree builder arguments we define in defaults are redundant
with previously hardcoded defaults in augur tree. This change should fix
a related segmentation fault on some systems, but it also allows users
to define more flexible tree builder arguments that conflict with the
defaults (e.g., bootstrap mode in IQ-TREE).

## Related issue(s)

Related to https://github.com/nextstrain/augur/issues/780

## Testing

 - [ ] Tested by CI
 - [ ] Tested with full Nextstrain builds

## Release checklist

### DIFF
--- a/defaults/parameters.yaml
+++ b/defaults/parameters.yaml
@@ -115,7 +115,8 @@ combine_sequences_for_subsampling:
   warn_about_duplicates: true
 
 tree:
-  tree-builder-args: "'-ninit 10 -n 4'"
+  tree-builder-args: "'-ninit 10 -n 4 -me 0.05'"
+  override_default_args: true
 
 # TreeTime settings
 refine:
--- a/docs/src/reference/change_log.md
+++ b/docs/src/reference/change_log.md
@@ -3,6 +3,10 @@
 As of April 2021, we use major version numbers (e.g. v2) to reflect backward incompatible changes to the workflow that likely require you to update your Nextstrain installation.
 We also use this change log to document new features that maintain backward compatibility, indicating these features by the date they were added.
 
+## v12 (? February 2022)
+
+ - Add a new workflow parameter in the `tree` config named [`override_default_args`](https://docs.nextstrain.org/projects/ncov/en/latest/reference/configuration.html#override_default_args) that enables overriding default tree builder arguments with the values defined by [`tree-builder-args`](https://docs.nextstrain.org/projects/ncov/en/latest/reference/configuration.html#tree-builder-args). This release of the workflow sets this parameter to `true` by default, which may be a breaking change for users who have modified the default tree builder arguments. Upgrade to [Augur 14.0.0](https://github.com/nextstrain/augur/blob/master/CHANGES.md#1400-8-february-2022) and run `augur tree -h` to see the default arguments for the tree builder you use. [See the original Augur pull request](https://github.com/nextstrain/augur/pull/839), for more details.
+
 ## New features since last version update
 
 - 11 February 2022: Add colors to default Auspice config for Nextclade quality control columns and a filter for overall Nextclade QC status. [PR #861](https://github.com/nextstrain/ncov/pull/861).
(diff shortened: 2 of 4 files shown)


jameshadfield:
```
Failed to compile.
ModuleNotFoundError: Module not found: Error: Can't resolve 'json-loader' in '/vercel/path0/packages/web'
```

This new version of auspice had a number of npm package upgrades, so perhaps that's related to the failed build here?

j23414 (approved):

joverlee521:
Bumping this PR since we've had people as about this in our discussion forum ([1](https://discussion.nextstrain.org/t/overriding-tree-builder-args/1051), [2](https://discussion.nextstrain.org/t/getting-node-support-values-in-the-ncov-workflow/1200)).

victorlin:
Assigning myself to fix the merge conflict with configuration.md, since I migrated that to [workflow-config-file.rst](https://github.com/nextstrain/ncov/blob/338e07e0c393f1288da3be61ebce6bca2d7d0ed4/docs/src/reference/workflow-config-file.rst).

huddlej:
Thank you, @joverlee521 for bumping and @victorlin for signing up to fix the merge conflict.

The main reason I hadn't merged this already was that this PR does make a potentially breaking change. It sets the override of tree args as the default behavior, but if people already have custom tree-builder args in their config files, Snakemake's deep merge of their config with the defaults will cause the default tree builder args to be overridden by the user's arguments.

As a specific example, the original defaults prior to this PR were:

```
tree:
  tree-builder-args: "'-ninit 10 -n 4'"
```

The default arguments for IQ-TREE in `augur tree` are:

```
-ninit 2 -n 2 -me 0.05
```

As a result, the config entry above produces an IQ-TREE argument like:

```
-ninit 2 -n 2 -me 0.05 -ninit 10 -n 4
```

The redundant arguments in that command can cause IQ-TREE to crash, for some operating systems and versions of IQ-TREE. We try to solve this problem with the new argument that allows overriding the default arguments.

The proposed defaults in this PR are:

```
tree:
  tree-builder-args: "'-ninit 10 -n 4 -me 0.05'"
  override_default_args: true
```

As a result, the config entry above produces an IQ-TREE argument like:

```
-ninit 10 -n 4 -me 0.05
``` 

It is possible that users have defined their own tree builder arguments, based on the workflow behavior prior to this PR. For example, a common pattern would be to tell IQ-TREE to collapse zero-length branches in the tree instead of forcing bifurcations:

```
tree:
  tree-builder-args: "'-czb'"
```

In the current workflow, this config produces the following IQ-TREE argument, where the first three arguments are `augur tree` defaults:

```
-ninit 2 -n 2 -me 0.05 -czb
```

In this PR's implementation, this config would get deep merged with our new defaults to look like so:

```
tree:
  tree-builder-args: "'-czb'"
  override_default_args: true
```

As a result of this deep merge, the resulting IQ-TREE command would be:

```
-czb
```

This command could produce quite different results than the command with the current workflow behavior above.

Given that our recommended tree builder arguments require overriding the default `augur tree` values, I think we need `override_default_args` to be true by default. It's frustrating that there isn't a clearer way to let users know about this change of behavior short of updating the change log and making a new release, though.

Another (less than ideal) solution would be to make `override_default_args` false by default, so everyone keeps the redundant IQ-TREE arguments we've had this whole time. If users experience a crash with IQ-TREE, then we can recommend that they enable the override in their config. This solution maintains the status quo, so it shouldn't break any workflows that aren't already broken, but it does put the onus of resolving a known bug on the users who experience the bug.

defaults/parameters.yaml
```diff
@@ -115,7 +115,8 @@ combine_sequences_for_subsampling:
   warn_about_duplicates: true
 
 tree:
-  tree-builder-args: "'-ninit 10 -n 4'"
+  tree-builder-args: "'-ninit 10 -n 4 -me 0.05'"
+  override_default_args: true
```
victorlin, line 119: The mix of `-` and `_` in parameter names right next to each other seems a bit off. Looking at the rest of this file, `_` is used for everything but `tree-builder-args`. It's probably beyond this PR to change that, so probably fine to use underscores for `override_default_args` here to follow broader convention.

PR #878, opened 2022-03-02
add: annotation updates from 2022-02-17 to 2022-02-28

j23414:

Update annotations up to Feb 28th. Let me know if I missed anything.

## Related issue(s)

Related to https://github.com/nextstrain/ncov-ingest/pull/288

### DIFF
--- a/defaults/color_ordering.tsv
+++ b/defaults/color_ordering.tsv
@@ -158,6 +158,7 @@ location	Jenin
 
 ### Syria
 # Aleppo
+location	A'Zaz
 location	Afrin
 location	Al Bab
 location	Ariha
@@ -473,6 +474,7 @@ location	Ariyalur
 location	Bhadradri
 location	Chegalpattu
 location	Chengalpattu
+location	Chengalpet
 location	Chennai
 location	Coimbathore
 location	Coimbatore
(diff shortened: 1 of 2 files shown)


MoiraZuber:
This looks almost good to go. The only changes I'd suggest are the following:

`country	French Guiana	4.0039882	-52.999998` -> Should be a French division
`country	Puerto Rico	18.19897685	-66.26152219418967` -> Should be a USA division

These can be a little difficult to figure out, since you'll have to check which divisions these countries contain. You could try to search the metadata somehow, but that's a little nasty with the file being so large now... What I usually do is a little trick of sorts. I add a rule such as

`*/French Guiana/*/*	Europe/France/French Guiana/`

which is my version of a sort of grep for all divisions below country French Guiana, which will need adjustment. Towards the top of the script output, you'll see which rules were applied, so there you'll see something like this:

`South America/French Guiana/Saint Laurent Du Maroni/	Europe/France/French Guiana/`

Now I only need to replace the rule from before by 

`South America/French Guiana/Saint Laurent Du Maroni/	Europe/France/French Guiana/Saint Laurent Du Maroni`

This way, I manage to adjust French Guiana to division level, while still preserving the "Saint Laurent Du Maroni" division as location. (All divisions that come with French Guiana would be lost if we used the first rule, that's why I prefer this method)

This is all quite cumbersome and I'd totally understand if you don't want to bother with it! So feel free to just leave those countries for me to curate anytime

`division	Upper Austria / Voecklabruck / Voecklamarkt	48.0022211	13.4850818`: I'll comment on this in Slack

PR #884, opened 2022-03-09
Support disabling the `--clock-filter-iqd` parameter for augur refine

### LINKED ISSUE #852: Allow users to disable clock filter for augur refine
## Current Behavior

When users try to build Omicron-specific trees against a background of non-Omicron strains, even the relaxed clock filter settings we now use can lead to strains being unexpectedly filtered out.

## Expected behavior

Users should be able to turn off the clock filter completely through the config file.

## Possible solution
Allow users to set the `clock_filter_iqd` parameter in `refine` section of the configuration file to `false` and then modify the corresponding argument in the refine rule to optionally build a command line argument for the clock filter instead of always using the clock filter argument.

jgadling:

Support disabling the `--clock-filter-iqd` parameter for `augur refine`. Since `--clock-filter-iqd 0` doesn't make sense, we can assume that setting the value to 0 (or even false) should disable it.

## Related issue(s)

Fixes #852
Related to #

## Testing

For a very simple test config:
```yaml
inputs:
  - name: "root_ref"
    metadata: "data/references_metadata.tsv"
    sequences: "data/references_sequences.fasta"
  - name: "example_data"
    metadata: "data/example_metadata_aus.tsv.xz"
    sequences: "data/example_sequences_aus.fasta.xz"

builds:
  test:
    subsampling_scheme: test_subsampling
    title: test tree

refine:
  keep_polytomies: True
  clock_filter_iqd: 0
  
subsampling:
  test_subsampling:
    test:
      max_sequences: 10
      exclude: "--exclude-where 'region=europe'"
```

```
~/ncov# grep clock_filter_iqd my_profiles/builds.yaml
  clock_filter_iqd: 5
~/ncov# rm results/test/tree.nwk && snakemake --printshellcmds auspice/ncov_test.json --profile /tmp/ncov/my_profiles/ --until refine 2>&1 | grep 'augur refine'
        augur refine             --tree results/test/tree_raw.nwk             --alignment results/test/filtered.fasta
             --metadata results/test/metadata_adjusted.tsv.xz             --output-tree results/test/tree.nwk
             --output-node-data results/test/branch_lengths.json             --root Wuhan/Hu-1/2019
             --timetree             --keep-polytomies             --clock-rate 0.0008
             --clock-std-dev 0.0004             --coalescent opt             --date-inference marginal 
            --divergence-unit mutations             --date-confidence
             --no-covariance             --clock-filter-iqd 5 2>&1 | tee logs/refine_test.txt

~/ncov# sed -i 's/filter_iqd:.*/filter_iqd: 0/' my_profiles/builds.yaml
~/ncov# grep clock_filter_iqd my_profiles/builds.yaml
  clock_filter_iqd: 0
~/ncov# rm results/test/tree.nwk && snakemake --printshellcmds auspice/ncov_test.json --profile /tmp/ncov/my_profiles/ --until refine 2>&1 | grep 'augur refine'
        augur refine             --tree results/test/tree_raw.nwk             --alignment results/test/filtered.fasta
             --metadata results/test/metadata_adjusted.tsv.xz             --output-tree results/test/tree.nwk
             --output-node-data results/test/branch_lengths.json             --root Wuhan/Hu-1/2019
             --timetree             --keep-polytomies             --clock-rate 0.0008
             --clock-std-dev 0.0004             --coalescent opt             --date-inference marginal
             --divergence-unit mutations             --date-confidence
             --no-covariance              2>&1 | tee logs/refine_test.txt
```

## Release checklist

### DIFF
--- a/docs/src/reference/change_log.md
+++ b/docs/src/reference/change_log.md
@@ -6,6 +6,7 @@ We also use this change log to document new features that maintain backward comp
 ## New features since last version update
 
 - 11 February 2022: Add colors to default Auspice config for Nextclade quality control columns and a filter for overall Nextclade QC status. [PR #861](https://github.com/nextstrain/ncov/pull/861).
+- 8 Mar 2022: Support disabling clock filters in the refine step by setting `clock_filter_iqd: 0` in the `refine` section. [PR #884](https://github.com/nextstrain/ncov/pull/884), [Issue #852](https://github.com/nextstrain/ncov/issues/852).
 
 ## v11 (3 February 2022)
 
--- a/docs/src/reference/configuration.md
+++ b/docs/src/reference/configuration.md
@@ -449,7 +449,7 @@ Valid attributes for list entries in `inputs` are provided below.
 
 ### clock_filter_iqd
 * type: integer
-* description: Remove tips that deviate more than this number of interquartile ranges from the root-to-tip by time regression.
+* description: Remove tips that deviate more than this number of interquartile ranges from the root-to-tip by time regression. Disable clock filtering by specifying `0`
 * default: `4`
 
 ### keep_polytomies
--- a/docs/src/reference/configuration.md
+++ b/docs/src/reference/configuration.md
@@ -450,7 +450,7 @@ Valid attributes for list entries in `inputs` are provided below.
 ### clock_filter_iqd
 * type: integer
 * description: Remove tips that deviate more than this number of interquartile ranges from the root-to-tip by time regression. Disable clock filtering by specifying `0`
-* default: `4`
+* default: `8`
 
 ### keep_polytomies
 * type: boolean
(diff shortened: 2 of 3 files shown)


danrlu:
Test runs from this branch. Spot the difference: 
image

huddlej (approved):
Thank you for putting this together, @jgadling, and I'm sorry for the late review!

The Snakemake implementation looks like exactly what we need to toggle the clock filter and @danrlu's screenshots confirm that this works as expected. Thank you also for updating the default value in the docs!

I was originally concerned about using `0` to disable the clock filter instead of a YAML value like [`null` that indicates the lack of a value](https://yaml.org/type/null.html), but it turns out that [augur refine internally treats the clock filter IQD value as false-y when it is 0 and skips the filter](https://github.com/nextstrain/augur/blob/945b49af50cc72c4343673491531ae630db40040/augur/refine.py#L40), so this implementation would work the same with an `augur refine` command on the terminal as in the workflow.

PR #886, opened 2022-03-13
Fix subsampling_scheme min-date and max-date parameters bug.

abattenhouse:
The min-date and max-date subsampling_scheme parameters do not work properly. This 4-line addition to the _get_setting fundtion in the main_workflow.smk file fixes the problem.

## Related issue(s)
None that I know of.

## Testing
I ran into the problem when I specified a min-date parameter in one of my custom Nextstrain builds. After I made this proposed change it works fine.

## Release checklist
Should not introduce backward incompatible changes since this functionality did not work before.

### DIFF
--- a/workflow/snakemake_rules/main_workflow.smk
+++ b/workflow/snakemake_rules/main_workflow.smk
@@ -191,6 +191,10 @@ def _get_specific_subsampling_setting(setting, optional=False):
                     value = f"--exclude-ambiguous-dates-by {value}"
                 elif setting == 'group_by':
                     value = f"--group-by {value}"
+                elif setting == 'min_date':        # FIXED by Anna Battenhouse
+                    value = f"--min-date {value}"
+                elif setting == 'max_date': 
+                    value = f"--max-date {value}"  
         elif value is not None:
             # If is 'seq_per_group' or 'max_sequences' build subsampling setting,
             # need to return the 'argument' for augur


huddlej (commented):
Hi @abattenhouse, thank you for submitting this! The min and max date parameters in ncov's filter logic don't behave like you'd expect due to an implementation choice we made at the beginning of the pandemic that we were then unable to change without breaking everyone's workflows.

The min/max date filters should work when defined like `min_date: --min-date 2020-01-01`, for example, instead of (like you'd expect) `min_date: 2020-01-01`. [The SARS-CoV-2 workflow configuration guide provides some examples and more details](https://docs.nextstrain.org/projects/ncov/en/latest/reference/workflow-config-file.html#min-date). You can also find some real world examples in our Nextstrain builds ([like this one](https://github.com/nextstrain/ncov/blob/3e55b799293042db67a371e2c709ba5c5c33ec34/nextstrain_profiles/nextstrain-gisaid/builds.yaml#L164)).

Can you try running your builds with the original workflow (so not the code in this branch) and using the format from the examples linked to above for your min/max dates and let us know if that fixes the issue?

victorlin:
Closing due to inactivity. Please re-open if you're still experiencing any issues.

I've created a separate issue to discuss changing the syntax to be more intuitive: #1144

PR #894, opened 2022-03-22
docs: convert to reST, re-organize and update contents across all pages

### LINKED ISSUE #808: Outline revised documentation based on integration with docs.nextstrain.org
### Context
Now that the "SARS-CoV-2 workflow" docs are implemented in RTD and integrated as a subproject of the main docs at docs.nextstrain.org, we can revise these SC2-specific docs to take advantage of their new context and avoid duplicating materials.

### Description
We need an outline of what the new documentation organization should be given that we expect users to complete the main Nextstrain tutorials as prerequisites for the SC2 tutorial. We also need to consider the primary use cases that people have now (compared to when the docs were written).

[See related meeting notes for more context](https://docs.google.com/document/d/1HxnvQew-2y3ZIB4LWPlHle3nJGQiDTS0uP1uxoAtAuI/edit).

### Tasks

- [ ] #863
- [ ] Reorganize guides
- [ ] Reorganize reference material

### LINKED ISSUE #863: Rewrite and reorganize ncov tutorial

victorlin:
[Preview](https://nextstrain--894.org.readthedocs.build/projects/ncov/en/894/tutorial/intro.html)

- Convert pages to reST
- Rewrite tutorial
- Update contents of existing pages to reflect current workflow
- Re-organize content

### Related issues

- Resolves #863
- Resolves #808

### Post-merge tasks

- [ ] sync `redirects.yml` using [readthedocs-cli](https://github.com/nextstrain/readthedocs-cli)


docs/src/tutorial/index.rst
```diff
@@ -0,0 +1,16 @@
+********
+Tutorial
+********
+
+These tutorials will guide you to create a :term:`Nextstrain build <docs.nextstrain.org:build>` for genomic surveillance.
```
huddlej, line 5: I just realized that we don't provide any preface for readers to explain the overall goals of the tutorial, a list of [learning outcomes](https://teaching.cornell.edu/teaching-resources/designing-your-course/setting-learning-outcomes), or the scope of the tutorial (e.g., we're not going to cover everything in genomic epidemiology). The main index page provides the overall goals with:

> This template and tutorial will walk you through the process of running a basic phylogenetic analysis on SARS-CoV-2 data. We’ve created these resources with the goal of enabling Departments of Public Health to start using Nextstrain to understand their SARS-CoV-2 genomic data within 1-2 hours.

This text could live in an "Introduction" page of the tutorial (instead of the top-level docs where it feels out of place now). Then we could expand on these goals to include learning outcomes and details about the scope like:

```rst
At the end of this tutorial, you will be able to:

- create phylogenetic trees of SARS-CoV-2 genomes from different sources including GISAID and Nextstrain-curated GenBank data
- visualize the resulting trees in Auspice
- define subsampling logic for your own genomic epidemiological analysis
- (optional?) configure parameters of the SARS-CoV-2 workflow that produces the phylogenetic trees
- (optional?) configure how your phylogenetic trees and associated metadata appear in Auspice
 
After completing this tutorial, you may wish to `learn more about genomic epidemiology <link to Alli's gen epi book or some other resources?>`_ or `review all possible options to configure your SARS-CoV-2 analyses with Nextstrain <../reference/configuration.html>`_.
```

docs/src/tutorial/example-data.rst
```diff
@@ -0,0 +1,89 @@
...
+
+   .. code:: text
+
+      cd ncov
+
+3. Download the example tutorial repository into a new directory ``ncov-tutorial/``:
+
+   .. code:: text
+
+      $ git clone https://github.com/nextstrain/ncov-tutorial
```
huddlej, line 33: Minor point: could we drop the leading shell prompt `$` and example output to match the other `git clone` command in the setup page? It's handy when one can copy and paste an entire code block in the terminal without knowing to select the relevant subset.

docs/src/tutorial/custom-data.rst
```diff
@@ -0,0 +1,162 @@
...
+       metadata: data/custom.metadata.tsv
+       sequences: data/custom.sequences.fasta
+
+   refine:
+     root: "Wuhan-Hu-1/2019"
+
+   builds:
+     custom-build:
+       title: "Build with custom data and example data"
+       auspice_config: ncov-tutorial/auspice-config-custom-data.json
```
victorlin, line 106: I referenced the auspice config JSON file as a path relative to `ncov/` so it is prefixed with `ncov-tutorial/`. This works fine as long as readers follow the directions in the previous tutorial, but it's a bit weird to think about when looking at the contents of `ncov-tutorial/` directly without context that the workflow is being run one directory above.

victorlin: Placing `ncov/` within `ncov-tutorial/` could make this more straightforward, but this is a new approach which needs some tinkering to get it working (see [Slack thread](https://bedfordlab.slack.com/archives/C02Q4Q5S85A/p1646334194519489?thread_ts=1646334170.382399&cid=C02Q4Q5S85A)).

docs/src/tutorial/example-data.rst
```diff
@@ -0,0 +1,89 @@
...
+      [...more output...]
+
+Run the workflow
+----------------
+
+From within the ``ncov/`` directory, run the ``ncov`` workflow using a configuration file provided in the tutorial directory:
+
+.. code:: text
+
+   nextstrain build . --cores 4 --configfile ncov-tutorial/example-data.yaml
```
huddlej, line 44: This failed for me at the tree building step with the following error:

```
ERROR: Shell exited 2 when running: iqtree -nt 4 -s results/default-build/masked_filtered-delim.fasta -m GTR -ninit 2 -n 2 -me 0.05 -ninit 10 -n 4 > results/default-build/masked_filtered-delim.iqtree.log
Command output was:
  ERROR: You have specified more threads than CPU cores available
```

I've only allocated 3 of my 4 total CPUs to Docker containers which IQ-TREE detects here. If I run the above command without specifying the number of cores, the CLI will use all cores available to the container:

```
$ nextstrain build . --configfile ncov-tutorial/example-data.yaml
Building DAG of jobs...
Using shell: /bin/bash
Provided cores: 3
[...snip...]
```

And the workflow completes as expected. I think we can safely drop the `--cores` argument and corresponding reference below.

victorlin: Oh that's neat! I didn't know that feature of Nextstrain CLI. Agree on dropping `--cores` for all 3 tutorials then.

docs/src/tutorial/example-data.rst
```diff
@@ -0,0 +1,89 @@
...
+
+         inputs:
+           - name: reference_data
+             metadata: https://data.nextstrain.org/files/ncov/open/reference/metadata.tsv.xz
+             sequences: https://data.nextstrain.org/files/ncov/open/reference/sequences.fasta.xz
+
+         refine:
+           root: "Wuhan-Hu-1/2019"
+
+      This provides the workflow with one input named ``reference_data``. The metadata and sequences files are downloaded directly from the associated URLs.
```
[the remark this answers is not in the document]
victorlin: I wasn't aware of this page. These instructions in the genomic surveillance tutorial can also be replaced then:

https://github.com/nextstrain/ncov/blob/d14620a193f189a75131fec7ee3a0fd19819cc41/docs/src/tutorial/genomic-surveillance.rst?plain=1#L132-L136

docs/src/tutorial/custom-data.rst
```diff
@@ -0,0 +1,162 @@
...
+   .. code:: text
+
+      cd ncov
+
+Curate data from GISAID
+-----------------------
+
+.. warning::
+
+   These steps require a GISAID account. `Register <https://www.gisaid.org/registration/register/>`__ if you do not have an account yet. However, registration may take a few days. Follow `alternative data preparation methods <../guides/data-prep.html>`__ if you wish to continue this tutorial in the meantime.
```
huddlej, line 28: What if we moved this information to the first step of the "Setup" page (just before installing Nextstrain)? Requesting a GISAID account is such a major part of SARS-CoV-2 analyses, we could prioritize this information and avoid surprising users with this prerequisite in the middle of the tutorial.

docs/src/tutorial/example-data.rst
```diff
@@ -0,0 +1,89 @@
...
+From within the ``ncov/`` directory, run the ``ncov`` workflow using a configuration file provided in the tutorial directory:
+
+.. code:: text
+
+   nextstrain build . --cores 4 --configfile ncov-tutorial/example-data.yaml
+
+Break down the command
+~~~~~~~~~~~~~~~~~~~~~~
+
+The workflow can take several minutes to run. While it is running, you can learn about the parts of this command:
```
huddlej, line 49: Just for our own reference, the example build ran in 7 minutes with 3 CPUs. The refine step ran for 4 min 40 seconds of that total time.

docs/src/tutorial/example-data.rst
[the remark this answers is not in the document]
huddlej: As I work through this, it would be nice to link to the configuration guide here, too.

docs/src/tutorial/example-data.rst
```diff
@@ -0,0 +1,89 @@
...
+
+Run this command to start the :term:`docs.nextstrain.org:Auspice` server, providing ``auspice/`` as the directory containing output dataset files:
+
+.. code:: text
+
+   nextstrain view auspice/
+
+Navigate to ``http://127.0.0.1:4000/ncov/default-build``. The resulting :term:`docs.nextstrain.org:dataset` should show a phylogeny of ~200 sequences:
+
+.. figure:: ../images/dataset-example-data.png
```
huddlej, line 84: This screenshot shows the tree colored by clade, but the default color-by for this build is by country, so my tree from running the above commands looks like this instead:

image

We should update the default Auspice config JSON to color by `clade_membership` by default, to make it match our own Nextstrain builds and this screenshot.

huddlej: 993040395ade8796766e894813cb26f2329f01e1 should fix this.

docs/src/tutorial/custom-data.rst
```diff
@@ -0,0 +1,162 @@
...
+
+   .. image:: ../images/gisaid-augur-pipeline-download.png
+      :width: 400
+      :alt: GISAID EpiCoV Search
+
+   .. note::
+
+      You may see different download options, but it is fine as long as **Input for the Augur pipeline** is available.
+
+8. Extract by opening the downloaded ``.tar`` file in your file explorer. It contains two files: one ending with ``.metadata.tsv`` and another with ``.sequences.fasta``.
```
huddlej, line 72: What if we ask users to download the tar file into the `ncov/data/` directory and extract it there? Maybe I'm too used to command line file operations, but opening two separate Finder windows to copy from `~/Downloads/` to `~/ncov/data/` felt like too many extra steps.

docs/src/tutorial/custom-data.rst
```diff
@@ -0,0 +1,162 @@
...
+10. Rename the files as ``custom.metadata.tsv`` and ``custom.sequences.fasta``.
+
+Run the workflow
+----------------
+
+From within the ``ncov/`` directory, run the ``ncov`` workflow using a pre-written ``--configfile``:
+
+.. code:: text
+
+   nextstrain build . --cores 4 --configfile ncov-tutorial/custom-data.yaml
```
huddlej, line 83: This ran as expected (in 6 min with 3 CPUs), but the workflow printed a bunch of annoying warnings about the lack of a subsampling scheme for the build like this one:

```
WARNING: No valid subsampling scheme is defined for build 'custom-build'. Skipping subsampling and using all available samples.
```

This is a warning that I added to the workflow back when I changed the default behavior to include all samples (last spring or summer). We could either drop the warning completely or update the builds entry to explicitly specify the subsampling scheme like this:

```
builds:
  custom-build:
    subsampling_scheme: all
```

with the explanation that we will define a better scheme in the next part of the tutorial.

I sort of lean toward removing the warning, but maybe there is pedagogical value in exposing the subsampling scheme logic earlier?

docs/src/tutorial/custom-data.rst
huddlej: Agreed that it is a little weird without context, although the way it's written here is great until we figure out another repo/configs organization.

If the `ncov-tutorial/` directory were instead named `my-builds/` or `my-analyses/`, the resulting path is a little less strange to me (as in, "these are my custom files to control the workflow" instead of "the tutorial lives inside of the workflow, for some reason").

```yaml
builds:
   custom-build:
     title: "Build with custom data and example data"
     auspice_config: my-analyses/auspice-config-custom-data.json
```

This kind of naming convention is part of what I was thinking with treating the `ncov-tutorial` repo like a template, so folks can reference their own set of files as they like.

docs/src/tutorial/custom-data.rst
```diff
@@ -0,0 +1,162 @@
...
+
+Navigate to ``http://127.0.0.1:4000/ncov/custom-build``. The resulting :term:`docs.nextstrain.org:dataset` should have similar phylogeny to the previous dataset, with additional sequences:
+
+.. figure:: ../images/dataset-custom-data-highlighted.png
+   :alt: Phylogenetic tree from the "custom data" tutorial as visualized in Auspice
+
+
+1. The custom dataset name ``custom-build`` can be seen in the dataset selector, as well as the dataset URL.
+2. The custom dataset title can be seen at the top of the page.
+3. The custom coloring is used by default. You can see which sequences are from the custom data added in this tutorial.
```
huddlej, line 158: This worked beautifully. Really nice touch, @victorlin!

docs/src/tutorial/genomic-surveillance.rst
```diff
@@ -0,0 +1,158 @@
...
+   .. code:: text
+
+      cd ncov
+
+Curate data from GISAID
+-----------------------
+
+.. warning::
+
+   These steps require a GISAID account. `Register <https://www.gisaid.org/registration/register/>`__ if you do not have an account yet. However, registration may take a few days. Follow `alternative data preparation methods <../guides/data-prep.html>`__ if you wish to continue this tutorial in the meantime.
```
huddlej, line 28: Same comment as the "custom data" tutorial: this should probably be in the Setup or at least listed as a prerequisite.

docs/src/tutorial/genomic-surveillance.rst
```diff
@@ -0,0 +1,158 @@
...
+   .. image:: ../images/gisaid-epicov-search.png
+      :width: 400
+      :alt: GISAID EpiCoV Search
+
+2. Select a custom set of records. As an example, this guide will look for all records that pass the following criteria:
+
+   1. From Idaho
+   2. Collected between ``2022-02-15`` and ``2022-03-15``
+   3. Has a complete genome
+   4. Has an exact collection date
```
huddlej, line 43: We need a step 5 here that tells users to click the checkbox in the top left to select all records (across all search result pages).

docs/src/tutorial/genomic-surveillance.rst
```diff
@@ -0,0 +1,158 @@
...
+1. Navigate to `GISAID <https://www.gisaid.org/>`__, **Login**, and go to **EpiCoV** > **Search**.
+
+   .. image:: ../images/gisaid-epicov-search.png
+      :width: 400
+      :alt: GISAID EpiCoV Search
+
+2. Select a custom set of records. As an example, this guide will look for all records that pass the following criteria:
+
+   1. From Idaho
+   2. Collected between ``2022-02-15`` and ``2022-03-15``
```
huddlej, line 41: I'm not sure how to best handle this, but these dates will quickly get out of date with the reference data we use in the rest of the tutorial and we're not guaranteed to always get ~200 records with this search when future data submitters add more records for these dates.

What if we asked users to just set the minimum collection date to the last month or two, filtering until they get ~200 records?

docs/src/tutorial/genomic-surveillance.rst
```diff
@@ -0,0 +1,158 @@
...
+   2. ``usa_context``
+
+      - This selects sequences from the ``background_data`` input, up to a maximum of 1000 sequences.
+      - Sequences are subsampled evenly across all combinations of ``division``, ``year``, ``month``, with sequences genetically similar to ``custom_sample`` prioritized over other sequences.
+
+   3. ``global_context``
+
+      - This selects sequences from the ``reference_data`` input.
+
+This configuration file works without modifications, but the following steps guide you to customize names and titles, if appropriate:
```
huddlej, line 126: What if we moved this line right after the section header above (`Edit the workflow configuration file`), so users know early on that they can run these commands without modifying a YAML if they want.

docs/src/tutorial/genomic-surveillance.rst
```diff
@@ -0,0 +1,158 @@
...
+4. Extract by opening the downloaded ``.tar`` file in your file explorer. It contains two files: one ending with ``.metadata.tsv`` and another with ``.sequences.fasta``.
+5.  Move these extracted files to the ``data/`` directory in the ``ncov/`` directory.
+6.  Rename the files as appropriate, for example ``idaho.metadata.tsv`` and ``idaho.sequences.fasta``.
+
+Edit the workflow configuration file
+------------------------------------
+
+This time, you will make some modifications to the pre-written configuration file ``genomic-surveillance.yaml`` before running the workflow. The contents are copied below (comments excluded):
+
+.. code-block:: yaml
```
huddlej, line 65: There is a lot going on in this configuration now! I like how you've broken everything down below, @victorlin, but I wonder if we could break the configuration itself into smaller sections to tackle individually. For example, we could show the next `inputs` section, describe the additional `background_data` input and new Idaho inputs and also how we're able to use the pre-aligned sequences with the `aligned` entry point.

Then we could show the `builds` section and call out the new `subsampling_scheme` key and other bits the user may want to change.

Then we could show the `subsampling` section and go into the detail about what each component does.

docs/src/tutorial/genomic-surveillance.rst
```diff
@@ -0,0 +1,158 @@
...
+         max_sequences: 5000
+       usa_context:
+         query: --query "(background_data != 'yes') & (country == 'USA')"
+         max_sequences: 1000
+         group_by: division year month
+         priorities:
+           type: proximity
+           focus: custom_sample
+       global_context:
+         query: --query "(reference_data == 'yes')"
```
huddlej, line 100: Mostly a note-to-self to edit in these queries `ncov-tutorial` repo and in this branch so we're querying contextual data from both background and reference data. We'll want the `usa_context` query to be `query: --query "(custom_data != 'yes') & (country == 'USA')"` and the `global_context` query to be `query: --query "(custom_data != 'yes')"`.

The current queries will end up excluding the background data from all of the subsampling rules.

docs/src/tutorial/genomic-surveillance.rst
```diff
@@ -0,0 +1,158 @@
...
+   3. Replace ``north-america`` in each of the ``metadata`` and ``aligned`` URLs with your choice of regional dataset.
+
+Run the workflow
+----------------
+
+From within the ``ncov/`` directory, run the ``ncov`` workflow:
+
+.. code:: text
+
+   nextstrain build . --cores 4 --configfile ncov-tutorial/genomic-surveillance.yaml
```
huddlej, line 145: This command had a subtle issue where the combined metadata (`results/combined_metadata.tsv.xz`) and sequences (`results/combined_sequences_for_subsampling.fasta.xz`) already existed from the "custom data" run, so the workflow didn't re-run the steps to combine these files and include the newly defined inputs. As a result the workflow crashed when trying to run an `augur filter` query that referred to a nonexistent `background_data` column.

To fix the issue, I had to tell the workflow to rebuild everything with the `--forceall` flag:

```
nextstrain build . --forceall  --configfile ncov-tutorial/genomic-surveillance.yaml
```

This seems like a Snakemake bug or a bug in our workflow; the new "background_data" entry and updated contents of the "custom_data" files should trigger a rebuild of files that depend on them.

docs/src/tutorial/custom-data.rst
huddlej: After sleeping on this, I've changed my mind and slightly prefer exposing the `subsampling_scheme: all` in this "custom data" tutorial. This will preemptively answer the question that users would probably ask themselves about "which data are going into this analysis?". We can say something like "For now, we will set a subsampling scheme that uses all input data; in the next section of the tutorial, we will show how to customize this scheme for genomic surveillance".

victorlin: Is `all` a predefined subsampling scheme? I think also good to add this to [the configuration guide](https://docs.nextstrain.org/projects/ncov/en/latest/reference/configuration.html).

docs/src/tutorial/genomic-surveillance.rst
huddlej: Still looking into the DAG issue with Snakemake, but noting that this build ran in 42 min on 3 CPUs with the config in ncov-tutorial@cb2f69fe413a6b979687b52b897f6fdd2a8c4da9.

docs/src/tutorial/example-data.rst
victorlin: Which version of Nextstrain CLI and Snakemake are you using? With `nextstrain-cli=3.2.1` and `snakemake=7.3.1` I get an error when dropping `--cores`:

> Error: you need to specify the maximum number of CPU cores to be used at the same time. If you want to use N cores, say --cores N or -cN. For all cores on your system (be sure that this is appropriate) use --cores all. For no parallelization use --cores 1 or -c1.

So, the better solution might be to use `--cores all`.

docs/src/tutorial/custom-data.rst
victorlin: I want to keep this tutorial cross-platform. If we provide shell-specific commands such as `curl`/`wget`, or even line breaks using `\`, it won't work for Windows PowerShell users.

For the download step, we could summarize the intent and nest the drag-drop instructions, like:

> 8. Extract the files to the `data/` directory in the `ncov/` directory:
>     1. Extract by opening the downloaded ``.tar`` file in your file explorer. It contains two files: one ending with ``.metadata.tsv`` and another with ``.sequences.fasta``.
>     2. Move these extracted files to the `data/` directory in the `ncov/` directory.
> 9. Rename the files as ``custom.metadata.tsv`` and ``custom.sequences.fasta``.

But that seems a bit verbose to me.

docs/src/tutorial/genomic-surveillance.rst
victorlin: The selection here is mostly for example purposes, I think we should encourage the user to pick something relevant to their own interest given:

1. Best results with ~200 sequences
2. Max 5,000 sequences
3. More sequences = longer run time

I tried to make this point in the `.. note::` below the screenshot, but can change the step to say something like:

> Select a custom set of records. We recommend a set of ~200 sequences as a good balance between a meaningful amount of data and run time. Note that more records will take the workflow longer to run, and the GISAID web interface has a maximum of 5,000 records per download.
> 
> As an example, this guide will look for all records that pass the following criteria:
> ...

docs/src/tutorial/genomic-surveillance.rst
victorlin: I like this idea! I've just been writing and trying to fill the gaps, not thinking about organization much. I'll make a change to address this.

docs/src/tutorial/custom-data.rst
huddlej: Oh yeah, I wasn't thinking to provide platform-specific instructions, but just to say "Download into `ncov/data/`" as the default behavior, just to make life easier across platforms instead of downloading wherever (e.g., not into `~/Downloads/` by default).

docs/src/tutorial/custom-data.rst
huddlej: Yeah, [it's one of the few predefined schemes were provide](https://github.com/nextstrain/ncov/blob/a5a7384ff2783b5f59677bb9c50f02a1c3c04feb/defaults/parameters.yaml#L200-L204).

docs/src/tutorial/example-data.rst
huddlej: Good point. I'm using the nextstrain.cli 3.2.0 with the Docker image `nextstrain/base:build-20220315T184512Z` which uses Snakemake 5.10.0 (where number of cores isn't required). Dropping `--cores` will break things if we ever [update Snakemake some day](https://github.com/nextstrain/docker-base/pull/23/files#r832562522) or for folks running with the native mode.

Using `--cores all` works for me with the CLI + Docker though, so that's a great solution.

docs/src/tutorial/example-data.rst
huddlej: I think it makes sense for the default behavior for all users to match the defaults for our live SARS-CoV-2 builds, but I've just pinged Emma on Slack about this, since she originally changed from clade to country back in June 2020 (1c2dcdb20).

docs/src/tutorial/genomic-surveillance.rst
huddlej: Probably another point worth discussing with the broader group. There could be two different sets of users:

1. Those who want to run the tutorial from start to end by just running commands. These users may just want to get a sense of how these types of analyses can be configured through this worked example before they embark on their own analyses.
2. Those who want to use the tutorial as a guide to setup their own genomic epi analysis.

Thinking back to [the documentation system](https://documentation.divio.com/tutorials/) that we've previously used to guide our docs, a tutorial like this should focus on users in the first group. They are learning by doing first. Then, they can use what they've learned to implement their own genomic epi analysis.

I think this suggests sticking with a specific example like Idaho, instead of asking users to curate a new dataset. The only issue could be that the fixed time window of Idaho data will not change while the reference data we use will continue to move forward in time. Maybe this is a non-issue since the reference data will always including genomes from the history of the pandemic.

That said, I'm also not sure we _need_ the reference data to keep updating through time. I thought the original goal of the reference data was to provide representative genomes for this tutorial. The fact that we generate these daily with the other ncov builds is an implementation detail that has nothing to do with this original goal.

docs/src/tutorial/example-data.rst
victorlin: Pulled this out into a separate PR which can get merged first: #897

docs/src/tutorial/example-data.rst
victorlin: Done: 666ba683

docs/src/tutorial/example-data.rst
victorlin: Done: 3464dd5d

docs/src/tutorial/example-data.rst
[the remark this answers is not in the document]
victorlin: Done: 9132b2da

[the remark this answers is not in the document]
victorlin: @huddlej maybe not for the first tutorial, I'd think a link to the full configuration guide is more appropriate in the final genomic surveillance tutorial.

docs/src/tutorial/custom-data.rst
victorlin: I have my browser set to automatically download to `~/Downloads/` without prompt, and I imagine some readers will have the same as well.

With ad130748 I've re-ordered and re-worded a bit so the steps should accommodate all behaviors.

docs/src/tutorial/custom-data.rst
```diff
@@ -0,0 +1,162 @@
...
+             }
+           ],
+           "display_defaults": {
+             "color_by": "custom_data"
+           }
+         }
+
+      This JSON does two things:
+
+      1. Create a new coloring ``custom_data`` which reflects a special metadata column generated by the ncov workflow. Each data input produces a new final metadata column with categorical values ``yes`` or ``no`` representing whether the sequence was from the input.
```
victorlin, line 134: Originally from @huddlej: https://github.com/nextstrain/ncov/commit/672d2055794ac803e11f957d8b5833fda2a87521#r69298892

> Each data input produces a new final metadata column with categorical values...

The workflow only adds these input-specific columns when there are multiple inputs during [the `combine_input_metadata` rule](https://github.com/nextstrain/ncov/blob/464041debc66a86a33d3c1c697026a1adeb1d326/workflow/snakemake_rules/main_workflow.smk#L35-L58). We may need to clarify this behavior here, so users aren't confused when they look for a similar column with a single input (not that such a column would be useful!).

docs/src/tutorial/genomic-surveillance.rst
victorlin: No longer necessary with a709eccd, based on https://github.com/nextstrain/ncov/pull/894#discussion_r832713912

docs/src/tutorial/genomic-surveillance.rst
victorlin: Ran [updated config](https://github.com/nextstrain/ncov-tutorial/blob/57a8d91021396ab72d077a5041b28a06d6abfa0e/genomic-surveillance.yaml) in 27 min on an M1 Mac using 8 cores (native runtime).

docs/src/tutorial/genomic-surveillance.rst
victorlin: Fixed: 4e5894f6 and https://github.com/nextstrain/ncov-tutorial/commit/57a8d91021396ab72d077a5041b28a06d6abfa0e

docs/src/tutorial/genomic-surveillance.rst
victorlin: Done: 460fc340

docs/src/tutorial/example-data.rst
```diff
@@ -0,0 +1,87 @@
...
+
+   .. code:: text
+
+      cd ncov
+
+3. Download the example tutorial repository into a new directory ``ncov-tutorial/``:
+
+   .. code:: text
+
+      git clone https://github.com/nextstrain/ncov-tutorial
```
jameshadfield, line 33: Can we add `ncov-tutorial` to the `.gitignore` in the ncov repo? This would avoid the following situation

```
$ git status
On branch tutorial
Your branch is up to date with 'origin/tutorial'.

Untracked files:
  (use "git add <file>..." to include in what will be committed)
        ncov-tutorial/

nothing added to commit but untracked files present (use "git add" to track)
```

docs/src/tutorial/example-data.rst
```diff
@@ -0,0 +1,87 @@
+Run the workflow using example data
+===================================
+
+In this tutorial, you will run the workflow using example reference data.
```
jameshadfield, line 4: I think we could expand this a bit to give a better understanding for _why_ we are asking users to do this. Something like:

> The aim of this first tutorial is to introduce our SARS-CoV-2 workflow.
To do this, we will run the workflow using a small set of reference data which we provide.
This tutorial leads on to subsequent tutorials where we will walk through more complex scenarios.

docs/src/tutorial/example-data.rst
```diff
@@ -0,0 +1,87 @@
...
+
+         inputs:
+           - name: reference_data
+             metadata: https://data.nextstrain.org/files/ncov/open/reference/metadata.tsv.xz
+             sequences: https://data.nextstrain.org/files/ncov/open/reference/sequences.fasta.xz
+
+         refine:
+           root: "Wuhan-Hu-1/2019"
+
+      This provides the workflow with one input named ``reference_data``, which is a small dataset maintained by the Nextstrain team. The metadata and sequences files are downloaded directly from the associated URLs. `See the complete list of SARS-CoV-2 datasets we provide through data.nextstrain.org <https://docs.nextstrain.org/projects/ncov/en/latest/reference/remote_inputs.html>`_.
```
jameshadfield, line 67: This is the first time users have seen a so I'd expand a touch more. E.g. replace "is a YAML file with contents" with "is a YAML file which parameterises the workflow to produce the analysis we want". And also add a link to the reference page (https://nextstrain--894.org.readthedocs.build/projects/ncov/en/894/reference/configuration.html). 

(I've got thoughts about that reference page, but I'll add them later.)

docs/src/tutorial/example-data.rst
```diff
@@ -0,0 +1,87 @@
...
+      The ``refine`` entry specifies the root sequence for the example GenBank data.
+
+Visualize the results
+---------------------
+
+Run this command to start the :term:`docs.nextstrain.org:Auspice` server, providing ``auspice/`` as the directory containing output dataset files:
+
+.. code:: text
+
+   nextstrain view auspice/
```
jameshadfield, line 78: Similar to a previous comment, I had to replace this with

```
nextstrain view --native  auspice/
```

docs/src/tutorial/example-data.rst
```diff
@@ -0,0 +1,87 @@
...
+             metadata: https://data.nextstrain.org/files/ncov/open/reference/metadata.tsv.xz
+             sequences: https://data.nextstrain.org/files/ncov/open/reference/sequences.fasta.xz
+
+         refine:
+           root: "Wuhan-Hu-1/2019"
+
+      This provides the workflow with one input named ``reference_data``, which is a small dataset maintained by the Nextstrain team. The metadata and sequences files are downloaded directly from the associated URLs. `See the complete list of SARS-CoV-2 datasets we provide through data.nextstrain.org <https://docs.nextstrain.org/projects/ncov/en/latest/reference/remote_inputs.html>`_.
+
+      The ``refine`` entry specifies the root sequence for the example GenBank data.
+
```
jameshadfield, line 70: I'd add a line or section here explaining what the _output_ of this command is going to be: a dataset JSON named `auspice/ncov_default-build.json` which we will visualise in the next step. This allows (a) us to explain why it has that name and (b) when we visualise the data in the next section it's not a surprise where the JSON file came from

jameshadfield: Might also explain that a bunch of intermediate files are created in `results/`

jameshadfield (commented):
Nice work @victorlin! I've read through and tested the first tutorial. Will try to cover the rest tomorrow.

docs/src/tutorial/example-data.rst
jameshadfield: Another thought -- explicitly add a "builds" section to the config and explain this. This is a touch more complexity up-front, but in my opinion it reduces the amount of magic or unexpected things happening. Along the lines of explicit is better than implicit. That's my 2 cents. @huddlej what do you think about this?

docs/src/tutorial/custom-data.rst
```diff
@@ -0,0 +1,158 @@
...
+     custom-build:
+       title: "Build with custom data and example data"
+       auspice_config: ncov-tutorial/auspice-config-custom-data.json
+
+This is the same as the previous file, with some additions:
+
+1. A second input for the custom data, referencing the metadata and sequences files downloaded from GISAID.
+2. A ``builds`` section that defines one output :term:`docs.nextstrain.org:dataset` using:
+
+   1. A custom name ``custom-build``
```
jameshadfield, line 107: I suggest a bit more verbosity here:

1. A custom name ``custom-build``, which will be used to create the dataset filename, in this case `auspice/ncov_custom-build.json`
2. A custom title ``Build with custom data and example data``, which will be shown when you visualise the dataset in Auspice.
3. An Auspice config file,  `ncov-tutorial/auspice-config-custom-data.json`, which defines various parameters for how the data should be visualised in Auspice. It has the following contents:

docs/src/tutorial/genomic-surveillance.rst
```diff
@@ -0,0 +1,177 @@
+Run the workflow using a genomic surveillance configuration
+===========================================================
+
+In this tutorial, you will replace the custom data with meaningful focal data and use a configuration commonly found in genomic surveillance applications.
```
jameshadfield, line 4: Again (i'm a broken record!) I suggest we frame this in terms of aims:

> In the previous tutorial we showed how to analyse a small set of GISAID ("custom") data in the context of a small set of reference data. For genomic surveillance applications, we often have a set of data specific to our question of interest, for instance a set of sequences from a particular geographic area, which is referred to as the focal set. We want to analyse this focal set in a global context, and since there are millions of global sequences we want to subset these based on genomic proximity to our focal set.

docs/src/tutorial/example-data.rst
```diff
@@ -0,0 +1,87 @@
...
+      git clone https://github.com/nextstrain/ncov-tutorial
+
+Run the workflow
+----------------
+
+From within the ``ncov/`` directory, run the ``ncov`` workflow using a configuration file provided in the tutorial directory:
+
+.. code:: text
+
+   nextstrain build . --cores all --configfile ncov-tutorial/example-data.yaml
```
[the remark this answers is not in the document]
victorlin: This sounds like a CLI issue, but I haven't seen it myself... what's your `nextstrain --version`?

docs/src/tutorial/example-data.rst
```diff
@@ -0,0 +1,87 @@
...
+
+.. code:: text
+
+   nextstrain view auspice/
+
+Navigate to ``http://127.0.0.1:4000/ncov/default-build``. The resulting :term:`docs.nextstrain.org:dataset` should show a phylogeny of ~200 sequences:
+
+.. figure:: ../images/dataset-example-data.png
+   :alt: Phylogenetic tree from the "example data" tutorial as visualized in Auspice
+
```
jameshadfield, line 90: Seems super obvious to us, but should add a comment here about how to close the server (does ctrl+c still work on windows?)

docs/src/tutorial/genomic-surveillance.rst
```diff
@@ -0,0 +1,177 @@
...
+1. The file paths in the second input are changed to ``idaho.metadata.tsv`` and ``idaho.sequences.fasta``.
+2. There is an additional input ``background_data`` for a regional North America dataset built by the Nextstrain team, for additional context.
+
+builds
+******
+
+The output dataset is renamed ``idaho``, representative of the new custom data in the second input.
+
+1. The title is updated.
+2. There is a new entry ``subsampling_scheme: idaho_scheme``. This is described in the following section.
```
jameshadfield, line 125: Perhaps a bit more on the _why_. E.g. "Without this the dataset would use all the provided data, which in this case is many thousand sequences."

docs/src/tutorial/running.md
```diff
@@ -3,21 +3,21 @@
 >This section focuses on how to running the basic example build to give you a chance to practice and get a sense of how things work. The next section covers customizing and configuring your own build.
```
jameshadfield, line 3: This page feels really out of place - the layout is different from the preceeding 3 tutorials, it uses config files in `./my_profiles` not `./ncov-tutorial`, and doesn't really cover any new ground. I'd suggest removing the tutorial part, but keeping the "Troubleshoot common issues" part.

docs/src/tutorial/custom-data.rst
```diff
@@ -0,0 +1,158 @@
...
+   :alt: Phylogenetic tree from the "custom data" tutorial as visualized in Auspice
+
+
+1. The custom dataset name ``custom-build`` can be seen in the dataset selector, as well as the dataset URL.
+2. The custom dataset title can be seen at the top of the page.
+3. The custom coloring is used by default. You can see which sequences are from the custom data added in this tutorial.
+
+   .. note::
+
+      You may not see all 10 custom sequences - some can be filtered out due to quality checks built into the ncov workflow.
```
jameshadfield, line 157: (Adding here so the discussion can be threaded).

We have a single page in the Guide section named "Preparing your data" which acts a bit like a tutorial as well as explaining the formatting of the files as well as explaining how to download from GISAID. I would suggest we:

1. Add a section at the end of this tutorial ("Run the workflow using custom data") saying that you can easily replace the GISAID sample here with your own private sequences, but to do this the metadata & fasta must be in a specific format (which here GISAID manages for us). This can then point to a page labelled "Preparing your data"
2. Modify the page "Preparing your data" to only consider the FASTA / TSV formatting. All the GISAID information, as well as how to run the analysis, is covered in the tutorials so it shouldn't be repeated in that guide.

We only have one guide... will we add more? If not, let's shift that page to the "Reference material" section. cc @huddlej for thoughts.

docs/src/tutorial/example-data.rst
victorlin: Done: 0868b4d0

docs/src/tutorial/example-data.rst
victorlin: Done: 2b5fda17

docs/src/tutorial/example-data.rst
victorlin: Done: 35331f4f

docs/src/tutorial/example-data.rst
victorlin: Done: fbdb4d58

docs/src/tutorial/genomic-surveillance.rst
victorlin: Done: 563598af

docs/src/tutorial/custom-data.rst
victorlin: Fixed with 03e713ea, https://github.com/nextstrain/ncov-tutorial/commit/aef9f19.

Created a separate issue to document predefined subsampling schemes: #900.

docs/src/tutorial/custom-data.rst
victorlin: Moved to prerequisite sections: 563598af

I went for prerequisite section over "Setup" page since the "example data" tutorial can still be completed without a GISAID account.

docs/src/tutorial/index.rst
victorlin: Done: 534c9c05

jameshadfield (changes requested):
Awesome @victorlin! Thanks for taking the lead on this. I think the three tutorials are great. In addition to the in-line comments, I have bigger picture comments here about removing / streamlining the other pages. Perhaps this is a separate PR?

**remove my_profiles**
The three tutorials plus the "preparing your data" page cover pretty much all the contents that `./my_profiles` did. How do we feel about
* removing `./reference/multiple_inputs` and corresponding `./data/` files.
* shift any useful config YAMLs into the `ncov-tutorial` page as examples. Perhaps this needs a corresponding page in the "reference material" along the lines of "more complicated configuration file examples"

**reorder the pages in "reference material"**

* Orientation / overview pages should come first

docs/src/tutorial/example-data.rst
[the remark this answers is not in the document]
jameshadfield:
```
$ nextstrain --version
nextstrain.cli 3.2.1
```

Installed into a fresh conda environment yesterday following the instructions in our docs

docs/src/tutorial/custom-data.rst
```diff
@@ -0,0 +1,158 @@
+Run the workflow using custom data
+==================================
+
+In this tutorial, you will run the workflow using custom focal data in addition to the example reference data. The reference data will serve as background context for the new data.
```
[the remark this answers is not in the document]
victorlin: Done: a008ed21

I changed analyse -> analyze for consistency. Don't mean to put your spelling dialect down! (my spell checker disagrees)

docs/src/tutorial/custom-data.rst
victorlin: Done: da49a0ae

docs/src/tutorial/genomic-surveillance.rst
victorlin: Thanks! I wouldn't have been able to write it that nicely so I appreciate you being a broken record here.

Done: a008ed21

docs/src/tutorial/example-data.rst
victorlin: Done: 7092a32e

docs/src/tutorial/genomic-surveillance.rst
victorlin: Done: 95d9f372

I reworded it a bit:

> Without this, the output dataset would use all the provided data, which in this case is thousands of sequences that are often disproportionally representative of the underlying population.

docs/src/tutorial/running.md
victorlin: This was retained in 8c7a4115ee215a4a3a54260b1e3dc040be61c216 but I have yet to update the contents.

docs/src/tutorial/example-data.rst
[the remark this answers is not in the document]
victorlin: I tested on a fresh Ubuntu VM without docker installed, and can't reproduce this. @tsibley any idea?

<details>
<summary>For the record, this is my output:</summary>

```
$ nextstrain check-setup
nextstrain-cli is up to date!

Testing your setup…

# docker is not supported
✘ no: docker is installed
✘ no: docker run works
? unknown: containers have access to >2 GiB of memory
✔ yes: image is new enough for this CLI version

# native is not supported
✘ no: snakemake is installed
✘ no: augur is installed
✘ no: auspice is installed

# aws-batch is not supported
✘ no: job description "nextstrain-job" exists
✘ no: job queue "nextstrain-job-queue" exists
✘ no: S3 bucket "nextstrain-jobs" exists

No good.  No support for any Nextstrain environment.
```

</details>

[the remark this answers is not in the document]
jameshadfield: What about if you install augur, auspice etc via conda so that there is support for the native runtime? Does the command then work without needing `--native`?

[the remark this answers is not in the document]
tsibley: @jameshadfield Looks like maybe you forgot `--set-default` when you ran:

```
$ nextstrain check-setup
...
All good!  Supported Nextstrain environments: native
```

The default runtime is Docker if none is set, so that's why your subsequent `nextstrain build` used it.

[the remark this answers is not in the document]
victorlin: Ohh ok. Somehow I missed that the error was from `nextstrain build`.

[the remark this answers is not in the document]
jameshadfield: https://github.com/nextstrain/cli/issues/164

docs/src/tutorial/custom-data.rst
victorlin: Done: ac84d6eb

docs/src/tutorial/custom-data.rst
```diff
@@ -0,0 +1,158 @@
...
+      :width: 200
+      :alt: GISAID login
+
+3. In the top left navigation bar, select **EpiCoV** then **Search**.
+
+   .. image:: ../images/gisaid-epicov-search.png
+      :width: 400
+      :alt: GISAID EpiCoV Search
+
+4. Select the first 10 sequences.
```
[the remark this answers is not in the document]
victorlin: Done: efcd0357

docs/src/tutorial/custom-data.rst
victorlin: Done: 7e68a150

victorlin: Relatedly, do you think we should instruct users to create their own repo from of the `ncov-tutorial` template repo in the first example or at the end?

I think the end would be good enough since we aren't instructing any file changes in these tutorials.

docs/src/tutorial/example-data.rst
victorlin: Updated to ignore `my-analyses/` according to new instructions (see https://github.com/nextstrain/ncov/pull/894#discussion_r840122613)

docs/src/tutorial/running.md
victorlin: Done: 5a45a987

docs/src/reference/orientation-files.rst
```diff
@@ -0,0 +1,56 @@
...
+-  ``defaults/auspice_config.json``: default visualization configuration file. Override these settings using ``auspice_config``. `See the customization guide for visualizations <customizing-visualization.md>`__.
+-  ``defaults/include.txt``: list of strain names to *include* during subsampling and filtering (one strain name per line)
+-  ``defaults/exclude.txt``: list of strain names to *exclude* during subsampling and filtering (one strain name per line)
+
+Custom configuration files
+--------------------------
+
+These will be stored in ``my-analyses/``.
+
+- ``my-analyses/.../builds.yaml``: workflow configuration file used by ``--configfile`` where you can define and parameterize the builds you want to run. `See the customization guide <customizing-analysis.md>`__.
```
victorlin, line 22: Concern about removal of `my_profiles/` with the current replacement of `my-analyses/`:

A decent amount of reference material relies on `my_profiles/` or an equivalent being present. This is fine when `my_profiles/` is guaranteed to be in the repo when cloned.

However, see the above modified snippet of [orientation-files](https://docs.nextstrain.org/projects/ncov/en/latest/analysis/orientation-files.html#workflow-configuration-files-we-might-want-to-customize) which assumes that the reader has `my-analyses/`.

This reference material would only make sense for readers who have gone through the tutorial, or at least the step where they clone `ncov-tutorial` into `my-analyses/`.

Do we want to push hard for users to go through that cloning step? If so, then maybe it would be best to put that step in the setup instructions. If not, I can't think of a good alternative here.

docs/src/reference/customizing-visualization.rst
```diff
@@ -0,0 +1,111 @@
+Customizing your Auspice visualization
+======================================
+
+Just as we can specify a build-specific analysis options in the ``builds.yaml`` file, we can also specify build-specific visualization options in this directory.
+
+Looking at the ``builds.yaml`` file, the last few lines are:
```
victorlin, line 6: I'd like to refer to this file as something other than `builds.yaml` since:

1. This is an artifact of `my_profiles/` where each profile had a `builds.yaml` file.
2. We promote usage of more descriptive names like [example-data.yaml](https://github.com/nextstrain/ncov-tutorial/blob/main/example-data.yaml).

Ideas:

- **_workflow configuration file_** (not sure, since there's so many other files that "configure" the workflow)
- **_config file_** since it's given to Snakemake's `--configfile` parameter and this is the term used in [Snakemake docs](https://snakemake.readthedocs.io/en/stable/snakefiles/configuration.html)
- Add more of an explanation on the `orientation-files` page. Maybe this file can be the only "config file" while all other files are "customization" files

docs/src/tutorial/custom-data.rst
victorlin: Started this: c03dcd5fc1e3d006d1e20c760f2ac320ee84d284

docs/src/reference/customizing-visualization.rst
victorlin: Note: it's currently referenced by description in the [example-data tutorial](https://github.com/nextstrain/ncov/blob/abef0ec9805a40b353c0547011bf332ab156470d/docs/src/tutorial/example-data.rst?plain=1#L57), and as `--configfile` in the [custom-data tutorial](https://github.com/nextstrain/ncov/blob/abef0ec9805a40b353c0547011bf332ab156470d/docs/src/tutorial/custom-data.rst?plain=1#L86).

docs/src/tutorial/custom-data.rst
victorlin: Will link to docs on predefined subsampling schemes once #905 is merged.

docs/src/reference/customizing-visualization.rst
tsibley: I don't have the context of how this fits into other language used in this doc project, but to me this is a "workflow configuration file" which I might more casually refer to as a "workflow config file" or just "config file" in the context. All the same thing to me. (In places where the filename mattered/wasn't arbitrary, I would want to use that to be specific, at least once to establish context.)

jameshadfield: > I'd like to refer to this file as something other than builds.yaml 

👍 How about "Just as we can specify build-specific configuration via `snakemake --configfile <your-config.yaml>`, we can..."

docs/src/reference/orientation-files.rst
jameshadfield: Yeah, removing a directory that people may expect is always a bit complicated. To smooth the path, how about:

1. A `.. hint::` in the orientation-files explaining that `my_profiles` is no longer
2. Leave a `./my_profiles/README.md` with the contents of (1). We can make a reminder to delete this in 2 (?) months.

I don't know best practices around template repos etc etc, but I did find it a bit strange that we clone `ncov-tutorial` into `my-analyses`. An alternative would be to keep that folder as the canonical tutorials and ask people to add their config files to `./my-analyses` (perhaps initialising it with a `./my-analyses/README.md`).

docs/src/tutorial/custom-data.rst
jameshadfield: Looks great!

I rearranged the sidebar in https://github.com/nextstrain/ncov/pull/906 which makes this section more prominant. 

I also think we need to add this as a hint / note in the "Run the workflow using custom data" page, as it seems natural to want to go from that tutorial to using your own (non-GISAID) data.

docs/src/reference/customizing-visualization.rst
victorlin: I think **_config file_** should be sufficient, as long as we distinguish it from other files used to configure the workflow e.g. `auspice_config.json`

docs/src/reference/orientation-files.rst
victorlin: Relevant discussion here: https://github.com/nextstrain/ncov/pull/894#discussion_r832683570

I think the point of ncov-tutorial as a template repo is to make it easier to build upon the tutorial files, rather than copy/pasting from another directory or starting from scratch.

Two potential options:

### 1. With template repo:

- In setup steps: direct users to make their own `my-ncov-analyses` repo from the `ncov-tutorial` template repo.
    - Not `my-analyses` since I imagine something like `https://github.com/victorlin/my-analyses` is too generic.
- Reference `my-ncov-analyses/` in all the these docs.
    - I don't think this is the right approach, it assumes people have gone through the tutorial setup. But I can't think of a better option.

### 2. Without template repo (@jameshadfield's suggestion above):

- Provide a file `ncov/my-analyses/README.md`.
    - Primary purpose is to have `my-analyses/` already available, similar to current `my_profiles/`.
- In example-data tutorial: `git clone ncov-tutorial` without renaming.
- At the end of genomic-surveillance tutorial: copy `ncov-tutorial/genomic_surveillance.yaml` to `my-analyses/genomic_surveillance.yaml`, then edit there.
    - Add note to initialize `my-analyses/` as a git repo to enable version control (recommended).
- Reference `my-analyses/` in all the these docs.

---

I still feel like this would be simpler [if `ncov/` was used as a standalone directory](https://bedfordlab.slack.com/archives/C02Q4Q5S85A/p1646334170382399). I'm playing with @huddlej's suggestion on [Modularization](https://snakemake.readthedocs.io/en/stable/snakefiles/modularization.html) to see if this experience can be improved.

docs/src/reference/customizing-visualization.rst
victorlin: Trying to make this more reference-able in a new glossary page ([preview](https://nextstrain--894.org.readthedocs.build/projects/ncov/en/894/reference/glossary.html), work in progress).

docs/src/reference/orientation-files.rst
victorlin: [Latest approach from Slack](https://bedfordlab.slack.com/archives/C02Q4Q5S85A/p1649794496745729?thread_ts=1649180310.784879&cid=C02Q4Q5S85A):

1. For the tutorial, we tell readers to `git clone ncov-tutorial` without renaming the directory.
2. Add a guide to create a personal repo from `ncov-tutorial` template.
    - We recommend naming the repo `my-ncov-analyses/` since it will be automatically excluded by `ncov/.gitignore`.
    - `git clone https://github.com/victorlin/my-ncov-analyses` will create an "analysis directory" `ncov/my-ncov-analyses/`.
3. For orientation-files which currently has lots of references to `my_profiles/`, we add a new section describing the "analysis directory" and move current `my_profiles/` references to this new section.

1 and 2 are done: 3644c310afbe6cec582fc6981778ee0abad3fc8f...334a0d087e3bc1ffde0784539610e7ba474a2af0

docs/src/tutorial/genomic-surveillance.rst
victorlin: The "Customize this example" section mentioned above has been moved to a new ["Next steps" page](https://nextstrain--894.org.readthedocs.build/projects/ncov/en/894/tutorial/next-steps.html). 334a0d087e3bc1ffde0784539610e7ba474a2af0

docs/src/reference/orientation-files.rst
victorlin: I think this should be good with [the new `tutorial/next-steps` page](https://nextstrain--894.org.readthedocs.build/projects/ncov/en/894/tutorial/next-steps.html) + [a rewrite of the `reference/files` page](https://nextstrain--894.org.readthedocs.build/projects/ncov/en/894/reference/files.html#analysis-directory) linking to the former:

reference-files-analysis-directory

docs/src/tutorial/custom-data.rst
victorlin: The experience here has been updated based on https://github.com/nextstrain/ncov/pull/894#discussion_r853401870

docs/src/tutorial/custom-data.rst
victorlin: Done: https://github.com/nextstrain/ncov/blob/65084a99402e30ef0401501b7d0bfd391d9c9bfa/docs/src/tutorial/custom-data.rst?plain=1#L83-L85

docs/src/reference/multiple_inputs.md
```diff
@@ -41,7 +41,7 @@ my_profiles/example_multiple_inputs/my_auspice_config.json
```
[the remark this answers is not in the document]
jameshadfield: Don't think we want to keep anything here. It was written at a time when most analysis was n=1 inputs and we were introducing the concept of multiple input. Now we introduce that almost from the get-go.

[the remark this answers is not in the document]
victorlin: Cool, that makes things easier. Deleted.

docs/src/reference/orientation-files.rst
victorlin: > 1. A `.. hint::` in the orientation-files explaining that `my_profiles` is no longer
> 2. Leave a `./my_profiles/README.md` with the contents of (1). We can make a reminder to delete this in 2 (?) months.

Added these as suggested by @jameshadfield.

docs/src/tutorial/next-steps.rst
```diff
@@ -0,0 +1,73 @@
...
+   Now that you have created your own analysis directory, this must be modified, e.g.
+
+   .. code:: yaml
+
+      auspice_config: my-ncov-analyses/auspice-config-custom-data.json
+
+Additional resources
+====================
+
+- Learn more about genomic epidemiology. (TODO: link to Alli's gen epi book or some other resources)
```
victorlin, line 71: @jameshadfield @huddlej Anyone have the link for this?

j23414: Maybe this?
> Add links to [CZI GenEpi](https://czgenepi.org/resources) and [CDC video](https://www.cdc.gov/amd/training/covid-19-gen-epi-toolkit.html) tutorials to our SARS-CoV-2 docs

huddlej: Those links are a great start! I don't know if Alli's book is out yet, but I will ask her.

victorlin: This was shared at office hours! https://alliblk.github.io/genepi-book/intro.html

docs/src/reference/auspice-config-guide.rst
```diff
@@ -0,0 +1,106 @@
...
+
+.. contents:: Table of Contents
+   :local:
+
+Custom color schemes
+--------------------
+
+If you'd like to specify a custom color scale, you can add a ``colors.tsv`` file, where each line is a tab-delimited list of a metadata column name; a metadata value; and a corresponding hex code.
+
+The first few lines of the example file look like this:
```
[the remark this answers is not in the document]
jameshadfield: s/the example file/such an example file could/

docs/src/tutorial/example-data.rst
```diff
@@ -0,0 +1,99 @@
...
+Prerequisites
+-------------
+
+1. :doc:`setup`. These instructions will install all of the software you need to complete this tutorial and others.
+
+Setup
+-----
+
+1. Activate the ``nextstrain`` conda environment:
```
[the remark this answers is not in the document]
jameshadfield: I'm a bit lost -- is this needed because the canonical docker-runtime flow would be to `git clone` etc from within the docker container (i.e. within `nextstrain shell .`)? If this is so, then I'd advocate for "Create a conda environment named nextstrain and install the Nextstrain CLI:" to also install `git`.

docs/src/tutorial/videos.rst
```diff
@@ -2,7 +2,7 @@
 Video tutorial walkthrough
```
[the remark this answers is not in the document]
victorlin: Added redirects.yml: 2b581c456a920562230e342542bc79743a966182...7e834099372fd1477bca6bc738bad704c8761a88

Added TODO in PR description to sync this file once changes are merged.

docs/src/tutorial/setup.rst
```diff
@@ -66,27 +66,3 @@ Alternately, download a specific version of the workflow that you know works for
```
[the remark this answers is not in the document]
victorlin: Done.

docs/src/tutorial/example-data.rst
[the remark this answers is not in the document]
victorlin: `git` is both included in Native install directions and [installed in the Docker image](https://github.com/nextstrain/docker-base/blob/5e61a58d0e0b6b49cd99204b449eb82204c1ed7e/Dockerfile#L20), so I think that's not an issue.

I don't think this should block merge of the PR, so I've written up in more detail here: nextstrain/docs.nextstrain.org#108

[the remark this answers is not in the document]
victorlin: I got confused when writing this. Closed the issue and resolving this conversation.

docs/src/tutorial/next-steps.rst
victorlin: Done.

docs/src/tutorial/genomic-surveillance.rst
victorlin: @huddlej I am running this using Docker runtime on my computer (very slowly due to nextstrain/docker-base#35), but I am at the `refine` step and can see that this already ran successfully:

> python3 scripts/combine_metadata.py --metadata results/sanitized_metadata_reference_data.tsv.xz results/sanitized_metadata_custom_data.tsv.xz results/sanitized_metadata_background_data.tsv.xz --origins reference_data custom_data background_data --output results/combined_metadata.tsv.xz 2>&1 | tee logs/combine_input_metadata.txt

So, I don't think I am able to reproduce this 😕

victorlin:
Merging with @jameshadfield's [approval on Slack](https://bedfordlab.slack.com/archives/C02Q4Q5S85A/p1651192343194159?thread_ts=1650501286.584649&cid=C02Q4Q5S85A).

PR #895, opened 2022-03-22
Swap open builds starting points from S3 bucket to data.nextstrain.org

trvrb:

This change accomplishes a couple things:
1. The `data.nextstrain.org/files` URLs are what are advertised in our docs and in the page footer. This makes it more obvious how the build is working.
2. The data.nextstrain.org URLs are backed by CloudFront and should be more performant than direct S3 bucket downloads.

## Testing

I've tested locally.

## Release checklist

I don't believe this change necessitates anything in the release checklist.

### DIFF
--- a/nextstrain_profiles/nextstrain-open/builds.yaml
+++ b/nextstrain_profiles/nextstrain-open/builds.yaml
@@ -18,8 +18,8 @@ upload:
 # as we re-align everything after subsampling.
 inputs:
   - name: open
-    metadata: "s3://nextstrain-data/files/ncov/open/metadata.tsv.gz"
-    aligned: "s3://nextstrain-data/files/ncov/open/sequences.fasta.xz"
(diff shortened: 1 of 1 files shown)


victorlin (approved):
Seems fine to me. Though, if we are running this on AWS Batch, I'd assume S3 downloads to be more performant than data.nextstrain.org.

tsibley:
I expect fetching through Cloudfront for our builds to be both slower and more expensive. I think this should be reverted. As [our docs](https://docs.nextstrain.org/projects/ncov/en/latest/reference/remote_inputs.html) say:

> If you’re running workflows on AWS or GCP compute that fetch this data, please use the S3 or GS URLs, respectively, for cheaper (for us) and faster (for you) data transfers. Otherwise, please use the https://data.nextstrain.org URLs.

Issue #789, opened 2021-11-23
Build names do not allow "." characters


huddlej:
**Current Behavior**
Users who want to create a build for a specific Pango lineage with a build name like "B.1.1.7" will get an obscure validation error from Snakemake because the build config schema (and the workflow's wildcard constraints) do not support "." characters.

**Expected behavior**
Users should definitely be able to define builds named like "B.1.1.7" especially since build outputs are stored in a directory.

**Solutions**

- Update the config YAML schema in the workflow to allow "." characters
- Update the wildcard constraints in the Snakefile to match the config YAML schema

**Related issues/PRs**
- #522 
- #524

abattenhouse:
FYI, as an outside, non-Nextstrain team user performing custom Nextstrain builds, I wasted more than 4 hours one night figuring out where my early build error came from -- it was because I had a number in my build name. Incredibly frustrating!

PR #896, opened 2022-03-22
doc: convert config guide to reST, document name restrictions for `builds:`

victorlin:

This PR adds a warning message to the **builds** section:

> image

This is documented internally with [workflow config schema][1] and [various wildcard constraints][2].

[1]: https://github.com/nextstrain/ncov/blob/464041debc66a86a33d3c1c697026a1adeb1d326/workflow/schemas/config.schema.yaml
[2]: https://github.com/nextstrain/ncov/blob/464041debc66a86a33d3c1c697026a1adeb1d326/Snakefile#L132-L138

To add the warning block, I first had to convert to reStructuredText.

## Related issue(s)

- #789
- #482

## Testing

_N/A_

## Release checklist

_N/A_


jameshadfield:
Thanks @victorlin! Looks good. Perhaps explicitly call out the fact that they cannot have a full stop?

Also, the schema should be updated to avoid the build name ending with `_measurements` or `_root-sequence`, both of which are reserved for sidecar files. But this can be another PR / issue.

John & I talked about re-organising this file, and I'll try to do this in a separate PR on top of the tutorial branch.

docs/src/reference/configuration.rst
```diff
@@ -0,0 +1,1033 @@
...
+-  type: object
+-  description: Named builds to produce by the workflow from the given inputs. Builds are indexed by name and include any number of build attributes that can be used to control subsampling, Auspice configuration, and more.
+
+.. warning::
+
+   Build names currently only allow alpha characters, underscores, and hyphens (``A-Z``, ``a-z``, ``_``, ``-``), but must not contain ``tip-frequencies`` as it is a special string used for Nextstrain builds.
+
+   Note that these are not allowed:
+
+   - Periods (``.``)
```
victorlin, line 71: _from https://github.com/nextstrain/ncov/pull/896#issuecomment-1080134719_

> @jameshadfield: Perhaps explicitly call out the fact that they cannot have a full stop?

Added here along with some other explicit examples.

docs/src/reference/configuration.rst
```diff
@@ -0,0 +1,1033 @@
...
+
+builds
+------
+
+-  type: object
+-  description: Named builds to produce by the workflow from the given inputs. Builds are indexed by name and include any number of build attributes that can be used to control subsampling, Auspice configuration, and more.
+
+.. warning::
+
+   Build names currently only allow alpha characters, underscores, and hyphens (``A-Z``, ``a-z``, ``_``, ``-``), but must not contain ``tip-frequencies`` as it is a special string used for Nextstrain builds.
```
victorlin, line 67: I also noticed that just `tip-frequencies` passes the regex, is this fine? Or should it be matching for `_tip-frequencies`?

jameshadfield: We should err on the side of caution here and make sure the build name doesn't end with [any of the 5 suffixes we filter out in the nextstrain server](https://github.com/nextstrain/nextstrain.org/blob/d3a1fb7185ca3348c4121535484145db4619b714/src/utils/index.js#L83-L88): 

```js

  // All JSON files which aren't a sidecar file with a known suffix are assumed to
  // be v2+ JSONs (aka "unified" JSONs)
  const sidecarSuffixes = ["meta", "tree", "root-sequence", "seq", "tip-frequencies", "measurements"];
  const datasets = jsonFiles
    .filter((filename) => !sidecarSuffixes.some((suffix) => filename.endsWith(`_${suffix}.json`)))
```

victorlin: Added this to issue #900

PR #903, opened 2022-04-01
Revert "Swap open builds starting points from S3 bucket to data.nextstrain.org"

tsibley:
This reverts commit 18cb66a75aa65b0302f3e0904c980c8f681846a3.¹

I expect fetching through CloudFront in our builds to be both slower and more
expensive for us.  As our docs² say:

> If you’re running workflows on AWS or GCP compute that fetch this data,
> please use the S3 or GS URLs, respectively, for cheaper (for us) and
> faster (for you) data transfers. Otherwise, please use the
> https://data.nextstrain.org URLs.

¹ https://github.com/nextstrain/ncov/pull/895
² https://docs.nextstrain.org/projects/ncov/en/latest/reference/remote_inputs.html

### DIFF
--- a/nextstrain_profiles/nextstrain-open/builds.yaml
+++ b/nextstrain_profiles/nextstrain-open/builds.yaml
@@ -18,8 +18,8 @@ upload:
 # as we re-align everything after subsampling.
 inputs:
   - name: open
-    metadata: "https://data.nextstrain.org/files/ncov/open/metadata.tsv.gz"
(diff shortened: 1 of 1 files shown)


victorlin (approved):

tsibley:
Thanks for the eyes on this @victorlin. I'm also going to wait for @trvrb to weigh in before merging since it was his change this reverts.

trvrb:
Got it. This makes sense. Thanks Tom.

trvrb (approved):

PR #904, opened 2022-04-04
docs: Add copy button to code blocks

victorlin:

Using https://pypi.org/project/sphinx-copybutton/
    
Also https://github.com/nextstrain/docs.nextstrain.org/commit/cc2e2a85c66e36591d83b34b7195b1eff18b3ce7

## Related issue(s)

_N/A_

## Testing

_N/A_

## Release checklist

_N/A_

### DIFF
--- a/docs/conda.yml
+++ b/docs/conda.yml
@@ -12,3 +12,4 @@ dependencies:
     - sphinx-markdown-tables
     - sphinx-argparse
     - sphinx-tabs
+    - sphinx-copybutton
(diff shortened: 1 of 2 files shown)


tsibley (approved):
As an aside, it'd be nice if we could bundle this with https://github.com/nextstrain/sphinx-theme somehow as it feels more like "basic functionality" than other extensions (which often require explicit usage in a document). Not sure if you can bundle other extensions into extensions/themes, or if the theme could essentially vendor the JS/CSS/hooks from sphinx-copybutton.

tsibley:
https://github.com/nextstrain/sphinx-theme/issues/24

PR #907, opened 2022-04-07
Organise config reference docs

jameshadfield:
See commit messages for individual changes, but I generally find this reorganisation (especially the TOC) to be clearer. While searching for a key is still how many will use this page, I imagine a lot of people are skim-reading this page to get an overview of the available configuration. 

[RTD Preview](https://nextstrain--907.org.readthedocs.build/projects/ncov/en/907/reference/configuration.html)

When reading the actual content, I find `<h3>` and `<h4>` headings to be indistinguishable, which makes it hard to know whether a title is a top-level config key or a key of another config object. For instance, can you tell that "warn_about_duplicates" is a parameter of the parent config option "combine_sequences_for_subsampling"? I'll try to tackle this, but I'm wary of making broad changes to our docs theme. 

image

P.S. @victorlin I wasn't sure what the following directive was doing so may have accidentally messed something up...
```
.. contents::
   :local:
```


docs/src/reference/configuration.rst
```diff
@@ -3,62 +3,89 @@ Configuration parameters for Nextstrain SARS-CoV-2 workflow
...
+       metadata: data/other_metadata.tsv.xz
+       aligned: data/other_aligned.fasta.xz
 
--  description: List of input names (i.e., “origins”) for which intermediate files should be uploaded to S3 by the ``upload`` rule.
--  examples
+Valid attributes for list entries in ``inputs`` are provided below.
 
-   -  ``["gisaid"]``
+.. contents::
+   :local:
```
victorlin, line 39: > @jameshadfield: I wasn't sure what the following directive was doing

This adds a "table of contents" for individual sub-headings. Since we are only exposing the top-level YAML keys using `:depth: 2` in the main TOC, deeper keys like `builds.title` don't have a link up top. These smaller scoped TOCs provide these links:

builds-toc

victorlin (approved):
Awesome, this page is looking a lot better than it was a couple weeks ago!

docs/src/reference/configuration.rst
```diff
@@ -141,410 +168,513 @@ title
...
+
+
+combine_sequences_for_subsampling
+---------------------------------
+
+-  type: object
+-  description: Configuration of logic to combine sequences from multiple input files into a single file for subsampling.
+
+warn_about_duplicates
+~~~~~~~~~~~~~~~~~~~~~
```
victorlin, line 648: > @jameshadfield For instance, can you tell that "warn_about_duplicates" is a parameter of the parent config option "combine_sequences_for_subsampling"?

Yeah, this is something that can be improved. The individual `.. contents::` helps a bit for `<h3>`s which have a lot of `<h4>`s, such as `build`. I didn't add one here since there is only one `<h4>` but maybe a description + `.. contents::` here would help.

docs/src/reference/configuration.rst
```diff
@@ -141,410 +168,513 @@ title
...
+   hCoV-19/USA/CZB-3456/2021   -3.1
 
-min_date
-~~~~~~~~
 
--  type: float or string
--  description: Minimum collection date for strains to include in the analysis used by ``augur filter --min-date``. Dates can be numeric floating point values (e.g., ``2019.74``) or ISO 8601-style strings (e.g., ``2019-10-01``).
--  default: ``2019.74``
+Cross-Workflow configuration
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
victorlin, line 349: Could you add a short description here (and other `<h2>`s) like you did for **Main Workflow configuration**? I'm not sure what "Cross-Workflow" means.

victorlin (changes requested):

docs/src/reference/configuration.rst
jameshadfield: Ahh, those mini-TOCs are really good. I'll go through and see if I can add them anywhere.

I tried playing around with encoding the structure rather than just the variable name, but I'm not sure what syntax to use (I presume you'd follow how type-hints are done in python?) and I'm not sure readers would find it clearer.

image

docs/src/reference/configuration.rst
jameshadfield: For sure! [64ab92c](https://github.com/nextstrain/ncov/pull/907/commits/64ab92c9733c13f6e4020334609c278447f2cf91)

jameshadfield:
[New RTD preview](https://nextstrain--907.org.readthedocs.build/projects/ncov/en/907/reference/configuration.html) which adds custom styles to this page (and this page only):

image

docs/src/reference/configuration.rst
jameshadfield: Added styles in [743712d](https://github.com/nextstrain/ncov/pull/907/commits/743712d379b02907db382710d89c56db9b060776) which I think fix this :)

docs/src/reference/configuration.rst
```diff
@@ -141,455 +177,638 @@ title
...
-   -  ``min_tips``
-   -  ``min_frequency``
-   -  ``max_frequency``
+   -  ``mask_from_beginning``
+   -  ``mask_from_end``
+   -  ``mask_sites``
+- Valid attributes:
 
 .. contents::
    :local:
```
victorlin, line 719: _Observable prior to PR, thus non-blocking and can be addressed separately_

This renders a bit redundant:

logistic_growth parameters

I suggest dropping the "required" bullet point and putting required-ness inline with the attributes, such as like so for `refine`:

refine parameters proposal

docs/src/reference/configuration.rst
```diff
@@ -141,455 +177,638 @@ title
...
+
+   hCoV-19/USA/CZB-1234/2021   8.2
+   hCoV-19/USA/CZB-2345/2021   0
+   hCoV-19/USA/CZB-3456/2021   -3.1
+
+
+Cross-Workflow configuration
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+These parameters are general high-level parameters which may affect multiple Snakemake rules, or modify which rules are run.
```
victorlin, line 360: Two concerns:

1. I can understand the need to distinguish between **_Main Workflow Configuration_** and **_Cross-Workflow Configuration_**, but from a technical view **_Main Workflow Configuration_** also falls under the definition here.
2. With current descriptions, it isn't clear that `subsampling` is also a per-rule configuration.

Suggestion:

> **Primary configuration**
> 
> These parameters define the main inputs and outputs of the workflow, as well as the commonly used `subsampling` rule. Often these will be the only parameters you need to modify.
> 
> **Secondary configuration**
> 
> These parameters are other high-level parameters which may affect multiple Snakemake rules, or modify which rules are run.
> 
> **Per-Rule configuration**
> 
> Each top-level parameter here corresponds to a single Snakemake rule.
> `subsampling` is a commonly used rule configuration which is described separately in the **Primary configuration** section.

victorlin (changes requested):

docs/src/reference/configuration.rst
```diff
@@ -141,455 +177,638 @@ title
...
+-  default: ``true``
 
-metadata
-~~~~~~~~
 
--  type: string
--  description: Path to a local or remote (S3, HTTP(S), GS) tab-delimited metadata file supported by Augur. Metadata can be uncompressed or compressed.
--  examples:
+priorities:
+-----------
```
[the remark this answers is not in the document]
jameshadfield: Thanks! Fixed in [442a5fd](https://github.com/nextstrain/ncov/pull/907/commits/442a5fd19424d0874172d02e6972e1d834f99c4f)

docs/src/reference/configuration.rst
jameshadfield: Agreed! Fixed in [9b267de](https://github.com/nextstrain/ncov/pull/907/commits/9b267de3ddbe58a2f79b3f0c69af08ebc2a58841) but I didn't add the "[optional]" prefix because in practice none of those attributes are required:

> On the surface we lose the distinction about what parameters were
required, however this was misleading as in practice all required
parameters are defined in our defaults, and Snakemake merges dicts
so that it is perfectly valid to only specify a subset of "required"
parameters.

docs/src/reference/configuration.rst
jameshadfield: Thanks for the improvements :) switched to these in [734bd66](https://github.com/nextstrain/ncov/pull/907/commits/734bd664f45071faee8769b8bb234403e206fd4e)

docs/src/reference/configuration.rst
```diff
@@ -141,421 +178,539 @@ title
...
----------------
 
--  type: object
--  description: Parameters for estimation of logistic clade growth based on logit-transformed clade frequencies.
--  required:
+Per-Rule configuration
+^^^^^^^^^^^^^^^^^^^^^^
+
+Each top-level parameter here corresponds to a single Snakemake rule.
+Note that `subsampling`` is a commonly used rule configuration which is described separately in the Primary configuration section.`
```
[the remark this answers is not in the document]
jameshadfield: 🤦 now fixed. Thanks for catching!

PR #912, opened 2022-04-11
Replace 27259 as 21M defining mut as reverted in BA.5

corneliusroemer:
- fix: replace 27259 in clades.tsv as reverted in BA.5

@trvrb This PR contains the other commits, but let's just rebase on merge then they disappear.


trvrb:
Superseded by https://github.com/nextstrain/ncov/pull/913.

PR #916, opened 2022-04-13
Colorings update

trvrb:

This PR makes some minor changes to the Nextstrain GISAID and Nextstrain open profiles to:
- Fix capitalization to be consistent
- Correct "PANGO" to "Pango"
- Add a proper legend for "epiweek"
- Remove "age" and "sex" which aren't useful for large-scale analyses from GISAID profile
- Revise filter ordering to match color ordering
- Consistently order region / country / division / location

This PR also updates color ordering to incorporate the latest Pango lineages.

## Testing

Tested locally.


jameshadfield (approved):
All good. 

Would be nice to have a template `auspice_config.json` so we don't have to make so many duplicate changes

PR #917, opened 2022-04-13
Update emerging lineages

trvrb:

Current emerging lineages are quite outdated. This isn't as comprehensive as we might like things, but it's a good start.

This includes just BA.1.1 and BA.2.3 which were obvious enough to delineate from the current regional builds.

Screen Shot 2022-04-12 at 8 05 42 PM

## Testing

Tested locally.


corneliusroemer:
I added BA.4/5 and the BA.2 sublineages with S:L452 as they all have a good chance of becoming big.

I suspect that's what we want emerging lineages to be - things that may grow large and are interesting to highlight even if still not too big.

If you don't like, simply revert the commits from the branch @trvrb 

I wouldn't call BA.2.3 emerging at this point - it's early diversity, good to have labeled but not really anything that grows unusually at this point.

corneliusroemer:
The nextclade reference trees are perfect for topology. Thanks for making me publish them - I find them really useful now myself. Hadn't though about using them quite like this.
They are nice and clean (because well curated through Pango)
You could for example just colour by S:452 and see the 452 lineages at a spot!
New pango designation release just out, will quickly do a new run so we have a new reference build (that may or may not make it into Nextclade - but good to have just for inspection of diversity!)

corneliusroemer (approved):
I like

trvrb:
Trial build running on GISAID data at https://github.com/nextstrain/ncov/actions/runs/2168585685. This will deploy to:
- https://nextstrain.org/staging/ncov/gisaid/trial/emerging-lineages/global
- https://nextstrain.org/staging/ncov/gisaid/trial/emerging-lineages/africa
etc...
when finished.

trvrb:
Okay. I've just done a fair bit of testing here on top of latest GISAID builds and I'm quite happy with robustness of emerging lineage designations. I'll go ahead and merge. Thanks @corneliusroemer!

PR #918, opened 2022-04-13
Template auspice config files for our core builds

jameshadfield:
The auspice config files were almost duplicates of each other, resulting
in 14 very similar JSONs across the open & genbank profiles. This made
it harder than it should be to change config values (e.g. see
https://github.com/nextstrain/ncov/pull/916) and also made it all-too-
easy for things to fall out-of-sync.

The adds a custom rule which is only run for nextstrain core builds.
The rule is intended to be human readable and makes it clear which
are the values which differ across open/genbank builds as well as across
the different builds within those profiles. The vast majority of the
config is unchanged.

There is one change in content made here: The orderings of {region, country,
division} colorings are now in that order for both profiles; previously
the open profile had them reversed.


nextstrain_profiles/nextstrain-gisaid/config.yaml
```diff
@@ -9,4 +9,4 @@ show-failed-logs: True
 restart-times: 2
 reason: True
 stats: stats.json
-set-threads: tree=4
```
jameshadfield, line 12: This gave me a _very_ cryptic error when running locally:

```
Invalid threads definition: entries have to be defined as RULE=THREADS pairs (with THREADS being a positive integer).
Unparseable value: '{tree:'.
```
The error went away if I used a `shell` block within `make_auspice_config` rather than a `run` block. This may be an issue on my end, and commenting out this line isn't a good solution, but I ran out of debugging ideas!

workflow/snakemake_rules/main_workflow.smk
```diff
@@ -1360,13 +1360,27 @@ rule build_description:
...
+    """
+    Auspice configs are chosen via this heirarchy:
+    1. A build-specific JSON
+    2. A `make_auspice_config` rule, if it exists
+    3. The file specified in the `config.files` dict
+    """
+    if "auspice_config" in config["builds"].get(w.build_name, {}):
+        return config["builds"][w.build_name]["auspice_config"]
+    if "make_auspice_config" in rules.__dict__:
+        return rules.make_auspice_config.output
```
tsibley, line 1373: Would probably call this rule just `auspice_config` to align with how we name ~all other rules (all of which "make" something :-).

Also, `rules.__dict__` is more idiomatically spelled `vars(rules)` FWIW, but personally I don't think it much matters either way.

workflow/snakemake_rules/nextstrain_auspice_config.smk
```diff
@@ -0,0 +1,160 @@
...
+        ## What are the parameters which vary across builds?
+        default_geo_resolution = "country" if build_name in ["reference", "global", "africa", "south-america", "asia", "europe"] else "division"
+        default_map_triplicate = True if build_name in ["reference", "global"] else False
+        data_provenance = [{"name": "GISIAD"}] if is_gisaid else [{"name": "GenBank", "url": "https://www.ncbi.nlm.nih.gov/genbank/"}]
+        gisaid_clade_coloring = {"key": "GISAID_clade", "title": "GISAID Clade", "type": "categorical"} if is_gisaid else None
+        gisaid_epi_isl_coloring = {"key": "gisaid_epi_isl", "type": "categorical"} if is_gisaid else None
+        location_coloring = {"key": "location", "title": "Location", "type": "categorical"} if is_gisaid else None
+        location_filter = "location" if is_gisaid else None
+        originating_lab_filter = "originating_lab" if is_gisaid else None
+        submitting_lab_filter  = "submitting_lab" if is_gisaid else None
```
tsibley, line 25: These conditionals assume GenBank/Open if not GISAID instead of explicitly testing for GenBank/Open, which is the better/more future-proof practice. Would prefer to explicitly test, plus also make the test against `inputs` consider the whole value and not just single-key presence. There's a number of ways to make this approach syntactically palatable, but here's one example:

```suggestion
        input_set = set(config["inputs"])
        build_name = wildcards.build_name

        ## What are the parameters which vary across builds?
        default_geo_resolution = "country" if build_name in ["reference", "global", "africa", "south-america", "asia", "europe"] else "division"
        default_map_triplicate = True if build_name in ["reference", "global"] else False

        if input_set == {"gisaid"}:
            data_provenance = [{"name": "GISIAD"}]
            gisaid_clade_coloring = {"key": "GISAID_clade", "title": "GISAID Clade", "type": "categorical"}
            gisaid_epi_isl_coloring = {"key": "gisaid_epi_isl", "type": "categorical"}
            location_coloring = {"key": "location", "title": "Location", "type": "categorical"}
            location_filter = "location"
            originating_lab_filter = "originating_lab"
            submitting_lab_filter  = "submitting_lab"
        elif input_set == {"open"}:
            data_provenance = [{"name": "GenBank", "url": "https://www.ncbi.nlm.nih.gov/genbank/"}]
            gisaid_clade_coloring = None
            gisaid_epi_isl_coloring = None
            location_coloring = None
            location_filter = None
            originating_lab_filter = None
            submitting_lab_filter  = None
        else:
            raise Exception(f"rule auspice_config doesn't know how to handle inputs: {input_set}")
```

nextstrain_profiles/nextstrain-gisaid/builds.yaml
```diff
@@ -4,6 +4,7 @@ auspice_json_prefix: ncov_gisaid
 custom_rules:
   - workflow/snakemake_rules/export_for_nextstrain.smk
   - nextstrain_profiles/nextstrain-gisaid/subsampling_ranges.smk
+  - workflow/snakemake_rules/nextstrain_auspice_config.smk
```
tsibley, line 7: As the new rule is part of export steps, it seems the rule could just be part of the existing `export_for_nextstrain.smk` file?

tsibley (commented):
Good to make this easier and more consistent.

This sort of thing shouts a little bit for something like [CUE](https://cuelang.org), [Jsonnet](https://jsonnet.org), or [YAML Template Engine (YTE)](https://github.com/koesterlab/yte) (which [Snakemake integrates support for](https://snakemake.readthedocs.io/en/stable/snakefiles/rules.html?highlight=yte#template-rendering-integration) in newer versions). (With YTE, we'd render a YAML template and then convert to JSON for Augur.)

nextstrain_profiles/nextstrain-gisaid/config.yaml
tsibley: Hmm. It seems worth solving this before merge to me, or do you think not?

Can you paste more of the error/logs and how you're invoking the workflow? I can't reproduce this locally so far…

tsibley: I mean, I can reproduce the exact message if I do very specifically this, but obviously this is nonsense:

```diff
-# set-threads: tree=4
+set-threads: "{tree:"
```

Trying something less nonsense like:

```diff
-# set-threads: tree=4
+set-threads:
+  tree: 4
```

gets me a different (and reasonablish) error: `Unparseable value: "{'tree': 4}".`

jameshadfield: Yeah for sure I'd like to solve this - that's why I flagged it in review 😉 

```
$ snakemake --version
7.3.8

## with the "set-threads: tree=4" commented out
$ snakemake --cores 1 --profile nextstrain_profiles/nextstrain-gisaid/ -pf results/global/auspice_config.json
normal output and everything works as expected and creates the config JSON

## with "set-threads: tree=4"  in ./nextstrain_profiles/gisaid/config.yaml
## i.e. the config file is identical to current master
$ snakemake --cores 1 --profile nextstrain_profiles/nextstrain-gisaid/ -pf results/global/auspice_config.json
Config file defaults/parameters.yaml is extended by additional config specified via the command line.
Building DAG of jobs...
Using shell: /usr/local/bin/bash
Provided cores: 1 (use --cores to define parallelism)
Rules claiming more threads will be scaled down.
Conda environments: ignored
Job stats:
job               count    min threads    max threads
--------------  -------  -------------  -------------
auspice_config        1              1              1
total                 1              1              1

Cannot send slack message as the config does not define a channel and/or token.
Select jobs to execute...

[Thu May  5 13:07:21 2022]
Job 0: Making a custom auspice config.
Reason: Forced execution

Invalid threads definition: entries have to be defined as RULE=THREADS pairs (with THREADS being a positive integer). Unparseable value: '{tree:'. 
```

nextstrain_profiles/nextstrain-gisaid/builds.yaml
jameshadfield: Sure! I tend to think of that file more as an "upload assets" file, but it's actually more general than that so I'm happy to put the config JSON rule there.

workflow/snakemake_rules/nextstrain_auspice_config.smk
jameshadfield: Sure!

workflow/snakemake_rules/main_workflow.smk
jameshadfield: Sure! Done.

jameshadfield:
Thanks for 👀 tom - much appreciated!

Force pushed to incorporate requested changes & rebase onto master. 

> This sort of thing shouts a little bit for something like [CUE](https://cuelang.org/), [Jsonnet](https://jsonnet.org/), or [YAML Template Engine (YTE)](https://github.com/koesterlab/yte) (which [Snakemake integrates support for](https://snakemake.readthedocs.io/en/stable/snakefiles/rules.html?highlight=yte#template-rendering-integration) in newer versions). (With YTE, we'd render a YAML template and then convert that JSON for Augur.)

I see things a bit differently - I think everyone in the team is familiar with python, and I think the added rule is straightforward to read through. I'd hope everyone feels comfortable editing it. On the contrary, I think few (?) people would be familiar with the ins-and-outs of whatever templating language was picked. That's not to say we shouldn't use a templating language somewhere someday, but that the move will come with a learning burden duplicated over many people.

nextstrain_profiles/nextstrain-gisaid/config.yaml
jameshadfield: I've reinstated the `set-threads` setting to [allow an AWS test run](https://github.com/nextstrain/ncov/actions/runs/2273245463). If that passes then I'm happy to go with it, as that'll be n=2 setups which can't reproduce the issue, but I would really like to know what's going on!

tsibley:
> I see things a bit differently…

Nod. I see where you're coming from, and I do think the current approach in this PR is just fine. I see this sort of thing as a tradeoff—and I definitely tend to skew more towards "it's worth learning X, at least a little"—but this case isn't at a tipping point for me.

nextstrain_profiles/nextstrain-gisaid/config.yaml
tsibley: Ah, this is a Snakemake version issue. The Conda environment defined by this repo gets a 6.x version of Snakemake and the Nextstrain Docker runtime has a 5.x version, neither of which exhibit the "Invalid threads definition" error. However, if I use 7.3.8, 7.4.0, or 7.5.0 then I can reproduce. The error goes away again with 7.6.0.

Looking at the code/issues finds these:

- https://github.com/snakemake/snakemake/commit/0771062a07f0e2cfe9ee45a2276aa61b096eb6e1
- https://github.com/snakemake/snakemake/pull/1617
- https://github.com/snakemake/snakemake/issues/1622

jameshadfield: Ouch. Well, at least we've figured it out.

> The Conda environment defined by this repo gets a 6.x version of Snakemake

I'm missing something... where is this environment? We have `workflow/envs/nextstrain.yaml`, but that doesn't specify snakemake (as it's intended to be run by snakemake itself). The [ncov docs](https://docs.nextstrain.org/projects/ncov/en/latest/tutorial/example-data.html) say to install via the [nextstrain docs](https://docs.nextstrain.org/en/latest/install.html) which don't pin versions at all: `mamba create ... snakemake`. Augur itself specifies `"snakemake >=5.4.0"`.

jameshadfield:
Test builds worked as expected & the snakemake 🐛 has been solved so I plan to merge this on Monday (NZ)

nextstrain_profiles/nextstrain-gisaid/config.yaml
tsibley: Ah, you're right. I was confused. `workflow/envs/nextstrain.yaml` does include Snakemake transitively and it gets 6.8.0 (not the latest), but the `mamba create` command in our install docs gets latest Snakemake (7.6.1).

tsibley: (There was something else going on yesterday where I tried created a new Conda env with just Snakemake and got an older version, but I now suspect that related to missing channels and/or whatever the default Python pin was.)

PR #925, opened 2022-04-20
Omicron only

jameshadfield:
This PR adds an example which produces an omicron-only build. Viewable here: https://nextstrain.org/staging/ncov/omicron/2022-04-20

image

To do this we have to modify the rules and add an extra (optional) parameter `config["files"]["include_for_subsampling"]` so that we can include a set of strains for the alignment and subsequently exclude them for the tree building stage.

The resulting build has mutations in the correct co-ordinate system, but note that the ancestral state is the inferred omicron ancestral state, not the canonical Wuhan state. For instance, the root has [S:501Y](https://nextstrain.org/staging/ncov/omicron/2022-04-20?c=gt-S_501). 

One limitation of the implementation here is that you cannot have an omicron-only build within the larger (core) Nextstrain builds because the choice of which files to include is per-workflow, not per-build.

### DIFF
--- a/docs/src/reference/configuration.rst
+++ b/docs/src/reference/configuration.rst
@@ -470,9 +470,17 @@ include
 ~~~~~~~
 
 -  type: string
--  description: Path to a file with list of strains (one name per line) to include in the analysis regardless of priorities or subsampling during filtering.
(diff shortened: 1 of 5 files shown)


rneher:
Is this made obsolete by the `21L` builds we have now?
https://nextstrain.org/ncov/gisaid/21L/global/6m

jameshadfield:
Yeah. Too many dangling PRs!

PR #926, opened 2022-04-20
Sanitize sequences prior to nextclade run

huddlej:

Sanitizes sequences, removing any duplicate sequences, prior to running nextclade on subsampled data. This additional check should not add too much to the runtime of the workflow, but it will allow the workflow to continue running when input data contain duplicates and were not sanitized by earlier steps in the workflow (e.g., full alignment or multiple combined input sequences).

## Related issue(s)

Related to https://github.com/nextstrain/ncov-ingest/issues/281 and [this discussion board post](https://discussion.nextstrain.org/t/error-in-augur-tree-duplicated-sequence-name/970/7).

## Testing

 - [x] Tested by CI
 - [x] Tested with duplicate sequences

### DIFF
--- a/workflow/snakemake_rules/main_workflow.smk
+++ b/workflow/snakemake_rules/main_workflow.smk
@@ -478,7 +478,9 @@ rule build_align:
     params:
         outdir = "results/{build_name}/translations",
         genes = ','.join(config.get('genes', ['S'])),
-        basename = "aligned"
+        basename = "aligned",
+        strain_prefixes=config["strip_strain_prefixes"],
+        sanitize_log="logs/sanitize_sequences_before_nextclade_{build_name}.txt",
     log:
         "logs/align_{build_name}.txt"
     benchmark:
@@ -489,7 +491,11 @@ rule build_align:
         mem_mb=3000
     shell:
         """
-        xz -c -d {input.sequences} |  nextclade run \
+        python3 scripts/sanitize_sequences.py \
+            --sequences {input.sequences} \
+            --strip-prefixes {params.strain_prefixes:q} \
+            --output /dev/stdout 2> {params.sanitize_log} \
+            | nextclade run \
             --jobs {threads} \
             --input-fasta /dev/stdin \
             --reference {input.reference} \
--- a/workflow/snakemake_rules/main_workflow.smk
+++ b/workflow/snakemake_rules/main_workflow.smk
@@ -252,10 +252,17 @@ rule index_sequences:
     benchmark:
         "benchmarks/index_sequences.txt"
     conda: config["conda_environment"]
+    params:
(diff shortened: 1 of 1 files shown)


victorlin (approved):

huddlej:
I just tested this with the CI build but with the build inputs changed to:

```yaml
inputs:
  - name: gisaid
    metadata: "data/example_metadata.tsv"
    aligned: "data/aligned_gisaid.fasta.xz"
```

And with the aligned data containing an extra copy of the sequence for Australia/VIC05/2020. I ran the build like so to make sure that the duplicated sequences made it into the subsampled sequences:

```
snakemake \
  results/europe/europe_subsampled_sequences.fasta.xz \
  -p \
  -j 4 \
  --forceall \
  --use-conda \
  --conda-frontend mamba \
  --configfile nextstrain_profiles/nextstrain-ci/builds.yaml
```

This command failed when running `scripts/priorities.py` because the sequence index includes duplicates and the script throws the following pandas error:

```python
  File "scripts/priorities.py", line 24, in <module>
    combined = pd.concat([proximities, index], axis=1)
  File "/Users/jlhudd/projects/nextstrain/ncov/.snakemake/conda/3d27c4cf5509fba4b95c3fa6ba174aa7/lib/python3.8/site-packages/pandas/util/_decorators.py", line 311, in wrapper
    return func(*args, **kwargs)
  File "/Users/jlhudd/projects/nextstrain/ncov/.snakemake/conda/3d27c4cf5509fba4b95c3fa6ba174aa7/lib/python3.8/site-packages/pandas/core/reshape/concat.py", line 360, in concat
    return op.get_result()
  File "/Users/jlhudd/projects/nextstrain/ncov/.snakemake/conda/3d27c4cf5509fba4b95c3fa6ba174aa7/lib/python3.8/site-packages/pandas/core/reshape/concat.py", line 591, in get_result
    indexers[ax] = obj_labels.get_indexer(new_labels)
  File "/Users/jlhudd/projects/nextstrain/ncov/.snakemake/conda/3d27c4cf5509fba4b95c3fa6ba174aa7/lib/python3.8/site-packages/pandas/core/indexes/base.py", line 3721, in get_indexer
    raise InvalidIndexError(self._requires_unique_msg)
pandas.errors.InvalidIndexError: Reindexing only valid with uniquely valued Index objects
```

This error shows that there are multiple potential points of failure when duplicate sequences enter the workflow. A better solution than the current one might be to sanitize sequences on the way into augur filter. But, the other general solution that I mentioned on Slack is to make augur filter only emit unique sequences. This solves the problem at the Augur level instead of the workflow level, so there should be fewer points of failure. I'm leaning toward this latter solution and it seems like other folks prefer this approach, too.

PR #933, opened 2022-04-27
Update clades with 22A, 22B and 22C

trvrb:
### Overview

This elevates the following Pango lineages as labeled clades:
- BA.4 as 22A
- BA.5 as 22B
- BA.2.12.1 as 22C

Each of these still has the parenthetical "(Omicron)" designation as these are all still considered to be the Omicron variant by the WHO.

Here, we are following updated Nextclade designation rules that call for a clade to labeled if any of the following are met:
1. A VOC or VOI is recognized by the WHO and given a Greek letter label
2. A clade reaches >20% global frequency for 2 or more months
3. A clade reaches >30% regional frequency for 2 or more months
4. A clade shows consistent >0.05 per day growth in frequency where it's circulating and has reached >5% regional frequency

with additionally (2,3,4) requiring the clade to show multiple mutations in S1 or particular mutations of known biological relevance

Evidence of each of the three clades hitting these criteria is provided below:

### BA.4 as 22A

- Current view of BA.4 suggests consistent growth in Gauteng (0.09 per day) with it reaching 78% frequency on April 7, but not yet enough data in KwaZulu-Natal.
- We place BA.4 at [14% regional frequency in Africa](https://nextstrain.org/ncov/gisaid/africa?c=emerging_lineage&f_region=Africa).
- BA.4 possesses spike mutations L452R, F486V and R493Q relative to BA.1 and BA.2 and additionally posses del69/70 relative to BA2.

omicron-south-africa-ba4-ba5_logistic-growth-transformed-axis-ba4

### BA.5 as 22B

- Current view of BA.5 suggests consistent growth in Gauteng (0.07 per day) with it reaching 11% frequency on April 7 and consistent growth in KwaZulu-Natal (0.11 per day) with it reaching 56% frequency on April 7.
- We place BA.5 at [6% regional frequency in Africa](https://nextstrain.org/ncov/gisaid/africa?c=emerging_lineage&f_region=Africa).
- BA.5 possesses spike mutations L452R, F486V and R493Q relative to BA.1 and BA.2 and additionally posses del69/70 relative to BA2.

omicron-south-africa-ba4-ba5_logistic-growth-transformed-axis-ba5

### BA.2.12.1 as 22C

- Current view of BA.2.12.1 suggests consistent growth in multiple US states (generally around 0.1 per day, but CA may be faster at 0.15 and NY may be slower at 0.06). Its frequencies are greater than 20% in a number of states as of April 7.
[...]

PR #934, opened 2022-04-28
Restrict 6m builds to tips in the last 6m (post-tree-building)

jameshadfield:
This introduces a script which restricts the 6m builds to include tips from the previous 6 months only. This is done _after_ tree building, so that tree reconstruction / inference can use data from the entire pandemic. After tips are restricted we then partition the tips based on clades and visualise them as subtrees. If a clade splits after the cutoff (i.e. in the last 6 months) then it's not drawn as a separate subtree. 

I went through a number of iterations of this approach, and there are lots of judgement calls to be made. Very open to modifications / changes / comments 😄 

Specifically, most of my test datasets had some recent tips of an old clade (e.g 20A, 20C) and the coalescence of those tips was a long time ago. This pushed the root of at least one subtree quite far back. Even just a few delta strains will typically push the coalescence back to around the start of 2021. 

For instance, here's the global 6m build unrestricted (left), restricted to 6M (middle) and 3M (right). Notice the difference in minimum date due to tips from 20B not appearing in the last 3 months:

image

Here's the same restrictions but for the build from https://github.com/nextstrain/ncov/pull/933. Note how BA4/5 aren't split into separate subtrees as they branch off after the cutoff date. (I played around with using different cutoffs here, the further back you push this the more structure comes through.)

image

And for good measure, here's the h1n1pdm tree focused on the past 6 months (left) and restricted to the last 6 months (right) (colours are different because they aren't defined in the dataset, they're created by auspice):

image

Trial builds running now: https://github.com/nextstrain/ncov/actions/runs/2237163006

### DIFF
--- a/docs/src/reference/configuration.rst
+++ b/docs/src/reference/configuration.rst
@@ -179,6 +179,11 @@ title
 -  type: string
 -  description: Build-specific title to provide to ``augur export`` and display as the title of the analysis in Auspice.
 
+date_cutoff
+~~~~~~~~~~~
+
+-  type: string or number. Either a numeric date with the year as the integer part (e.g. 2020.42), or a date in ISO 8601 date format (i.e. YYYY-MM-DD) (e.g. '2020-06-04'), or a backwards-looking relative date in ISO 8601 duration format with optional P prefix (e.g. '1W', 'P1W').
+-  description: Restrict the build to tips observed after this cutoff, and create subtrees for each clade which originates before this time. Note that this happens *after* the tree has been inferred & temporally resolved.
 
 
 subsampling
--- a/nextstrain_profiles/nextstrain-gisaid/builds.yaml
+++ b/nextstrain_profiles/nextstrain-gisaid/builds.yaml
@@ -48,6 +48,7 @@ builds:
     subsampling_scheme: nextstrain_global_6m
     auspice_config: "nextstrain_profiles/nextstrain-gisaid/global_auspice_config.json"
     title: Genomic epidemiology of SARS-CoV-2 with subsampling focused globally over the past 6 months
+    date_cutoff: 6M
   global_all-time:
     subsampling_scheme: nextstrain_global_all_time
     auspice_config: "nextstrain_profiles/nextstrain-gisaid/global_auspice_config.json"
@@ -57,6 +58,7 @@ builds:
     region: Africa
     auspice_config: "nextstrain_profiles/nextstrain-gisaid/africa_auspice_config.json"
     title: Genomic epidemiology of SARS-CoV-2 with subsampling focused on Africa over the past 6 months
+    date_cutoff: 6M
   africa_all-time:
     subsampling_scheme: nextstrain_region_grouped_by_country_all_time
     region: Africa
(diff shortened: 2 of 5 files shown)


emmahodcroft:
This is cool James! Just to be sure I'm understanding, I'm guessing the idea here is to allow a more focused view of the tips, without having a tree that 'drags' back further into the past than people may be interested in looking? In other words, it allows a 'shortening' of the X-axis time span? (Or is there something grander at play than I'm picking up - very possible!)

In the same vein, when you say things like "push the coalescence back" you mean just visually, right? Since you say they're the same tree-building, the actual coalescence stays the same (seems to, across the views, I think), but the visual 'stretching' depends on how deeply branches go (which is definitely impacted by where you set the cut off).

Would the proposal be to make this an toggle-able view of the 6m builds, or the 'only' view? I do think it's very cool, but am wondering if people may get the idea that SARS-CoV-2 has split into different... somethings. I'll let the imagination of the headlines run wild in everyone's own head... 🙃 On the other hand, someone's always willing to misinterpret, and we can't let that stop us trying to get clearer & better visuals!

jameshadfield:
> Just to be sure I'm understanding, I'm guessing the idea here is to allow a more focused view of the tips, without having a tree that 'drags' back further into the past than people may be interested in looking?

Exactly! It's to focus on the within-clade relationships of what's circulating recently, rather than convey the evolutionary history of the different clades. 

If we showed the entire connected history of those selected tips then (a) the root node would often be so far back that we don't gain that much horizontal space and (b) it won't work for recombinants. So I'm partitioning the data (via `clade` label) and showing the connected history of those partitions. Unfortunately, if you get a couple of recent "20A" samples (for instance), then that set of tips coalesces a long time ago, thus pushing back the x-axis min-date. There are ways to fix this, for instance specifying another time limit and cutting branches off if they go before this, but they're not ideal.

jameshadfield:
Force-pushed & rerunning ([GitHub Action](https://github.com/nextstrain/ncov/actions/runs/2242011591)) as our docker image python doesn't like walruses!

emmahodcroft:
Thanks James! That makes sense and I really like this idea.

One thought - though may be beyond the scope of this PR - is whether this is something that would be desirable/possible to introduce on Auspice end rather than build end. Like another version of the date slider, where instead of a 'vertical' cutoff (all being along the same line) it was a 'tip' cutoff (like the above) where it then only shows back to the coalescent point of tips in the specified period. And instead of 'greying out' the branches prior to this, do away with them completely and rescale the X-axis. This would let users view or not view the deeper tree, or set their limit to 3m, 6m, etc (of course, there could be a default, too!), rather than 'losing' the information entirely. Could also then be applied to a tree without additional running!

jameshadfield:
> Like another version of the date slider, where instead of a 'vertical' cutoff (all being along the same line) it was a 'tip' cutoff (like the above) where it then only shows back to the coalescent point of tips in the specified period.

For sure, we want something like this (for multiple reasons) and I've sketched out a few ideas but nothing seems quite right at the moment in terms of UI. Note that the coalescent is often further back than one might think (e.g. the coalescent for tips in the last 6 months of nCoV often goes back to early 2020, depending on the subsampling).

I'm interested in pushing this subtree idea because as we start to sample `X*` lineages I think this is the best way to present the data. I'd argue it's how omicron should be presented as well. In terms of focusing on temporal ranges however, we may be able to get there by auspice improvements and/or labelling nodes such that we can explode the tree appropriately within auspice.

victorlin:
@trvrb should this be closed in favor of #1132?

trvrb:
Yes. My preference would be to just go with the updated color scheme for the moment and shelve updated temporal sampling / cutoffs for the moment. I think these latter approaches make better use of the canvas to focus on what's circulating, but I believe add risk of confusion by readers. In addition, I'd think its a bit more semantically appropriate to only create subtrees when there's recombination.

PR #936, opened 2022-04-28
Migrate CI from Travis → GitHub Actions

tsibley:
For consistency across our org and within this repo (which was already
using GitHub Actions for other things).

Notable changes:

- The example_sequences.fasta.gz is no longer explicitly gunzipped;
  instead the build itself will do that automatically.

- Cram is run in parallel to the build. \o/

- The login to Docker Hub is dropped as it would require more plumbing
  through of secrets than I care to do right now.  It was added to
  address an acute issue of the way Travis shares network egress; if
  it's similarly an issue on GitHub Actions runners, we can do the
  plumbing then.

- Nextstrain CLI is installed from the latest stable release instead of
  from git.  This is actually preferred so a temporarily-broken master
  branch doesn't break this CI.

## Related issue(s)
Related to nextstrain/private#3.

## Testing
- [x] Check CI passes

## Before merge
- [x] Merge https://github.com/nextstrain/.github/pull/19
- [x] Drop `tmp!` commit at branch tip


victorlin (approved):
👋

PR #939, opened 2022-04-29
Update default masking from end to 200 bases

trvrb:

Update default mask parameters to mask 200 bases from the end of the genome rather than the existing 50. This was necessary because there is a large deletion in this region in circulating 21L viruses. This deletion is causing problems with alignment and the resulting mis-alignment appears as excess mutations in the tree.

The issue can be currently seen at https://nextstrain.org/ncov/gisaid/global?c=gt-nuc_29765,29766,29767,29768,29760&m=div.

mutations

## Testing

Testing has been fairly extensive with @corneliusroemer's https://nextstrain.org/groups/neherlab builds using https://github.com/neherlab/ncov-simple.

## Release checklist

### DIFF
--- a/defaults/parameters.yaml
+++ b/defaults/parameters.yaml
@@ -101,7 +101,7 @@ mask:
   # Number of bases to mask from the beginning and end of the alignment. These regions of the genome
   # are difficult to sequence accurately.
   mask_from_beginning: 100
-  mask_from_end: 50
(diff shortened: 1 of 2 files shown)


emmahodcroft:
Worth noting that if/when Nextalign can handle this deletion better, this masking could be reduced.

PR #941, opened 2022-05-03
docs: add CI, fix formatting issue

victorlin:
## Description of proposed changes

- Add CI for docs project (catches existing formatting issue on Terra guide).
- Fix formatting issue.

## Related issue(s)

_N/A_

## Testing

See GitHub checks on individual commits.

## Release checklist

_N/A_

## Pre-merge

- [x] drop temporary commit

PR #942, opened 2022-05-03
Minor edits to ncov tutorial

huddlej:

This PR introduces minor edits for new ncov tutorial content. See commit messages for more details.

The most controversial of these commits might be the most recent two in which I a) define a custom docutils directive to dynamically define the collection date range for the genomic surveillance tutorial (instead of hardcoding our own date values) and b) reduce the total number of samples in the genomic surveillance build by an order of 100 so the build runs in 3 minutes instead of 1.5 hours (with 3 CPUs).


docs/src/guides/run-analysis-on-terra.rst
```diff
@@ -50,19 +50,19 @@ III. Connect your data files to the wdl workflow
...
-  |Nextstrain_WRKFLW|  sequence_fasta | File  | this.sequences       |
-  +-----------------+-----------------+-------+----------------------+
+#. Most of the values will be blank but fill in the values below:
+
+  +-----------------+------------------+-------+----------------------+
+  |Task name        | Variable         | Type  |   Attribute          |
+  +=================+==================+=======+======================+
+  |Nextstrain_WRKFLW|  build_name      | String| this.ncov_example.id |
+  +-----------------+------------------+-------+----------------------+
+  |Nextstrain_WRKFLW|  configfile_yaml | File  | this.configfile_yaml |
```
victorlin, line 60: I addressed this in #941 but we can take whichever gets merged first.

readthedocs.yml
```diff
@@ -1,4 +1,4 @@
 ---
 version: 2
 conda:
-  environment: docs/conda.yml
+  environment: docs/environment.yml
```
victorlin, line 4: Will need to sync this change with #941

docs/src/tutorial/example-data.rst
```diff
@@ -37,7 +37,7 @@ Setup
 Run the workflow
 ----------------
 
-From within the ``ncov/`` directory, run the ``ncov`` workflow using a configuration file provided in the tutorial directory:
+From within the ``ncov/`` directory, run the workflow using a configuration file provided in the tutorial directory:
```
victorlin, line 40: Another small improvement, as the first reference of "configuration file" in these tutorials:

```suggestion
From within the ``ncov/`` directory, run the workflow using a :term:`configuration file <config file>` provided in the tutorial directory:
```

docs/src/tutorial/example-data.rst
```diff
@@ -54,7 +54,7 @@ The workflow can take several minutes to run. While it is running, you can learn
    - This required Snakemake parameter specifies the number of CPU cores to use (:doc:`more info <snakemake:executing/cli>`).
 - ``--configfile ncov-tutorial/example-data.yaml``
    - ``--configfile`` is another Snakemake parameter used to configure the ncov workflow.
-   - ``ncov-tutorial/example-data.yaml`` is a YAML file which provides custom workflow configuration including inputs and outputs. Contents with comments excluded:
+   - ``ncov-tutorial/example-data.yaml`` is a YAML file that provides custom workflow configuration including inputs and outputs. The contents of this file with comments excluded are:
```
victorlin, line 57: YAML detail can be kept in [the glossary entry](https://nextstrain--942.org.readthedocs.build/projects/ncov/en/942/reference/glossary.html#term-config-file).

```suggestion
   - ``ncov-tutorial/example-data.yaml`` is a configuration file that provides custom workflow configuration including inputs and outputs. The contents of this file with comments excluded are:
```

docs/src/tutorial/genomic-surveillance.rst
```diff
@@ -27,6 +32,9 @@ and activate the ``nextstrain`` conda environment:
 
       conda activate nextstrain
 
+.. |today| date::
```
victorlin, line 35: Can this be moved closer to the line defining `|last-month|`?

docs/src/tutorial/custom-data.rst
```diff
@@ -142,9 +146,9 @@ This is the same as the previous file, with some additions:
            }
          }
 
-      This JSON does two things:
+      This JSON tells Auspice to:
 
-      1. Create a new coloring ``custom_data`` which reflects a special metadata column generated by the ncov workflow. When there is more than one input, each data input produces a new final metadata column with categorical values ``yes`` or ``no`` representing whether the sequence was from the input.
+      1. Create a new coloring ``custom_data`` that reflects a special metadata column generated by the ncov workflow. When there is more than one input, each data input produces a new final metadata column with categorical values ``yes`` or ``no`` representing whether the sequence was from the input.
```
victorlin, line 151: Thanks for catching my small grammatical mistakes!

docs/src/tutorial/genomic-surveillance.rst
```diff
@@ -38,10 +46,13 @@ We will download a focal set of Idaho sequences from GISAID's EpiCoV database.
...
       :alt: GISAID EpiCoV Search
 
+.. |last-month| timedelta::
+   :weeks: -4
+
 2. Filter to sequences that pass the following criteria:
 
    1. From Idaho, USA
-   2. Collected within the last month (between ``2022-03-01`` and ``2022-04-01`` at the time of writing)
+   2. Collected within the last month (between |last-month| and |today| at the time of writing)
```
victorlin, line 55: Reword given dynamic dates:

```suggestion
   2. Collected within the last month (e.g. |last-month| to |today|)
```

victorlin, line 55: Convert `image` to `figure`, add a caption explaining the static dates:

```rst
   .. figure:: ../images/gisaid-select-sequences-idaho-highlighted.png
      :width: 700
      :alt: GISAID EpiCoV filter and select sequences

      Example of GISAID filters using collection date between 2022-03-01 and 2022-04-01
```

huddlej:
One change I wanted to make but did not was to add some mention of `--use-conda` as the best way to keep a consistent, current conda environment for users with the "native" runtime (and drop the current admonition box about running `mamba install` for users of this runtime). This change relates to [a general issue we have with our installation docs](https://github.com/nextstrain/docs.nextstrain.org/issues/96#issuecomment-1074263894), but it doesn't seem to be a "minor edit" even if it only involves a couple lines of changed text.

docs/src/conf.py
```diff
@@ -113,3 +115,40 @@ def prose_list(items):
...
+            if option in self.options:
+                kwargs[option] = int(self.options[option])
+
+        date_object = date.today()
+        if kwargs:
+            date_object = date_object + timedelta(**kwargs)
+
+        date_text = date_object.strftime("%Y-%m-%d")
+
+        return [nodes.Text(date_text)]
```
victorlin, line 152: Very neat!

victorlin (changes requested):

docs/src/tutorial/genomic-surveillance.rst
```diff
@@ -101,22 +112,22 @@ The workflow can take several minutes to run. While it is running, you can inves
      idaho_scheme:
        custom_sample:
          query: --query "(custom_data == 'yes')"
-         max_sequences: 5000
+         max_sequences: 50
```
victorlin, line 115: We should probably update the screenshot to reflect these changes. I will push a new commit.

victorlin (changes requested):

readthedocs.yml
huddlej: We could also switch the dev docs to use `mamba env create -f=conda.yml` if switching the name back is too disruptive.

docs/src/tutorial/genomic-surveillance.rst
huddlej: Good call! If you don't mind pushing an updated screenshot, that would be great.

docs/src/tutorial/genomic-surveillance.rst
huddlej: Do you want to update this when you update the screenshot?

docs/src/tutorial/genomic-surveillance.rst
huddlej: Fixed in 7ea33bf

docs/src/tutorial/genomic-surveillance.rst
victorlin: Will do!

readthedocs.yml
victorlin: Not a big deal, I think consistency is more important. Looking at a few other docs projects:

- docs.nextstrain.org: [environment.yml](https://github.com/nextstrain/docs.nextstrain.org/blob/master/environment.yml)
- Auspice: [docs/environment.yml](https://github.com/nextstrain/auspice/blob/master/docs/environment.yml)
- Augur: I can't find one?

Based on this, I'm on board with the move from conda.yml to environment.yml.

tsibley: If I can push back a little bit, I think `conda.yml` is better than `environment.yml`, as it specifies what software the file is for... not a huge deal, but to me `environment.yml` is like naming a file `data.tsv`. ;-)

victorlin: @tsibley I buy that point, would just want to see the other docs projects adopt `conda.yml` as well. I can PR those if it's what we settle on here.

docs/src/guides/run-analysis-on-terra.rst
victorlin: #941 was merged, so I've dropped 396a5793c2bcbc70d3eba421ec370b2f579cf014.

readthedocs.yml
victorlin: Moved back to `conda.yml`.

docs/src/tutorial/genomic-surveillance.rst
victorlin: Done.

docs/src/tutorial/next-steps.rst
```diff
@@ -47,9 +47,10 @@ Then, use the following steps to customize names, titles, and context:
 1. Change the ``custom_data`` input filenames from ``idaho.metadata.tsv`` and ``idaho.sequences.fasta`` to your own.
 2. Change the regional input dataset from North America to an appropriate region for your custom focal data. :doc:`See the complete list of available URLs <../reference/remote_inputs>`.
 3. Rename the output dataset from ``idaho`` to your own. Note the name restrictions.
-4. Rename the subsampling scheme from ``idaho_scheme`` to your own. Note the name restrictions.
-5. Reword the output dataset title to your own.
-6. Rename the ``usa_context`` sample and update the ``query`` accordingly.
+4. Reword the output dataset title to your own.
+5. Rename the subsampling scheme from ``idaho_scheme`` to your own. Note the name restrictions.
+6. For each sample, increase the ``max_sequences`` to your own.
```
victorlin, line 52: I also added this as part of c22bfe2e.

victorlin (approved):

PR #947, opened 2022-05-13
Update wdl/ncov-ingest branch to latest

j23414:

Update `wdl/ncov-ingest` branch


victorlin (changes requested):
Why the PR? [Looks like `wdl/ncov-ingest` is just at an older commit in `master`, nothing new](https://github.com/nextstrain/ncov/tree/wdl/ncov-ingest):

> This branch is 9 commits behind master.

So you could just do something like this locally:

```sh
git fetch
git switch wdl/ncov-ingest
git merge origin/master --ff-only
```

victorlin:
To keep a clean history on `master` (no merge commit), I would actually recommend it here in place of this PR. Or use the **Rebase and merge option** when merging here.

j23414:
In order to clean up and address comments:

> To keep a clean history, I would actually recommend it here in place of this PR. Or use the Rebase and merge option when merging here. Otherwise, merge commits might get a bit messy going from master -> wd/ncov-ingest -> master (eventually in a new PR).

I'm going to copy my changes out, and restart this branch. Since the "wdl/ncov-ingest" is in drastic flux at the moment, it's not like I'll lose a whole lot of work.

victorlin:
Oh actually since the last comment you've added a commit 5e07d8c2190a9eb0aa634f5d5c4f76dce33b7a6e to `wdl/ncov-ingest`, so the recommendation to keep it updated with `master` wouldn't be to fast-forward merge but instead `git rebase master` from the branch. Or if you want to copy changes out then you can `git cherry-pick 5e07d8c2` on another branch.

PR #952, opened 2022-05-25
Fix annotate metadata with numeric strains

### LINKED ISSUE #948: `annotate_metadata_with_index` raises `ValueError` when dtypes don't match
**Current Behavior**
Within [annotate_metadata_with_index](https://github.com/nextstrain/ncov/blob/08cd17c2521001e370bfaaed24620ebda0411356/scripts/annotate_metadata_with_index.py#L18), the index TSV is read without explicitly set dtypes. This can lead to numerical id strain values to be interpreted as numbers, leading to a `ValueError` when merging with the metadata. 

**Additional context**
See error reported by user: https://discussion.nextstrain.org/t/value-error-trying-to-merge-on-object-and-int64-columns/1106

huddlej:

Fix annotate metadata with numeric strains by setting the dtype of the "strain" column in the sequence index to "string".

## Related issue(s)

Fixes #948

## Testing

 - [x] Added functional test for issue, confirmed that test failed, and confirmed that fix in c49a5a6 makes the test pass
 - [ ] Tested by CI


.github/workflows/ci.yaml
```diff
@@ -18,7 +18,7 @@ jobs:
         with:
           python-version: "3.9"
       - run: pip install cram nextstrain-augur
-      - run: cram --shell=/bin/bash tests/sanitize-metadata.t
+      - run: cram --shell=/bin/bash tests/sanitize-metadata.t tests/annotate_metadata_with_index.t
```
huddlej, line 21: Eventually, we want to run all Cram tests in the tests directory with a single `cram tests/` command, but the complexity and run-time of `different-inputs.t` prevents us from running all tests in each CI run. Resolving that issue is for a different PR.

victorlin: Today I learned there are Cram tests in ncov, too!

scripts/annotate_metadata_with_index.py
```diff
@@ -18,6 +18,7 @@
     index = pd.read_csv(
         args.sequence_index,
         sep="\t",
+        dtype={"strain": "string"}
```
victorlin, line 21: _(non-blocking)_

Personally, I'd put `"strain"` in a variable like `strain_col`, since the value is now used 3 times in this file. Would be easier to see where it is used, in case there are ever plans to add support for `"name"` [like `augur.io.read_metadata()`](https://github.com/nextstrain/augur/blob/f68c4488de1ba6f7c9404b21dd2958bae552a736/augur/io.py#L47). But this is obviously fine!

victorlin (approved):

scripts/annotate_metadata_with_index.py
joverlee521: Ah yeah, we should be supporting both 'strain' and 'name' since that's what we say is allowed in the [docs](https://docs.nextstrain.org/projects/ncov/en/latest/guides/data-prep/local-data.html#required-metadata).

huddlej:
CI fails on the "push" because this PR branches from the workflow with the old Nextalign/Nextclade commands, but the CI runs with the latest Docker image using new Nextalign/Nextclade:

```
python3 scripts/sanitize_sequences.py             --sequences data/example_sequences.fasta.gz             --strip-prefixes hCoV-19/ SARS-CoV-2/             --output /dev/stdout 2> logs/sanitize_sequences_gisaid.txt             | nextalign             --jobs=2             --reference defaults/reference_seq.fasta             --genemap defaults/annotation.gff             --genes ORF1a,ORF1b,S,ORF3a,E,M,ORF6,ORF7a,ORF7b,ORF8,N,ORF9b             --sequences /dev/stdin             --output-dir results/translations             --output-basename seqs_gisaid             --output-fasta results/aligned_gisaid.fasta             --output-insertions results/insertions_gisaid.tsv > logs/align_gisaid.txt 2>&1;
        xz -2 -T 2 results/aligned_gisaid.fasta;
        xz -2 -T 2 results/translations/seqs_gisaid*.fasta
```

CI passes on the "pull" which runs the results we'd see after merging this PR into `master` where the alignment commands match the Docker environment.

```
python3 scripts/sanitize_sequences.py             --sequences data/example_sequences.fasta.gz             --strip-prefixes hCoV-19/ SARS-CoV-2/             --output /dev/stdout 2> logs/sanitize_sequences_gisaid.txt             | nextalign run             --jobs=2             --reference defaults/reference_seq.fasta             --genemap defaults/annotation.gff             --output-translations results/translations/seqs_gisaid.gene.{gene}.fasta             --output-fasta results/aligned_gisaid.fasta             --output-insertions results/insertions_gisaid.tsv > logs/align_gisaid.txt 2>&1;
        xz -2 -T 2 results/aligned_gisaid.fasta;
        xz -2 -T 2 results/translations/seqs_gisaid.gene.*.fasta
```

Given that the latter passes and is the check that matters, we should be safe to merge.

scripts/annotate_metadata_with_index.py
huddlej: Thank you both! The original code crashed when the input metadata used "name" instead of "strain" as a column. c6d6b49 fixes the issue by indexing the sequence index data frame by its hardcoded "strain" column (the only value it ever gets) and using a pandas `join` instead of a `merge` to perform the join on the data frame indices. This allows us to avoid knowing the name of the column that the metadata was indexed by and reduces the number of times we refer to "strain".

PR #953, opened 2022-05-25
Memory resource definitions

### LINKED ISSUE #949: Snakemake not passing down resource constrains
**Current Behavior**
When runing nextstrain build with the following command

nextstrain build . --configfile build.yaml --cores all --resources mem_mb=5000

Everything works up until the iqtree2 command is run. Here my OS kills the operation due to its memory limitations.
My current set up has 15gb of ram and when I check the "dmesg" I see the iqtree was using around 19gb of virtual memory which is why the OS killed it.

I am just confused as to why iqtrees is trying to use this much memory when i defined in the resource for snakemake to use only 5gb. It seems as if snakemake is not passing this down.

Have i defined the resources properly? Are there any other steps i can take to prevent this from happening?

image001

tsibley:
See commit messages.

## Related issue(s)
Fixes #949

## Testing
- [ ] CI passes
- [ ] Try to evaluate if the `mem_mb` as defined for `rule tree` is still accurate, or if we need to revise the definition to provide more memory. Passing `--mem` to IQ-TREE may make tree building run slower if the current value is insufficient. Graph various input sizes from our builds vs. memory used from benchmark files for this rule?

## Release checklist
Changes should not be backwards incompatible.

### DIFF
--- a/workflow/snakemake_rules/export_for_nextstrain.smk
+++ b/workflow/snakemake_rules/export_for_nextstrain.smk
@@ -22,6 +22,7 @@
 import re
 import requests
 import json
+from math import ceil
 from workflow.lib.persistent_dict import PersistentDict, NoSuchEntryError
 
 ruleorder: dated_json > finalize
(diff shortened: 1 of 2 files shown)


tsibley:
Another approach to setting a memory limit for the rule (rather than basing it off input size and guessing how much IQ-TREE needs) would be slice up the total memory available to the workflow by the number of concurrent jobs. This would avoid having to keep tabs on the current memory usage in benchmarks and update the rule's resources to match over time.

tsibley:
It would be good to merge this, but I'm wary to do so without first completing the review of memory usage correlation with input size (as described under the "Testing" section above). I will likely hold off on merging until that's done (maybe by me eventually, maybe by someone else interested if anyone).

PR #958, opened 2022-05-29
Fix templating of Nextstrain Auspice JSONs

trvrb:

Matching had been based on `build_name` which with splitting timespans out became `global_6m` rather than `global`. This is causing bugs on the live site, for example https://nextstrain.org/ncov/gisaid/global/6m has map resolution as "division" and lacks map triplicate.

This PR fixes this to split `build_name` into `build_region` and `build_timespan` and match based on `build_region`.

## Testing

I've tested locally so far, though trial builds are running and should be up at 
- https://nextstrain.org/staging/ncov/gisaid/trial/fix-geo-default/global/6m
- https://nextstrain.org/staging/ncov/gisaid/trial/fix-geo-default/oceania/6m
-

### DIFF
--- a/workflow/snakemake_rules/export_for_nextstrain.smk
+++ b/workflow/snakemake_rules/export_for_nextstrain.smk
@@ -140,10 +140,16 @@ rule auspice_config:
         input_set = set(config['inputs'])
         build_name = wildcards.build_name
 
+        if "_" in build_name:
+            build_region, build_timespan = build_name.split("_")
+        else:
+            build_region = build_name
+            build_timespan = ""
+
         ## What are the parameters which vary across builds?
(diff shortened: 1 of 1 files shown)


jameshadfield (approved):
Something to consider (for another PR) is if we want to enumerate the possible `build_regions` which essentially reminds whoever implements another build to update this rule; e.g.:

```py
if build_name in ["reference", "global", "africa", "south-america", "asia", "europe"]:
    default_geo_resolution = "country"
elif build_name in ["north-america", "oceania"]:
    default_geo_resolution = "division"
else:
    raise Exception(f"Unexpected build name '{build_name}' - please update rule 'auspice_config'")
```

But this PR is good to go as is, so I'm going to merge now 😄

trvrb:
Thank you James!

PR #962, opened 2022-06-16
Use Augur's `io.read_metadata`

huddlej:

Replaces all uses of `augur.utils.read_metadata` with the newer (and supported) `augur.io.read_metadata` function. This new function returns a pandas DataFrame instead of a dictionary, so many of the changes here update helper scripts to interact with DataFrames where possible.

## Related issue(s)

Related to https://github.com/nextstrain/augur/pull/934 and https://github.com/nextstrain/augur/issues/970

## Testing

 - Tested combined metadata with two input datasets
 - Confirmed that find clusters and recency annotation scripts produced identical output before and after changes

### DIFF
--- a/scripts/combine_metadata.py
+++ b/scripts/combine_metadata.py
@@ -1,6 +1,5 @@
 import argparse
-from augur.io import open_file
-from augur.utils import read_metadata
+from augur.io import open_file, read_metadata
 from Bio import SeqIO
 import csv
 import sys
@@ -45,7 +44,12 @@ if __name__ == '__main__':
     # READ IN METADATA FILES
     metadata = []
     for (origin, fname) in zip(args.origins, args.metadata):
-        data, columns = read_metadata(fname)
+        data = read_metadata(fname)
+        columns = ["strain"] + data.columns.tolist()
+        data = data.to_dict(orient="index")
+        for strain in data.keys():
+            data[strain]["strain"] = strain
+
         metadata.append({'origin': origin, "fname": fname, 'data': data, 'columns': columns, 'strains': {s for s in data.keys()}})
 
     # SUMMARISE INPUT METADATA
@@ -64,8 +68,8 @@ if __name__ == '__main__':
     for strain in combined_data:
         for column in combined_columns:
             if column not in combined_data[strain]:
-                combined_data[strain][column] = EMPTY    
-    
+                combined_data[strain][column] = EMPTY
+
     for idx in range(1, len(metadata)):
         for strain, row in metadata[idx]['data'].items():
             if strain not in combined_data:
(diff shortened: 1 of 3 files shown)


scripts/combine_metadata.py
```diff
@@ -45,7 +44,12 @@ def parse_args():
     # READ IN METADATA FILES
     metadata = []
     for (origin, fname) in zip(args.origins, args.metadata):
-        data, columns = read_metadata(fname)
+        data = read_metadata(fname)
+        columns = ["strain"] + data.columns.tolist()
+        data = data.to_dict(orient="index")
+        for strain in data.keys():
+            data[strain]["strain"] = strain
```
victorlin, line 51: This can also be accomplished using pandas functions:

```suggestion
        data = read_metadata(fname)
        data = data.rename_axis('strain').reset_index()
        data = data.to_dict(orient="index")
```

scripts/combine_metadata.py
```diff
@@ -64,8 +68,8 @@ def parse_args():
     for strain in combined_data:
         for column in combined_columns:
             if column not in combined_data[strain]:
-                combined_data[strain][column] = EMPTY    
-    
+                combined_data[strain][column] = EMPTY
+
     for idx in range(1, len(metadata)):
         for strain, row in metadata[idx]['data'].items():
             if strain not in combined_data:
```
victorlin, line 75: _(non-blocking)_

Some of this can probably be done more efficiently if data is kept as a DataFrame, but that should be a separate PR.

victorlin (changes requested):

scripts/combine_metadata.py
huddlej: This script needs to be rewritten to handle larger metadata files at some point, anyway, so I figured pandas optimization could happen there.

scripts/combine_metadata.py
huddlej: I tried a similar approach, too, but it doesn't produce quite the same output, since resetting the index makes the index values integers again and then orienting the dict by index produces a dict indexed by those (meaningless) integers. Also, when you read in the file, the index is already named "strain".

The other approach that worked for me was (using data from the `nextstrain-ci` profile):

```python
data = read_metadata("results/europe/metadata_adjusted.tsv.xz")
data.insert(0, "strain", data.index.values)
columns = data.columns
data = data.to_dict(orient="index")
```

I can push this instead, if it seems clearer. It is probably marginally faster, although see other comment about the general need to optimize this script. :D

victorlin: Ohh I see! That looks good.

victorlin (approved):

joverlee521 (approved):

PR #973, opened 2022-07-05
Update emerging lineages

trvrb:

This drops BA.1.1 and BA.2.3 as they are clearly not competitive with BA.4, BA.5 and other recent viruses. This adds BA.2.75 using defining mutations A22001G (S:K147E) and T22016C (S:W152R) to correspond with [Pango designation](https://github.com/cov-lineages/pango-designation/issues/773#issuecomment-1165610030)

## Testing

[Conducting trial run now](https://github.com/nextstrain/ncov/actions/runs/2618119952). Will convert from draft PR when this is complete.


trvrb:
So, the trial run was successful, but no BA.2.75 viruses were subsampled into the Asia / 2m build: https://nextstrain.org/staging/ncov/gisaid/trial/emerging-lineages-update/asia/2m?c=emerging_lineage.

I'd propose to just merge this now and keep tabs on daily updates to make sure it's working as it should. Though, the 147E / 152R signal seems pretty robust and I believe this should work fine.

PR #974, opened 2022-07-06
docs: Switch the build environment from Conda to a virtual env

tsibley:
Everything but make is pip-installable and the Conda environment is
currently having issues on both RTD (where builds are stopped due to
memory usage) and on GitHub Actions (where builds fail in pip during dep
resolution during the Conda env install).¹  I think it's reasonable
enough to assume make is around—it trades an assumption of Conda for an
assumption of Make—and we can always document another way to get make if
that turns out to be a barrier.

¹ https://bedfordlab.slack.com/archives/C01LCTT7JNN/p1657059168195909


tsibley:
Two issues:

- [ ] [RTD build failed](https://readthedocs.org/projects/ncov/builds/17364670/) due to some Sphinx compat issue
- [ ] [CI build failed](https://github.com/nextstrain/ncov/runs/7206198683?check_suite_focus=true) because our shared workflow assumes Conda (and also I didn't updated it)

tsibley:
Merged #975 instead.

PR #975, opened 2022-07-06
docs: Install more Python deps with Pip instead of Conda

tsibley:
Everything but make is pip-installable and the Conda environment is
currently having issues on both RTD (where builds are stopped due to
memory usage) and on GitHub Actions (where builds fail in pip during dep
resolution during the Conda env install).¹  Installing more things with
Pip may help it out.

¹ https://bedfordlab.slack.com/archives/C01LCTT7JNN/p1657059168195909

## Related issue(s)

Alternative to #974.

### DIFF
--- a/docs/conda.yml
+++ b/docs/conda.yml
@@ -3,13 +3,13 @@ channels:
   - defaults
 dependencies:
   - make
-  - recommonmark
-  - requests
-  - sphinx
   - pip
   - pip:
     - nextstrain-sphinx-theme>=2020.2
-    - sphinx-markdown-tables
+    - recommonmark
+    - requests
+    - sphinx
     - sphinx-argparse
-    - sphinx-tabs
     - sphinx-copybutton
+    - sphinx-markdown-tables
+    - sphinx-tabs
--- a/docs/conda.yml
+++ b/docs/conda.yml
@@ -9,6 +9,7 @@ dependencies:
     - recommonmark
     - requests
     - sphinx
+    - docutils<0.18
     - sphinx-argparse
     - sphinx-copybutton
     - sphinx-markdown-tables


tsibley:
We'll see how CI and RTD handle this, but it's not going well so far locally. Pip (via Conda) has been installing for several minutes now, presumably hung up on dep resolution (but I can't see output because Conda has captured it).

tsibley:
Both the CI and RTD builds are currently at 18 minutes and counting during installation/setup.

huddlej (approved):
It looks like this pinning of docutils [allows CI to pass with installation taking ~30 seconds](https://github.com/nextstrain/ncov/runs/7221908948?check_suite_focus=true#step:3:1). At some level, it would be nice to know why the more recent docutils versions cause pip to hang in Conda, but this PR does address what I had originally hoped for which is getting all of the Python-specific deps installed by a single dependency solver within a Conda environment.

tsibley:
+1 for merging this PR then! It would be good to understand what's going on… but at the moment I'm not keen to dive into the dep issues here given other things I'd rather be doing. :upside_down_face:

Some things I noticed though and might as well comment on…

The implication of docutils in some way recalls to mind https://github.com/nextstrain/cli/commit/d2463b72c153ea6720d920baff01003d51ec4c04.

> _But_, when I killed the process, the exception it printed included `/Users/jlhudd/miniconda3/lib/python3.10/site-packages/conda_env/pip_util.py` in the traceback! So, even though I requested Python 3.9 and even though the process shows `python` running from the `ncov-docs` environment directory, Conda is using Python 3.10's pip to install.

:scream: Seems like something (must be Conda??) is mucking with Python library search paths such that 3.9 found 3.10's site-packages, and that's a Very Bad No Good thing. Nothing good can come of it.

> Actually, I don't think there is a Python 3.10/docutils 0.18 conflict, since pip install with Python 3.10 into the Conda environment works just fine:

Conda invokes `pip install` with `--upgrade` (actually `-U` but samesies) but your repro attempt here doesn't include it. The presence/absence of that flag can dramatically change dep resolution, so I wonder if the issue can be repro'd under `--upgrade`?

huddlej:
> Conda invokes `pip install` with `--upgrade` (actually `-U` but samesies)...

Good catch! I saw the `-U` in the running process and my mind mapped it to `--user` instead of `--upgrade`. I just tested the same commands as above but with `-U` and it still ran just fine. 🤷🏻‍♂️

PR #979, opened 2022-07-13
Add Code/Screenshot Reuse Info

emmahodcroft:
I regularly get questions about how to reuse screenshots from Nextstrain (all so far about SC2). People don't seem to be finding the info at the bottom of Nextstrain.org.

I thought adding something to this README might help - the formatting (whether `#` or `##` and whether above or below 'Get in touch' could be changed.
Text is copied directly from what you can see at the bottom of Nextstrain.org.

We may also want to consider adding this to the footer of `ncov` builds for added visibility.

### DIFF
--- a/nextstrain_profiles/nextstrain-gisaid/nextstrain_description.md
+++ b/nextstrain_profiles/nextstrain-gisaid/nextstrain_description.md
@@ -22,3 +22,9 @@ At the specific request of GISAID, we:
  - maintain the prefix `hCoV-19/` in the names of viral isolates
  - disable download of full metadata TSV and provide instead an acknowledgments TSV in the "Download Data" link at the bottom of the page
  - refrain from sharing alignments or other intermediate files computed in our pipeline
+
+#### Reusing code or images
+
+All source code for Auspice, the visualization tool, is freely available under the terms of the [GNU Affero General Public License](https://github.com/nextstrain/auspice/blob/HEAD/LICENSE.txt).
+
+Screenshots may be used under a [CC-BY-4.0 license](https://creativecommons.org/licenses/by/4.0/) and attribution to nextstrain.org must be provided. A high-quality download option is available by clicking the **DOWNLOAD DATA** button at the bottom of the page and selecting **SCREENSHOT (SVG)**.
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -49,6 +49,10 @@ If you're a lab and you'd like to get started sequencing, please see:
 * [Funding opportunities for sequencing efforts](https://twitter.com/firefoxx66/status/1242147905768751106)
 * Or, if these don't meet your needs, [get in touch](mailto:hello@nextstrain.org)
 
+# Reusing code or images
+
+All source code is freely available under the terms of the [GNU Affero General Public License](https://github.com/nextstrain/auspice/blob/master/LICENSE.txt). Screenshots may be used under a [CC-BY-4.0 license](https://creativecommons.org/licenses/by/4.0/) and attribution to nextstrain.org must be provided.
```
victorlin, line 54: It might be better to move this and some other existing sections to the [workflow docs](https://docs.nextstrain.org/projects/ncov/en/latest/), making the README a brief overview of link-outs. I don't see a good place for this in particular, but maybe it could be the start of a new page, something like **Viewer's FAQ** alongside the [**Data Submitter’s FAQ**](https://docs.nextstrain.org/projects/ncov/en/latest/reference/data_submitter_faq.html).

emmahodcroft: My concern with moving it to somewhere else is that already it seems like people aren't finding this information - so I'm trying to make it more visible so that we have to field fewer questions about it. I think people unfortunately put only limited information into looking for this (as they don't really have any interest in the details of the builds or Nextstrain), so I think having it very visible increases the chance they see it. It's also very short, so doesn't take up much room.

victorlin: I don't see the average user on https://nextstrain.org/ncov/open/global/6m finding this section in the repo README. I see these "search paths" in order of likelihood:

1. <kbd>Ctrl</kbd>+<kbd>F</kbd> and search for "screenshot", "download", "image", or some combination of those keywords
2. Click the [DOCS](https://docs.nextstrain.org/en/latest/index.html) or [HELP](https://docs.nextstrain.org/en/latest/learn/about-nextstrain.html) link at the top left and do the same search
3. Go to google.com and do the same search

I think to address visibility, we should definitely put it in the footer of the public builds. Maybe also highlight the SVG download option rather than using an external screenshot tool:

> Looking to download these images? Click the **DOWNLOAD DATA** button at the bottom of the page and select the **SCREENSHOT (SVG)** option. Screenshots may be used under a [CC-BY-4.0 license](https://creativecommons.org/licenses/by/4.0/) and attribution to nextstrain.org must be provided.

victorlin: I'll look into updating this PR with that!

victorlin: Done: https://github.com/nextstrain/ncov/pull/979#event-7192144407

emmahodcroft: This is great, thanks @victorlin !

j23414:
Linking a [recent slack thread](https://bedfordlab.slack.com/archives/C017E82L6FM/p1673972665552209), seems related to this PR.

victorlin:
Merging this since it's a small description change.

emmahodcroft:
Thanks for re-finding this and merging this in @victorlin !

PR #986, opened 2022-07-24
update pango_lineages in color orderings

corneliusroemer:
Sorted list is generated using `utils/sort_lineages.py` script in
cov-lineages/pango-designation repo that produces
`sorted_lineages.txt` output


corneliusroemer:
@victorlin @tsibley docs build still fails here, probably the sphinx issue?

emmahodcroft:
Thanks for the update @corneliusroemer !

PR #993, opened 2022-08-15
implement escape score calculations

jbloom:
Escape scores relative to BA.2 are now calculated (only will be valid for BA.2 descendants).

The scores are calculated using escape calculator [here](https://jbloomlab.github.io/SARS2_RBD_Ab_escape_maps/escape-calc/), using all SARS-CoV-2 elicited antibodies.

@trvrb, can your review and merge? I am also raising GitHub issues for a few related points.


jbloom:
Sorry, meant to make this pull request into `ncov-escape`, not `ncov`.

PR #995, opened 2022-08-24
Update wdl ingest branch

j23414:

Update wdl ingest branch to see if it resolves some github action checks.

### Other Notes

I recommend keeping `wdl/temp_ingest` and `wdl/ncov` branches separate from each other or the list of Terra displayed parameters will become unmanageable. Since they serve different functions, each should be imported into Terra as different workflows.

Ergo, I propose only merging `wdl/ncov` into nextstrain/ncov in the end where `wdl/temp_ingest` will always be a separate branch.


victorlin:
See comments on https://github.com/nextstrain/docker-base/pull/75 for alternatives to this PR.

PR #997, opened 2022-08-24
Docs/cores

j23414:

Documentation changes according to https://github.com/nextstrain/measles/pull/6#issuecomment-1210990071

### DIFF
--- a/docs/src/tutorial/example-data.rst
+++ b/docs/src/tutorial/example-data.rst
@@ -41,7 +41,7 @@ From within the ``ncov/`` directory, run the workflow using a :term:`configurati
 
 .. code:: text
 
-   nextstrain build . --cores all --configfile ncov-tutorial/example-data.yaml
+   nextstrain build . --configfile ncov-tutorial/example-data.yaml
 
 Break down the command
 ~~~~~~~~~~~~~~~~~~~~~~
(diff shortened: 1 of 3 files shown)


docs/src/tutorial/example-data.rst
```diff
@@ -51,7 +51,7 @@ The workflow can take several minutes to run. While it is running, you can learn
 - ``nextstrain build .``
    - This tells the :term:`docs.nextstrain.org:Nextstrain CLI` to :term:`build <docs.nextstrain.org:build (verb)>` the workflow from ``.``, the current directory. All subsequent command-line parameters are passed to the workflow manager, Snakemake.
 - ``--cores all``
-   - This required Snakemake parameter specifies the number of CPU cores to use (:doc:`more info <snakemake:executing/cli>`).
+   - This optional parameter (not shown) can be passed to Snakemake specifying the number of CPU cores to use (:doc:`more info <snakemake:executing/cli>`). Nextstrain will pass a `--cores all` by default.
```
victorlin, line 54: I'd be in favor of removing this entirely to keep the tutorial simple. It's an additional detail that users can look up if they want, among other optional Snakemake parameters.

victorlin (changes requested):

docs/src/tutorial/example-data.rst
j23414: Thanks! Normally I'd do a "Commit suggestion" but hit an "Applying suggestion on deleted lines is not currently supported. 

Screen Shot 2022-08-25 at 3 26 54 PM

So I added the changes in a separate commit.

victorlin: I went around the disabled suggestions button by manually adding an empty suggestions block:

    ```suggestions
    ```

It rendered fine which was the main intent.

I think it'd be good to use a more descriptive title for 6b4201d1a3eda612f22300acf60cffa1f28c539b or even better, squash it into https://github.com/nextstrain/ncov/pull/997/commits/3095b26da23cd020335658857ca44fceb00dc77e.

victorlin (approved):

PR #1000, opened 2022-09-12
Update workflow for new augur clades version

jameshadfield:
This updates the workflow to use augur after augur PR https://github.com/nextstrain/augur/pull/728 has been merged. It removes two rules from the workflow.

In the situation where the ncov workflow is not updated but augur is, the pipeline will proceed without error however the 'emerging_lineage' branch label will be missing. (This is because the add_branch_labels.py script is looking for the key 'clade_annotation' which is no longer present.)

In the situation where the ncov workflow is updated but augur is not, rule.emerging_lineages will error due to unknown arguments. You can see this in the failing CI for this PR 😉 

The script add_branch_labels.py is no longer used, but not removed here, as it contains logic to export spike mutations as branch labels which may be useful at some point. If we do use this, it would be better to produce an intermediate node-data JSON with a custom branch label to avoid modifying the auspice JSON after export.

## Release checklist


jameshadfield:
Rebased and re-tested using the (now released) Augur v22.0.0

Note there is one change here which is to be fixed on Augur's side - the clade coloring entry is now the bottom entry in the (Auspice) dropdown rather than the top. PR for that forthcoming.

jameshadfield:
CI Failures on conda: `ERROR: results/europe/rbd_levels.json did not contain either `nodes` or `branches`. Please check the formatting of this JSON!` This has been fixed by https://github.com/nextstrain/augur/pull/1214 (merged) but this isn't yet in an Augur release.

CI Failures on Docker: `augur: error: unrecognized arguments: --membership-name emerging_lineage --label-name emerging_lineage`. This is related to the Docker caching issues (I think) resulting in Docker not yet using Augur v22.

jameshadfield:
Updated to require minimum augur version of 22.0.1

jameshadfield:
Force pushed (with no changes) to trigger a new CI run, now that augur 22.0.1 [is available in conda-base](https://github.com/nextstrain/conda-base/actions/runs/4997526487)

workflow/snakemake_rules/main_workflow.smk
```diff
@@ -1462,28 +1446,10 @@ rule export:
...
-        auspice_json = rules.export.output.auspice_json,
-        emerging_clades = rules.emerging_lineages.output.clade_data
-    output:
-        auspice_json = "results/{build_name}/ncov_with_branch_labels.json"
-    log:
-        "logs/add_branch_labels{build_name}.txt"
-    conda: config["conda_environment"]
-    shell:
-        """
-        python3 ./scripts/add_branch_labels.py \
```
huddlej, line 1477: Nice to see these temp rules deleted! Can we also delete `add_branch_labels.py` now that we don't use it anywhere else? We could always revive it later from Git, if we needed it...

jameshadfield: [Done](https://github.com/nextstrain/ncov/compare/21f88a250da3e30191fee54ac946b12128b86dd4..322b8a9076b93c676b062909fa37f1905a665161)

For reference (from the updated commit message):
> Note that [the script] contained unused functionality to export spike mutations; if we reinstate this in the future we should update the output format to produce a node-data JSON with a custom branch label to avoid modifying the auspice JSON after export.

PR #1010, opened 2022-10-05
Add `Nextclade_pango` to colours & filters in Auspice

emmahodcroft:

Nextclade pango-assigned lineage is sometimes faster at incorporating new lineages of interest (and also is guaranteed to assign these phylogenetically), so having these designations on the tree is useful for seeing them.

I have tried to do this so it applies to all Nextstrain builds (GISAID, open, etc) but would be good if someone can double-check that this should work.

## Related issue(s)

Should resolve #919 (?)

## Testing

~~UNTESTED~~
Edit: now tested via the test builds but not full builds - see below

### DIFF
--- a/workflow/snakemake_rules/export_for_nextstrain.smk
+++ b/workflow/snakemake_rules/export_for_nextstrain.smk
@@ -191,6 +191,11 @@ rule auspice_config:
                     "title": "Pango Lineage",
                     "type": "categorical"
                 },
+                {
+                    "key": "Nextclade_pango",
+                    "title": "Nextclade Pango Lineage",
+                    "type": "categorical"
+                },
                 gisaid_clade_coloring,
                 {
                     "key": "S1_mutations",
@@ -281,6 +286,7 @@ rule auspice_config:
                 "clade_membership",
                 "emerging_lineage",
                 "pango_lineage",
+                "Nextclade_pango",
                 "region",
                 "country",
                 "division",


corneliusroemer:
Thanks! I've started test builds via Github actions!

corneliusroemer:
So the CI test run works 🎉 - ridiculous that it's so easy! We should have done that way earlier already 🙃

image

corneliusroemer (approved):
Seems to work!

emmahodcroft:
Great! Thanks for starting the test. Having added these yesterday to CoV made me more confident even though the `auspice_config` files are different (mine is just a straight up config, here it's modified for Nextstrain) - so that was the part I was a little unsure about.

corneliusroemer:
Trial build, maybe ready in 4 hours:
https://nextstrain.org/staging/ncov/gisaid/trial/nextclade-gisaid-correct-branch/europe

jameshadfield (approved):
Working well.

The colours aren't great as they are automatically generated for auspice. We have some scripts to order these, but they're prone to becoming out of date which causes lots of greyscale tips (e.g. see pango lineage colouring in the trial build below). Let's solve that elsewhere!

image

corneliusroemer:
@jameshadfield are you sure you tried the right coloring? Nextclade_pango does not have grey, as it is not in color_ordering.

Having more or less random colorings is potentially a feature as it allows to spot where new lineages are more easily than if BQ.1 and BQ.1.1 were inly 0.1% different in color space - impossible to notice.

corneliusroemer:
You actually seem to have shown pango lineage, not Nextclade Pango Lineage. This is coloring for Nextclade Pango Lineage:

image
https://nextstrain.org/staging/ncov/gisaid/trial/nextclade-gisaid-correct-branch/global/2m?c=Nextclade_pango

PR #1014, opened 2022-10-12
Remove WH01 as reference

trvrb:

In very early versions of the workflow we needed to root by calling `augur refine --root Wuhan/Hu-1/2019 Wuhan/WH01/2019`. However, this strategy hasn't been used since June 2020 and we [currently default](https://github.com/nextstrain/ncov/blob/master/defaults/parameters.yaml#L122) to `--root Wuhan/Hu-1/2019`.

Thus, we still have Wuhan/WH01/2019 hanging out in force include and in references without any purpose. This commit removes extraneous references to Wuhan/WH01/2019.

It will change behavior to no longer force include Wuhan/WH01/2019 in trees, but I think this is as it should be.

## Testing

I've tested locally and am running GISAID / open trial builds. When this completes, results will be available at [nextstrain.org/staging/ncov/gisaid/trial/remove-WH01-reference/global/2m](https://nextstrain.org/staging/ncov/gisaid/trial/remove-WH01-reference/global/2m), etc... as well as [nextstrain.org/staging/ncov/open/trial/remove-WH01-reference/global/2m](https://nextstrain.org/staging/ncov/open/trial/remove-WH01-reference/global/2m).

## Release checklist


defaults/include.txt
```diff
@@ -1,3 +1,2 @@
 Wuhan/Hu-1/2019
 Wuhan-Hu-1/2019
```
victorlin, line 2: I was curious about the difference between `Wuhan/Hu-1/2019` vs. `Wuhan-Hu-1/2019`, and saw it's explained here: 429a1806e395a90e392ea27d588eb7aea017839a

victorlin (approved):

defaults/include.txt
trvrb: Yeah. Really datasets shouldn't ever have both, but it's still annoying to deal with.

trvrb:
Thanks for reviewing Victor. I'm going to go ahead and merge this.

PR #1016, opened 2022-10-17
Update filter docs

joverlee521:
Miscellaneous updates to docs for filters.
Hopefully can help with some of the common questions we get about "why are X sequences excluded?"
* Troubleshooting docs
    * link to workflow config filter references docs
    * link to default parameters.yaml file
    * point users to `filtered_log.tsv` output of filter rule (I couldn't find any reference to this file in the docs!)
* Workflow config reference docs
    * Note that `min_length` is applied to **masked** sequences


j23414 (approved):
Nice! And particularly liked

> The ``min_length`` filter is applied to the **masked** sequences, not the original input sequences.

PR #1021, opened 2022-10-21
Update data docs

joverlee521:
* move `aligned.fasta.xz` to list of `ncov-ingest` produced files since
this is now output from Nextclade in the daily ingest pipeline
* remove `mutation-summary.tsv.xz` since this is no longer updated in
the daily ingest pipeline
* Add `zst` files to the remote input list with warning that the files cannot be directly used as inputs for the pipeline.

### DIFF
--- a/docs/src/reference/remote_inputs.rst
+++ b/docs/src/reference/remote_inputs.rst
@@ -24,12 +24,12 @@ All available genomes and metadata
 
 Entire metadata & sequences data is uploaded from the ``ncov-ingest`` workflows for each of the ``gisaid`` and ``open`` sources:
 
--  ``metadata.tsv.gz``
--  ``sequences.fasta.xz``
--  ``nextclade.tsv.gz``
--  ``aligned.fasta.xz`` (Alignment via `Nextclade <https://docs.nextstrain.org/projects/nextclade/en/stable/user/output-files.html#aligned-nucleotide-sequences>`__. The default reference genome is `MN908947 <https://www.ncbi.nlm.nih.gov/nuccore/MN908947>`__ (Wuhan-Hu-1))
--  ``additional_info.tsv.gz`` (GISAID only)
--  ``flagged_metadata.txt.gz`` (GISAID only)
+-  ``metadata.tsv.gz`` and ``metadata.tsv.zst``
+-  ``sequences.fasta.xz`` and ``sequences.fasta.zst``
+-  ``nextclade.tsv.gz`` and ``nextclade.tsv.zst``
+-  ``aligned.fasta.xz`` and ``aligned.fasta.zst`` (Alignment via `Nextclade <https://docs.nextstrain.org/projects/nextclade/en/stable/user/output-files.html#aligned-nucleotide-sequences>`__. The default reference genome is `MN908947 <https://www.ncbi.nlm.nih.gov/nuccore/MN908947>`__ (Wuhan-Hu-1))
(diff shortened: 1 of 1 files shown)


victorlin (approved):
Looks good, just a small update.

docs/src/reference/remote_inputs.rst
```diff
@@ -54,6 +54,9 @@ Each regional build (``global``, ``africa``, ``asia``, ``europe``, ``north-ameri
 
    Please note that these files are uploaded in two batches (see above for details). This means that the full GenBank metadata and sequences are typically updated a couple of hours before the more processed files.
 
+.. warning::
+  The zstandard (zstd) files are not yet supported as direct inputs for the pipeline.
```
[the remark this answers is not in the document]
joverlee521: Hmm, looks like they didn't do a release post-merge yet...

We should probably still add a note to users that they need to use a the specific version of `xopen`. Since the ncov sanitize scripts are using `augur.io`, we should probably update the pin of `xopen` version in Augur and do an Augur release. Then we can update the docs here to say users must be using the Augur version XX.X.X to use zst inputs.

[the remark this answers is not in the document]
joverlee521: I created https://github.com/nextstrain/augur/issues/1073 we can discuss specifics there.

docs/src/reference/remote_inputs.rst
```diff
@@ -24,12 +24,12 @@ All available genomes and metadata
...
--  ``nextclade.tsv.gz``
--  ``aligned.fasta.xz`` (Alignment via `Nextclade <https://docs.nextstrain.org/projects/nextclade/en/stable/user/output-files.html#aligned-nucleotide-sequences>`__. The default reference genome is `MN908947 <https://www.ncbi.nlm.nih.gov/nuccore/MN908947>`__ (Wuhan-Hu-1))
--  ``additional_info.tsv.gz`` (GISAID only)
--  ``flagged_metadata.txt.gz`` (GISAID only)
+-  ``metadata.tsv.gz`` / ``metadata.tsv.zst``
+-  ``sequences.fasta.xz`` / ``sequences.fasta.zst``
+-  ``nextclade.tsv.gz`` / ``nextclade.tsv.zst``
+-  ``aligned.fasta.xz`` / ``aligned.fasta.zst`` (Alignment via `Nextclade <https://docs.nextstrain.org/projects/nextclade/en/stable/user/output-files.html#aligned-nucleotide-sequences>`__. The default reference genome is `MN908947 <https://www.ncbi.nlm.nih.gov/nuccore/MN908947>`__ (Wuhan-Hu-1))
+-  ``additional_info.tsv.gz`` / ``additional_info.tsv.zst`` (GISAID only)
+-  ``flagged_metadata.txt.gz`` / ``flagged_metadata.txt.zst`` (GISAID only)
```
tsibley, line 32: Maybe use ` and ` or `, ` for the separator instead of ` / ` so it's not confusable with a path? I had to do a double take.

image

joverlee521: Ah, didn't think of `/` as a path separator, just a shorthand for `or`. I'll update!

joverlee521: Updated in [force-push](https://github.com/nextstrain/ncov/compare/bf6bf933ef70645094d8a08cc3197739583a1c37..df80a63c4d46effaf182bbf4b55befc917a1c298).

tsibley: grazie!

PR #999, opened 2022-09-09
Update: TERRA Workflow Documentation

j23414:
## Description of proposed changes

Since occasional questions about our Terra workflows have been raised during office hours or via DM's, the goal is to update documentation of 3 WDL Terra workflows:

1. [ncov:master](https://dockstore.org/workflows/github.com/nextstrain/ncov:master?tab=info) - run the basic ncov workflow
2. [ncov:wdl/genbank_ingest](https://dockstore.org/workflows/github.com/nextstrain/ncov:wdl/genbank_ingest?tab=info) - pull a public dataset and send them through our preprocessing scripts.
3. [ncov:wdl/gisaid_ingest](https://dockstore.org/workflows/github.com/nextstrain/ncov:wdl/gisaid_ingest?tab=info) - pull a private dataset if a user has their own API key, account, and password. Mostly to make available our preprocessing scripts.

The workflows are separated so that only parameters specific to a particular usecase are shown in Terra.

## Related issue(s)

Related to #https://github.com/nextstrain/private/issues/11

## Testing

Since this PR should only change the rst files (documentation), testing mainly consist of ensuring that the documents build. Any problematic steps are being fixed if/when they arise with each onboarded user. Comments or suggestions are still welcome though.

Linking the changed pages for ease of proof-reading:

* [run-ingest-on-terra.rst](https://github.com/nextstrain/ncov/blob/wdl/docs/docs/src/guides/run-ingest-on-terra.rst)
* [run-analysis-on-terra.rst](https://github.com/nextstrain/ncov/blob/wdl/docs/docs/src/guides/run-analysis-on-terra.rst)

Or the live draft pages:

* [run-analysis-on-terra](https://nextstrain--999.org.readthedocs.build/projects/ncov/en/999/guides/run-analysis-on-terra.html)
* [run-genbank-ingest-on-terra](https://nextstrain--999.org.readthedocs.build/projects/ncov/en/999/guides/run-genbank-ingest-on-terra.html)
* [run-gisaid-ingest-on-terra](https://nextstrain--999.org.readthedocs.build/projects/ncov/en/999/guides/run-gisaid-ingest-on-terra.html)

PR #1024, opened 2022-10-27
fix: select_first was called with 1 empty values

j23414:

See commit for details. Apparently prior tests had always been passing a nextclade cache, so this bug did not surface earlier. Thanks @victorlin !

## Related to

* https://github.com/nextstrain/ncov/pull/999


j23414:
* [x] Test in Terra to see if error still occurs. 

Job started without the error, so I cancelled the job preemptively.

PR #1025, opened 2022-11-01
feat: deploy builds eagerly, not wait until all done

corneliusroemer:

Currently we only deploy builds once all builds are done.

Especially when testing, this causes unnecessarily large delays: some builds can be done as soon as 1.5hr after a build is started, but others may take 12hr.

This PR adds an intermediate rule that deploys a single build as soon as all files are ready. This was discussed briefly on Slack and @trvrb gave green light for this change: https://bedfordlab.slack.com/archives/CSKMU6YUC/p1666885515151009

Changes are minimal, I tested this using dry run, will make a test run, this should work well and help test things faster in the future.

To make it easier to check builds that have finished, I've added Slack messages including the URL for each successful deployment. We already output a lot of files anyways and having clickable URLs is quite useful.

## Testing

Test builds:
```
nextstrain build --aws-batch --no-download --attach 484fdc4c-4465-4f99-a791-6d991bfcee17 .
nextstrain build --aws-batch --no-download --attach 30b12b38-16e1-4f92-b7e3-761b282c8c14 .
```

https://nextstrain.org/staging/ncov/gisaid/trial/eager/global/all-time
https://nextstrain.org/staging/ncov/open/trial/eager/global/all-time

The first build was already deployed after less than 90min (reference, open)

### DIFF
--- a/workflow/snakemake_rules/export_for_nextstrain.smk
+++ b/workflow/snakemake_rules/export_for_nextstrain.smk
@@ -368,16 +368,39 @@ except:
     # means that the Snakefile won't crash.
     deploy_origin = "by an unknown identity"
 
+
+rule deploy_single:
+    input:
+        # Replication of all_regions but with build_name as wildcard
+        auspice_json = expand("auspice/{prefix}_{{build_name}}.json", prefix=config["auspice_json_prefix"]),
+        tip_frequencies_json = expand("auspice/{prefix}_{{build_name}}_tip-frequencies.json", prefix=config["auspice_json_prefix"]),
+        root_sequence_json = expand("auspice/{prefix}_{{build_name}}_root-sequence.json", prefix=config["auspice_json_prefix"]),
+        dated_auspice_json = expand("auspice/{prefix}_{{build_name}}_{date}.json", prefix=config["auspice_json_prefix"], date=config.get("build_date", get_todays_date())),
+        dated_tip_frequencies_json = expand("auspice/{prefix}_{{build_name}}_{date}_tip-frequencies.json", prefix=config["auspice_json_prefix"], date=config.get("build_date", get_todays_date())),
+        dated_root_sequence_json = expand("auspice/{prefix}_{{build_name}}_{date}_root-sequence.json", prefix=config["auspice_json_prefix"], date=config.get("build_date", get_todays_date()))
+    output:
+        temp(touch("auspice/{build_name}.deployed"))
+    params:
+        deploy_url = config["deploy_url"],
+        prefix = config["auspice_json_prefix"]
+    benchmark:
+        "benchmarks/deploy_single_{build_name}.txt"
+    run:
+        shell("nextstrain deploy {params.deploy_url:q} {input:q}")
+        if params.deploy_url == "s3://nextstrain-data":
+            slack_message = f"Deployed {wildcards.build_name} to https://nextstrain.org/{(params.prefix + '_' + wildcards.build_name).replace('_', '/')}"
+        else:
+            slack_message = f"Deployed {wildcards.build_name} to {params.deploy_url}"
+        send_slack_message(slack_message)
+
 rule deploy:
     input:
-        *rules.all_regions.input
+        expand("auspice/{build_name}.deployed", build_name=BUILD_NAMES),
     params:
         slack_message = f"Deployed to {config['deploy_url']} {deploy_origin}",
-        deploy_url = config["deploy_url"]
     benchmark:
         "benchmarks/deploy.txt"
     run:
-        shell("nextstrain deploy {params.deploy_url:q} {input:q}")
         send_slack_message(params.slack_message)
 
 rule upload:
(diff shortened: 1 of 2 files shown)


workflow/snakemake_rules/export_for_nextstrain.smk
```diff
@@ -367,16 +367,41 @@ except:
...
+        else:
+            slack_message = f"Deployed {wildcards.build_name} to {params.deploy_url}"
+        send_slack_message(slack_message)
+
 rule deploy:
     input:
-        *rules.all_regions.input
+        expand("auspice/{build_name}.deployed", build_name=BUILD_NAMES),
+    benchmark:
+        "benchmarks/deploy.txt"
```
victorlin, line 399: Why is this duplicated from a few lines below?

workflow/snakemake_rules/export_for_nextstrain.smk
```diff
@@ -367,16 +367,41 @@ except:
...
+        if params.deploy_url == "s3://nextstrain-data":
+            slack_message = f"Deployed {wildcards.build_name} to https://nextstrain.org/{(params.prefix + '_' + wildcards.build_name).replace('_', '/')}"
+        else:
+            slack_message = f"Deployed {wildcards.build_name} to {params.deploy_url}"
+        send_slack_message(slack_message)
+
 rule deploy:
     input:
-        *rules.all_regions.input
+        expand("auspice/{build_name}.deployed", build_name=BUILD_NAMES),
```
victorlin, line 397: How does this work? There is a separate `deploy` for every item in `BUILD_NAMES`?

workflow/snakemake_rules/export_for_nextstrain.smk
```diff
@@ -367,16 +367,41 @@ except:
...
+        "benchmarks/deploy_single_{build_name}.txt"
+    run:
+        shell("nextstrain deploy {params.deploy_url:q} {input:q}")
+        if params.deploy_url == "s3://nextstrain-data":
+            slack_message = f"Deployed {wildcards.build_name} to https://nextstrain.org/{(params.prefix + '_' + wildcards.build_name).replace('_', '/')}"
+        else:
+            slack_message = f"Deployed {wildcards.build_name} to {params.deploy_url}"
+        send_slack_message(slack_message)
+
 rule deploy:
```
victorlin, line 395: This would be more clear to me:

```suggestion
rule deploy_all:
```

victorlin (commented):
I don't know enough about Snakemake to give a hard approval, but looks like you've tested successfully so consider this a soft approval. I just have a few non-blocking comments.

workflow/snakemake_rules/export_for_nextstrain.smk
corneliusroemer: The `rule deploy` is the one that requests deployment of all builds.

Previously it asked for all files, and then deployed them itself.

Now it asks for a touchfile `auspice/{build_name}.deployed` instead, which causes the `rule deploy_single` to deploy one build at a time.

The speed up happens because previously deploy only happened once all input files = all builds were ready.

Now the actual deployments can happen as soon as all build files for a single build are done.

workflow/snakemake_rules/export_for_nextstrain.smk
corneliusroemer: It would be clearer but it would be a breaking change - as we explicitly request the `deploy` rule in some Github actions.

So rather than `deploy_all`/`deploy_single` it's less breaking to call them `deploy`/`deploy_single`

workflow/snakemake_rules/export_for_nextstrain.smk
```diff
@@ -367,16 +367,41 @@ except:
...
     input:
-        *rules.all_regions.input
+        expand("auspice/{build_name}.deployed", build_name=BUILD_NAMES),
+    benchmark:
```
corneliusroemer, line 398:
```suggestion
```

workflow/snakemake_rules/export_for_nextstrain.smk
corneliusroemer, line 399:
```suggestion
```

tsibley (commented):
Code looks good by inspection to me.

The branch history is one primary commit and then four fixup commits + a changelog update commit. Would be nice to rebase those fixups away into the first commit before merging.

corneliusroemer:
I'll rebase - never done this before so bear with me 🙃

victorlin:
Looks like there's a small changelog conflict too, so you'll have to update with latest master to resolve the conflict.

Something like this should do it:

```sh
# Update latest master
git fetch origin master:master

# Switch to the working branch if you're not there already
git switch earger-deploy

# This will open a rebase prompt, and you can mark fixup commits accordingly.
git rebase -i master

# Resolve merge conflicts (note the date format consistency fix in changelog)

git push --force-with-lease
```

corneliusroemer:
Thanks @victorlin - I used fork in the end, hope it's good now.  Looks good to me at least :)

PR #1032, opened 2022-11-28
Generate a representative 100k sample

jameshadfield:
This sample (metadata + sequences) is useful for local testing / dev work. It is not expensive to compute this weekly, so GitHub actions is set to run each monday. No timestamps are used as it's not intended to be accessed retrospectively.


jameshadfield:
GitHub action is unavailable as it's introduced on this branch. AWS build (triggered locally) ongoing: `AWS Batch Job ID: 05c94df1-a0ca-4209-92a1-6714aa89db58`

nextstrain_profiles/100k/config.yaml
```diff
@@ -0,0 +1,33 @@
+
+S3_DST_BUCKET: "nextstrain-ncov-private/100k/"
```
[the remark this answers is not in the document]
jameshadfield: Fixed by 329b2f66 (force push)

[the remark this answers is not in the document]
jameshadfield: (And the files have been moved to their correct S3 location manually.)

.github/workflows/rebuild-100k.yml
```diff
@@ -0,0 +1,57 @@
...
+  schedule:
+    - cron:  '* * * * 1'
+  # Manually triggered using GitHub's UI
+  workflow_dispatch:
+
+jobs:
+  gisaid:
+    runs-on: ubuntu-20.04
+    steps:
+    - uses: actions/checkout@v2
```
corneliusroemer, line 14:
```suggestion
    - uses: actions/checkout@v3
```

v2 is deprecated soon

.github/workflows/rebuild-100k.yml
```diff
@@ -0,0 +1,57 @@
...
+on:
+  # cron job once a week on mondays
+  schedule:
+    - cron:  '* * * * 1'
+  # Manually triggered using GitHub's UI
+  workflow_dispatch:
+
+jobs:
+  gisaid:
+    runs-on: ubuntu-20.04
```
corneliusroemer, line 12:
```suggestion
    runs-on: ubuntu-latest
```

.github/workflows/rebuild-100k.yml
```diff
@@ -0,0 +1,57 @@
...
+    steps:
+    - uses: actions/checkout@v2
+
+    - uses: nextstrain/.github/actions/setup-nextstrain-cli@master
+      with:
+        python-version: 3.9
```
corneliusroemer, line 18:
```suggestion
        python-version: 3.10
```

Should work with 3.10 IIRC

.github/workflows/rebuild-100k.yml
```diff
@@ -0,0 +1,57 @@
...
+        set -x
+
+        declare -a config
+        config+=(slack_token=$SLACK_TOKEN)
+
+        nextstrain build \
+          --aws-batch \
+          --detach \
+          --cpus 16 \
+          --memory 14GiB \
```
corneliusroemer, line 31: Don't we automatically get a lot more memory when we specify 16 CPUs?

Could future proof by allowing as much memory as we get for the CPUs

.github/workflows/rebuild-100k.yml
```diff
@@ -0,0 +1,57 @@
...
+        AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+        SLACK_TOKEN: ${{ secrets.SLACK_TOKEN }}
+
+    - name: Build info
+      run: |
+        echo "--> 100k sample rebuilding on AWS"
+        echo
+        echo "--> When completed, the following 2 files will be updated:"
+        echo "s3://nextstrain-ncov-private/100k/metadata.tsv.xz"
+        echo "s3://nextstrain-ncov-private/100k/sequences.fasta.xz"
```
corneliusroemer, line 49: We should either only or additionally use `zst` rather than `xz` it (de)compresses much faster

nextstrain_profiles/100k/README.md
```diff
@@ -0,0 +1,21 @@
...
+This is intended to run weekly via a GitHub action (which triggers a job to be run on AWS).
+It will make two files available:
+
+* `s3://nextstrain-ncov-private/100k/metadata.tsv.xz`
+* `s3://nextstrain-ncov-privatr/100k/sequences.fasta.xz`
+
+While this profile is not recommended to be run locally, you can see what rules would be run via:
+
+```
+snakemake --cores 1 --configfile nextstrain_profiles/100k/config.yaml -npf upload --dag | dot -Tpdf > dag.pdf
```
corneliusroemer, line 13:
```suggestion
snakemake -c1 --configfile nextstrain_profiles/100k/config.yaml -npf upload --dag | dot -Tpdf > dag.pdf
```

Can shorten ;)

nextstrain_profiles/100k/config.yaml
```diff
@@ -0,0 +1,33 @@
...
+subsampling:
+  100k_scheme:
+    50k_early:
+      group_by: "year month region"
+      max_sequences: 50000
+      max_date: "--max-date 1Y"
+    50k_late:
+      group_by: "year month region"
+      max_sequences: 50000
+      min_date: "--min-date 1Y"
```
corneliusroemer, line 33: Does this produce clean boundaries? Or will a certain month be (partially) double sampled?

.github/workflows/rebuild-100k.yml
corneliusroemer: Ah I see, since this runs in ncov rather than ingest this may not be so easy as ncov isn't yet `zstd` ready I suppose.

corneliusroemer (commented):
Looks good, I suppose there's no harm in setting this up and test/improve as we go.

nextstrain_profiles/100k/config.yaml
jameshadfield: I'll say that this sample isn't supposed to be truly representative for proper analyses -- it's just a way to test things locally and get an idea of what the output would look like.

.github/workflows/rebuild-100k.yml
jameshadfield: Yeah -- the intermediate files are already xz compressed, this is simply uploading them. zstd would be a bigger change to the pipeline internals.

nextstrain_profiles/100k/config.yaml
trvrb: I think if we get subsampling right, the "100k" dataset should produce nearly indistinguishable output for ncov global 6m (for example) to using the entire 14M sequence database.

nextstrain_profiles/100k/config.yaml
```diff
@@ -0,0 +1,33 @@
...
+    subsampling_scheme: 100k_scheme
+
+# remove sequences without division label in US
+filter:
+  exclude_where: "division='USA'"
+
+subsampling:
+  100k_scheme:
+    50k_early:
+      group_by: "year month region"
```
trvrb, line 27: I suspect that `year month country` will give a better basis for further analysis, as Europe will be dominated by Germany and the UK, Africa will be dominated by South Africa and North America will be dominated by the US.

nextstrain_profiles/100k/config.yaml
victorlin: This should produce clean boundaries since `1Y` translates to [an exact day](https://github.com/nextstrain/augur/blob/a2d7660dd1c2c61814ce9291efb9cc87c0fe2567/augur/dates.py#L55).

Both `--min-date` and `--max-date` are [inclusive](https://github.com/nextstrain/augur/blob/a2d7660dd1c2c61814ce9291efb9cc87c0fe2567/augur/filter.py#L384-L388) so there may be double-sampling of that day, but no more than that.

jameshadfield: > This should produce clean boundaries since 1Y translates to [an exact day](https://github.com/nextstrain/augur/blob/a2d7660dd1c2c61814ce9291efb9cc87c0fe2567/augur/dates.py#L55).

Assuming the month that falls in is `X`, that will leave some samples from `X` in both samples (`50k_early`, `50k_late`). This'll partition things evenly according to month, so in `50k_early` there'll be `~50,000*1/24` from `X` and in `50k_late` there'll be `~50,000*1/12` samples. Won't matter for the 6M build, but a little less than ideal.

victorlin: Oh you're right, yes there will be double sampling of the month due to `--group-by month`.

The behavior with the exact date cutoff is worth pointing out. For example:

- `--max-date 2022-03-08 --group-by month` will sample the first week of March as much as each of the previous full months.
- `--min-date 2022-03-08 --group-by month` will sample the remainder of March as much as each of the following full months.

nextstrain_profiles/100k/README.md
```diff
@@ -0,0 +1,21 @@
+## Aim
+
+To build a representative 100k dataset which is available for testing / developing builds locally.
+This is intended to run weekly via a GitHub action (which triggers a job to be run on AWS).
+It will make two files available:
+
+* `s3://nextstrain-ncov-private/100k/metadata.tsv.xz`
+* `s3://nextstrain-ncov-privatr/100k/sequences.fasta.xz`
```
tsibley, line 8:
```suggestion
* `s3://nextstrain-ncov-private/100k/sequences.fasta.xz`
```

nextstrain_profiles/100k/README.md
```diff
@@ -0,0 +1,21 @@
...
+
+While this profile is not recommended to be run locally, you can see what rules would be run via:
+
+```
+snakemake --cores 1 --configfile nextstrain_profiles/100k/config.yaml -npf upload --dag | dot -Tpdf > dag.pdf
+```
+
+To run manually you can trigger the GitHub action or trigger the job locally via:
```
tsibley, line 16:
```suggestion
To run manually you can trigger the GitHub action or run the job locally via:
```

.github/workflows/rebuild-100k.yml
```diff
@@ -0,0 +1,57 @@
+name: Rebuild 100k sample
+
+on:
+  # cron job once a week on mondays
+  schedule:
+    - cron:  '* * * * 1'
```
tsibley, line 6: This schedule is actually every minute on Mondays. :grimacing: We want something like this instead:

```suggestion
  # cron job once a week on Mondays at 12:42 UTC
  schedule:
    - cron:  '42 12 * * 1'
```

You might even be able to use `Mon` instead of `1`, but GitHub's docs aren't clear if they support that or not (though there's a suggestion they do).

workflow/snakemake_rules/common.smk
```diff
@@ -223,6 +223,13 @@ def _get_upload_inputs(wildcards):
 
     origin = config["S3_DST_ORIGINS"][0]
 
+    # the 100k sample is unique in that we only want to upload 2 files. Note that the S3_DST_ORIGINS will specify the directory
+    if len(config["builds"].keys())==1 and "100k" in config["builds"]:
+        return {
+            f"metadata.tsv.xz": "results/100k/100k_subsampled_metadata.tsv.xz",
+            f"sequences.fasta.xz": "results/100k/100k_subsampled_sequences.fasta.xz"
+        }
```
tsibley, line 231: _non-blocking_

This is a bit hackish. I wonder about letting the config optionally specify the set of files to upload instead.

tsibley (changes requested):

workflow/snakemake_rules/common.smk
tsibley: For your consideration: https://github.com/nextstrain/ncov/commit/595b96367d928d08945bd7ce3fc2f0d1bbe911e3

.github/workflows/rebuild-100k.yml
jameshadfield: You're right. Our [nextstrain-job queue](https://us-east-1.console.aws.amazon.com/batch/home?region=us-east-1#queues/detail/arn:aws:batch:us-east-1:827581582529:job-queue/nextstrain-job-queue) uses the [c5-instances-2023-01-17](https://us-east-1.console.aws.amazon.com/batch/home?region=us-east-1#compute-environments/ec2/detail/arn:aws:batch:us-east-1:827581582529:compute-environment/c5-instances-2023-01-17) compute environment which uses [c5 instance types](https://aws.amazon.com/ec2/instance-types/). So 16 CPUs will get us 32GiB memory. I've updated this line accordingly.

P.S. This compute environment specifies both `"allocationStrategy": "BEST_FIT",` and `"desiredvCpus": 192,` so I hope deploying a job as we're doing here will still get a 16 CPU model? cc @tsibley 

P.P.S. Given 16 CPUs + 32GiB memory, how does the scheduler choose between c5.4xlarge and c5d.4xlarge?

.github/workflows/rebuild-100k.yml
jameshadfield: Done :)

.github/workflows/rebuild-100k.yml
jameshadfield: Done :)

nextstrain_profiles/100k/README.md
jameshadfield: Going to leave as long form - hopefully we won't need to actually run this by hand very frequently and it's clearer (to me, at least)

nextstrain_profiles/100k/config.yaml
jameshadfield: Thanks for the clarifications! Added a comment to the build YAML explaining this.

P.S. This isn't just for our builds, it's going to happen everywhere. Even a simple single-sample scheme grouped by `year month` will oversample the current month unless we are running things on the final day of the month.

nextstrain_profiles/100k/config.yaml
jameshadfield: Done :)

nextstrain_profiles/100k/README.md
jameshadfield: Done :)

.github/workflows/rebuild-100k.yml
jameshadfield: Ha! Good catch. Thanks

workflow/snakemake_rules/common.smk
jameshadfield: Thanks! Incorporated in [a41a9f6](https://github.com/nextstrain/ncov/pull/1032/commits/a41a9f63e4f1ad7c05e1d176ecfc37d38817e5c5)

I changed `config["upload"]` to `config["upload_targets"]` because a bunch of our workflows already defined that key, although I don't think the workflow actually uses it 🤷 . For instance:

https://github.com/nextstrain/ncov/blob/0a093c0677a1ac9c64564572bd43cb7bf8602eda/nextstrain_profiles/nextstrain-gisaid/builds.yaml#L13-L14

nextstrain_profiles/100k/README.md
jameshadfield: Done :)

jameshadfield:
Finally got back to this. Have incorporated all comments & am currently [running on AWS](https://us-east-1.console.aws.amazon.com/batch/home?region=us-east-1#jobs/detail/d502086c-a1f1-4686-bb01-c5b285d4d75f).

I'll probably add a 21L version to this `builds.yml`, but will do so after the merge of #1029.

.github/workflows/rebuild-100k.yml
jameshadfield: Running a test build of this PR (via `nextstrain build` as a new GitHub action isn't runnable from a PR) with `--cpus 16 --memory 31GiB` was allocated a machine with 16 vCPUs. It's not clear which actual model it uses.

.github/workflows/rebuild-100k.yml
jameshadfield: Done :)

.github/workflows/rebuild-100k.yml
tsibley: > P.S. This compute environment specifies both `"allocationStrategy": "BEST_FIT",` and `"desiredvCpus": 192,` so I hope deploying a job as we're doing here will still get a 16 CPU model? cc @tsibley

Yes. `desiredvCpus` is a gauge value that floats up and down with demand (between `minvCpus`, which we should always have at 0, and `maxvCpus`). Batch modifies it as necessary for jobs, and compute resources in other services (e.g. ECS in our case) scale up and down in response. You typically don't modify it directly.

> P.P.S. Given 16 CPUs + 32GiB memory, how does the scheduler choose between c5.4xlarge and c5d.4xlarge?

c5d is a separate instance family than c5, so those instances are not made available in our default `nextstrain-job-queue`.

workflow/snakemake_rules/common.smk
tsibley: Hmm. If we don't use `upload`, then I'd be inclined to take it over for this use. `upload_targets` is not as clear (we don't use "target" very much at all in this workflow) or as consistent with other rule-based config (e.g. `filter`).

jameshadfield: Sure. Done in [6b7c9e9](https://github.com/nextstrain/ncov/pull/1032/commits/6b7c9e9424e620a3ca1ae3a3085b91e8e6f7c331).

tsibley: Ah, I also just did https://github.com/nextstrain/ncov/commit/0a04128b5d7274ab22f8be64a6f2e859fadbdf46.

jameshadfield: All good. I'll rebase & merge tomorrow.

PR #1037, opened 2022-12-14
Update scripts to use augur.io Developer API

joverlee521:
Starting with Augur 19.0.0, the only public augur.io functions are `read_metadata` and `read_sequences`. All other functions must be imported from the Developer API.


joverlee521:
Closing in favor of https://github.com/nextstrain/augur/pull/1114 and waiting on release of Augur 19.1.0.

PR #1047, opened 2023-02-13
docs: fix links to ZSTD-compressed full datasets

chaoran-chen:
The links to the ZSTD-compressed files such as https://data.nextstrain.org/files/ncov/open/global/metadata.tsv.zst don't work (error 404).


corneliusroemer (approved):
Thanks for the fix!

Issue #1054, opened 2023-03-30
Why use unaligned sequences as "aligned" input instead of the actually aligned sequences?


tsibley:
Why do we do this (also in our other profile configs):

https://github.com/nextstrain/ncov/blob/cfa73beacae90007648dbe1362af9e896cb3713e/nextstrain_profiles/nextstrain-open/builds.yaml#L28-L34

instead of something more self-explanatory like this?

```diff
-# Note: unaligned sequences are provided as "aligned" sequences to avoid an initial full-DB alignment
-# as we re-align everything after subsampling.
 inputs:
   - name: open
     metadata: "s3://nextstrain-data/files/ncov/open/metadata.tsv.gz"
-    aligned: "s3://nextstrain-data/files/ncov/open/sequences.fasta.xz"
+    aligned: "s3://nextstrain-data/files/ncov/open/aligned.fasta.xz"
     skip_sanitize_metadata: true
```

jameshadfield:
It's a hack 

https://github.com/nextstrain/ncov/blob/cfa73beacae90007648dbe1362af9e896cb3713e/nextstrain_profiles/nextstrain-gisaid/builds.yaml#L28-L29

jameshadfield:
My memory is that this trick preceded the existence of `aligned.fasta.xz`. I don't see any reason why not to use that input, but @corneliusroemer may know specifics.

emmahodcroft:
I looked at the git blame and looks like at least the comment was changed in Dec 2021. Here's the meeting notes from just prior to that commit - seems this was part of moving alignment (cached) to ingest and reorganizing `ncov` so that many things happened after subsampling (?): https://docs.google.com/document/d/1rV5S8_zA92MFj7l-6LpxkGjQCi9JUtPPAn6ccjlqvo8/edit#heading=h.7nuap5p45ey7

In [this PR](https://github.com/nextstrain/ncov/pull/814) (linked in above meeting notes) @huddlej writes:
> modifies the Nextstrain profiles to use `aligned` inputs for sequences even though they are not aligned (to skip right to subsampling)

Maybe he remembers a bit more about why this approach?

corneliusroemer:
I don't think I was involved at all with this ever :) But I had a look nonetheless.

The workflow behaves weirdly. When I run the dev-docs recommended command: `snakemake -pf all --profile nextstrain_profiles/nextstrain-open -n` the workflow assumes that the data is already there - but it definitely isn't, I don't have any data. What's going? 

[...]

PR #823, opened 2021-12-16
WIP: remove preprocessing pipelines

jameshadfield:
WIP

- [x] Remove preprocessing config files
- [x] Change main profiles to start from sequences but call them aligned
- [x] Remove conditional statements using `config.upload`
	- [x] config.schema.yaml
	- [x] slack messages
	- [x] upload rule doesn’t upload alignments / filtered (previous PR)
	- [x] upload files
- [x] remove `trigger_phylogenetic_rebuild`
- [ ] GitHub actions
	- [x] Remove GitHub actions which run preprocess
	- [x] Sister PR in ncov-ingest to change trigger to `gisaid/rebuild` / `open/rebuild`. This is https://github.com/nextstrain/ncov-ingest/pull/255
- [ ] Docs pass
	- [x] dev-docs
- [ ] Rebase upon parent PR #814 after changes made there
- [ ] CI

PR #1057, opened 2023-04-06
Start from actually-aligned sequences

tsibley:
Use aligned sequences as the aligned sequences input, rather than pass off unaligned sequences as the aligned sequences input.

This should be inconsequential to workflow behaviour or results, but it makes the config a bit more straightforward and less confusing.

In a quick dig thru history, it seems like ncov-ingest's aligned.fasta.xz was not _quite_ available when we first switched our profiles to use an "aligned" input instead of a "sequences" input.  The original use of "aligned" with unaligned sequences was driven by run time concerns and related to the move of the filtering step after the subsampling step and move of the "preprocess" steps from this workflow (ncov) to ncov-ingest.¹

Resolves <https://github.com/nextstrain/ncov/issues/1054>.

¹ <https://github.com/nextstrain/ncov/pull/814>
  <https://github.com/nextstrain/ncov/pull/823>

## Testing

I've been starting from `aligned.fasta.zst` when testing the [21L-rooted builds](https://github.com/nextstrain/ncov/pull/1029) and all seems fine.

## Release checklist

### DIFF
--- a/nextstrain_profiles/nextstrain-country/builds.yaml
+++ b/nextstrain_profiles/nextstrain-country/builds.yaml
@@ -25,12 +25,10 @@ include_hcov19_prefix: True
 files:
   description: "nextstrain_profiles/nextstrain-country/nextstrain_description.md"
 
-# Note: unaligned sequences are provided as "aligned" sequences to avoid an initial full-DB alignment
-# as we re-align everything after subsampling.
 inputs:
   - name: gisaid
     metadata: "s3://nextstrain-ncov-private/metadata.tsv.gz"
-    aligned: "s3://nextstrain-ncov-private/sequences.fasta.xz"
+    aligned: "s3://nextstrain-ncov-private/aligned.fasta.xz"
     skip_sanitize_metadata: true
 
 # Define locations for which builds should be created.
--- a/nextstrain_profiles/nextstrain-gisaid/builds.yaml
+++ b/nextstrain_profiles/nextstrain-gisaid/builds.yaml
@@ -25,12 +25,10 @@ include_hcov19_prefix: True
 files:
   description: "nextstrain_profiles/nextstrain-gisaid/nextstrain_description.md"
 
-# Note: unaligned sequences are provided as "aligned" sequences to avoid an initial full-DB alignment
-# as we re-align everything after subsampling.
 inputs:
   - name: gisaid
     metadata: "s3://nextstrain-ncov-private/metadata.tsv.gz"
-    aligned: "s3://nextstrain-ncov-private/sequences.fasta.xz"
+    aligned: "s3://nextstrain-ncov-private/aligned.fasta.xz"
     skip_sanitize_metadata: true
 
 # Define locations for which builds should be created.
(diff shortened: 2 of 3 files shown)


rneher:
If I remember correctly, the history of this is a bit convoluted. We used to require all sequences to be aligned since we generated subsamples by comparing all sequences to a focal sample. The workflow as it stands could run using aligned or unaligned sequences, I think. But we have an alignment step in the workflow

https://github.com/nextstrain/ncov/blob/trs/actually-aligned/workflow/snakemake_rules/main_workflow.smk#L60-L104

One might consider it a bit confusing that the aligned sequences are aligned again (and translated). There is also one caveat: The aligned sequences were aligned pairwise to the reference and insertions stripped. So some information got lost and translations in particular might be messed up if there are compensated frameshifts in some of the ORFs (I don't know of any examples for SC2). Using the original sequences avoids this and provides the `insertions.csv` file that could be used in downstream analysis.

tsibley:
I agree it's a bit convoluted. I wasn't involved in the history of it much, but I presume it got to be this way for decent reasons (e.g. to support different use cases). There's probably still room for improvement overall in the workflow graph, but that wasn't the intent of this PR.

The goal here was to provide what we claim to provide in the config. I don't particularly care if we provide unaligned sequences as the `sequences` input key vs. aligned sequences as the `aligned` input key, but more that what we provide should match what we say it is. :-)

Providing an `aligned` input key does skip that `align` rule, so they're not aligned again there. However, no matter what's provided as initial inputs, all sequences are always aligned a second time by `nextclade` in the `build_align` rule, after subsampling and filtering for each build:

https://github.com/nextstrain/ncov/blob/0a093c0677a1ac9c64564572bd43cb7bf8602eda/workflow/snakemake_rules/main_workflow.smk#L469-L509

rneher:
yes, it skips the `align` rule bit still does the `build_align` rule which takes input from `combine_samples`:

https://github.com/nextstrain/ncov/blob/0a093c0677a1ac9c64564572bd43cb7bf8602eda/workflow/snakemake_rules/main_workflow.smk#L425

this is very confusing. In the current workflow, if one pretends the unaligned input sequences are aligned, one aligns just once.  otherwise, things are aligned twice.

jameshadfield:
From memory: the `build_align` was introduced to improve alignments once QC & filtering had been done, as some erroneous sequences may have detrimentally affected the initial alignment (this may have been way back when we were using `mafft`). The reason we kept the initial `align` step around was to facilitate proximity filtering, which required an alignment; we no longer use this functionality and it's not computationally feasible for the entire dataset anymore.

Assuming proximity filtering is the only reason to align everything, perhaps modifying the workflow to skip `align` _unless_ we have proximity filtering is the way to go. Or we could drop proximity support altogether.

tsibley:
@rneher Yes, that's what I said.

@jameshadfield Thanks for providing that background. Seems to line up with my presumption above that "it got to be this way for decent reasons."

<p align="center">⁂</p>

The tiny change in this PR is meant to be only a tiny clarification with zero impact on functionality. I'm going to merge it. The discussion of what steps the workflow should or shouldn't be doing (and how to do it) can continue elsewhere.

PR #1065, opened 2023-04-22
Feat/clade display names

rneher:
this PR adds

 - a file defining mapping between bare year-letter nextstrain clades and display-names
 - a script that re-writes the clades file using the new display names
 - removes the previously duplicated clades file with the alternative names.


rneher:
Trial runs should appear here:

https://nextstrain.org/staging/ncov/open/trial/clade-display-names/global/all-time

rneher:
this worked as expected. other than for the problem that the colors aren't generated properly since the display names are not in the metadata.

rneher:
image

rneher:
this now works happily:
image

defaults/clade_display_names.yml
```diff
@@ -0,0 +1,32 @@
+19B: 19B
```
corneliusroemer, line 1: Is there a reason there's no `19A: 19A`?

corneliusroemer:
Once this is merged I'll have to adjust the Nextclade data workflows, so I'll see what you come up with here. If we wanted to maintain WHO names for colouring, would it make sense to define a `defaults/clade_who_mapping.yml`? Or shall I maintain this separately in the workflow? That would make it a little less convenient for updates as currently all the changes for new clades are in this `ncov` repo.

I remember we talked about moving the ground truth to a separate "clade definition" repo - but we can also just have it in here, there's no real circularity.

defaults/parameters.yaml
```diff
@@ -65,7 +65,8 @@ files:
   auspice_config: "defaults/auspice_config.json"
   lat_longs: "defaults/lat_longs.tsv"
   description: "defaults/description.md"
-  clades: "defaults/clades.tsv"
+  clades: "defaults/clades_nextstrain.tsv"
```
huddlej, line 68: I wonder if we should keep the default file named as `clades.tsv`, since this is what we refer to in the docs (e.g., [labeling clades](https://docs.nextstrain.org/projects/ncov/en/latest/guides/workflow-config-file.html#labeling-clades), [the workflow reference guide](https://docs.nextstrain.org/projects/ncov/en/latest/reference/workflow-config-file.html#clades), etc.) and call the derived file for the workflow something different?

If we do rename the default, we just need to update references to it in the docs.

corneliusroemer: Not just, as I use the exact file from another repo ;)

defaults/clade_display_names.yml
rneher: none that I can think of ;)

defaults/parameters.yaml
rneher: I had named it in analogy to `clades_who`, but the latter isn't really part of this workflow. I am fine with renaming it.

rneher:
@huddlej, did this address your concern?

huddlej:
@rneher Yeah, this looks great. Thank you!

rneher:
ran another trial run. looks good. merging now. 

https://nextstrain.org/staging/ncov/open/trial/clade-display-names-colors/asia/all-time?label=clade:22F%20%28XBB%29&m=div

PR #1062, opened 2023-04-18
sanitize_metadata: Filter duplicates after strain column transformations

victorlin:
## Description of proposed changes

Previously, duplicates were filtered before strain column transformations. These transformations could have resulted in a final output file containing duplicates.

Filter duplicates after strain column transformations but before other transformations to avoid unnecessarily transforming rows that would be dropped by duplicate filtering.

This also requires strain column transformations when mapping strain names to database IDs.

## Related issue(s)

Fixes #1049

## Testing

- [x] Added and modified a test to show new behavior.

## Release checklist

If this pull request introduces backward incompatible changes, complete the following steps for a new release of the workflow:

 - [ ] Determine the version number for the new release by incrementing [the most recent release](https://github.com/nextstrain/ncov/releases) (e.g., "v2" from "v1").
 - [ ] Update `docs/src/reference/change_log.md` in this pull request to document these changes and the new version number.
 - [ ] After merging, [create a new GitHub release](https://github.com/nextstrain/ncov/releases/new) with the new version number as the tag and release title.

If this pull request introduces new features, complete the following steps:

 - [ ] Update `docs/src/reference/change_log.md` in this pull request to document these changes by the date they were added.

PR #1066, opened 2023-04-27
Create locations DataFrame with a strain index

victorlin:

Prior to that commit, DataFrames were not indexed by strain and used the
default counter index. The concat relied on both the locations and
metadata DataFrames having the same counter index. This was broken in
the linked commit when the metadata was switched to index on strain.

Solutions are to either (1) revert to using counter index for both, (2)
use strain index for both, or (3) ignore the index when concatenating.

A strain index has benefits and is used elsewhere in the script, so this
commit implements solution (2).

## Related issue(s)

Follow-up to #1062

## Testing

- [x] Added and modified a test to show new behavior.

## Release checklist

### DIFF
--- a/scripts/sanitize_metadata.py
+++ b/scripts/sanitize_metadata.py
@@ -479,12 +479,13 @@ if __name__ == '__main__':
             # Parse GISAID location field into separate fields for geographic
             # scales. Replace missing field values with "?".
             if args.parse_location_field and args.parse_location_field in metadata.columns:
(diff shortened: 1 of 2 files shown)


victorlin:
Merging because this works for myself and [another user](https://discussion.nextstrain.org/t/ncov-errors-from-combine-metadata-py-due-to-unexpected-behavior-in-sanitize-metadata-py/1361/28).

PR #1069, opened 2023-05-13
use stochastic polytomy resolution

rneher:
Stochastic polytomy resolution is faster and leads to more typical trees. The previous (still default in treetime) mode resulted in caterpillar like trees.


rneher:
Not that the docker container as augur 22.0.1, I could finally test this. The trail open run is here:

https://nextstrain.org/staging/ncov/open/trial/stochastic-resolve/global/all-time

(and the other resolutions). Generally, I think this is a big improvement, in particular for recent builds with large polytomies. The `2m` resolution for example currently looks like this

image

and with stochastic resolution would look like this

image

(the difference is even starker for `1m`).

trvrb:
Comparing live tree and the trial run I think this is a huge improvement and fits with coalescent expectation for tree structure. Thanks @rneher!

PR #1070, opened 2023-05-16
Introduce minimum augur version check

jameshadfield:
I implemented this as part of updating #1000, but there are so many ways this pipeline may be run (and which I don't have experience with) that it's not appropriate for that PR. The downside of false-positives outweighs the benefits, so any version check has to be sure we are using the augur version we think we are - probably more so than in the code's current form. It would be worth it to catch these incompatibilities early as I'm sure this'll trip up some people. 

Note that this PR failed CI in a useful way :)
```
Your version of augur (21.1.0) needs to be upgraded to use this workflow.
The minimum augur version required is 22.0.0.
```

### DIFF
--- a/Snakefile
+++ b/Snakefile
@@ -17,6 +17,30 @@ import time
 # use of Bash's "strict mode", as we rely on that behaviour.
 shell.prefix("set -euo pipefail; export AUGUR_RECURSION_LIMIT=10000; ")
 
+# Check the installed version of augur matches the minimum for this workflow
+minumim_supported_augur_version = "22.0.0"
+try:
+    if not snakemake.workflow.workflow.use_conda:
+        from packaging import version # an explicit dependency of augur
+        from augur import version as augur_version
+        if version.parse(augur_version.__version__) < version.parse(minumim_supported_augur_version):
+            logger.error(textwrap.dedent(
+                f"\nYour version of augur ({augur_version.__version__}) needs to be upgraded to use this workflow." +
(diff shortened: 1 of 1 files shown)


Snakefile
```diff
@@ -17,6 +17,30 @@ import time
...
+minumim_supported_augur_version = "22.0.0"
+try:
+    if not snakemake.workflow.workflow.use_conda:
+        from packaging import version # an explicit dependency of augur
+        from augur import version as augur_version
+        if version.parse(augur_version.__version__) < version.parse(minumim_supported_augur_version):
+            logger.error(textwrap.dedent(
+                f"\nYour version of augur ({augur_version.__version__}) needs to be upgraded to use this workflow." +
+                f"\nThe minimum augur version required is {minumim_supported_augur_version}." +
+                "\nPlease see https://docs.nextstrain.org/en/latest/guides/manage-installation.html for help with updating your installation." +
```
jameshadfield, line 30: We could add "If you think this version check is not applicable for how you are running this workflow, you may remove it by setting `config['skip_augur_version_check'] = true`" here if we are worried (like I am) about not wanting to completely disable some rare (?) invocations of the workflow.

huddlej:
There is [an ancient (2-year-old now) PR](https://github.com/nextstrain/ncov/pull/611) that does something similar but only when the workflow errors. Checking preemptively seems maybe better, though.

victorlin:
Noting that this is analagous to https://github.com/nextstrain/mpox/commit/d9f6d474391c1e657a5deec1bfd0ebb90b1ac4d2.

PR #1084, opened 2023-09-20
Fix snakemake rule existence checks

jameshadfield:
We use the existence of certain rules to control the DAG and the way rules are stored internally changed in snakemake 7.32.1 <https://github.com/snakemake/snakemake/commit/65c79a48f956077839bb5ab1ea8d60a5f0ddecab>

This functionality is (probably?) only used for internal nextstrain builds where we use the `auspice_config` rule to programmatically generate the config.

On 2023-09-14 we [updated snakemake from 7.24.1 to 7.32.3 in our docker image](https://github.com/nextstrain/docker-base/pull/176). This caused our nCoV builds to (incorrectly) use the default auspice-config JSON rather than call the `auspice_config` rule to generate it. This can be seen by comparing the datasets from [Sept 12th](https://next.nextstrain.org/ncov/gisaid/global/6m/2023-09-12) and [Sept 15th](https://next.nextstrain.org/ncov/gisaid/global/6m/2023-09-15)

### DIFF
--- a/workflow/snakemake_rules/main_workflow.smk
+++ b/workflow/snakemake_rules/main_workflow.smk
@@ -1394,6 +1394,14 @@ rule build_description:
         with open(output.description, "w", encoding = "utf-8") as o:
             o.write(template.safe_substitute(context))
 
+def rule_exists(name):
+    # Some versions of Snakemake throw a WorkflowError even with the
+    # three-arg getattr(), so catch any error and assume non-existence.
+    try:
+        return bool(getattr(rules, name, None))
+    except:
+        return False
+
 def get_auspice_config(w):
     """
     Auspice configs are chosen via this heirarchy:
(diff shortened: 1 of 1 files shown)


trvrb (approved):

workflow/snakemake_rules/main_workflow.smk
```diff
@@ -1394,6 +1394,15 @@ rule build_description:
...
             o.write(template.safe_substitute(context))
 
+def rule_exists(name):
+    # Snakemake 7.32.1 changed the way rules are stored internally
+    # https://github.com/snakemake/snakemake/commit/65c79a48f956077839bb5ab1ea8d60a5f0ddecab
+    from packaging import version # an explicit dependency of augur
+    if version.parse(snakemake.__version__) < version.parse('7.32.1'):
+        return name in rules.__dict__
+    else:
+        return name in rules.__dict__['_rules']
```
tsibley, line 1404: I'd recommend avoiding the private internals and instead defining this in terms of accessibility via the public API:

```python
def rule_exists(name):
    try:
        return bool(getattr(rules, name, None))
    except:
        return False
```

This works on both versions.

tsibley (changes requested):

workflow/snakemake_rules/main_workflow.smk
jameshadfield: Sure. Done.

PR #1099, opened 2024-02-16
dev: Remove python-version input to setup-nextstrain-cli action

tsibley:
Deprecated and no longer used as setup-python isn't called.

Related-to: <https://github.com/nextstrain/.github/pull/55>


tsibley:
Merging as I expect this to have no impact, but note that CI doesn't test the actual changes here since they're limited to production workflows.

victorlin (commented):
👍

PR #1104, opened 2024-05-01
Group by week for 1m/2m Asia builds

victorlin:
To be consistent with other 1m/2m builds (6a94e46896909fd292ba865355d5759875ba700c).

Noticed while working on #1102.


trvrb (approved):
I haven't tested this, but I agree we should be consistent here and this makes sense to me.

victorlin:
I haven't tested either, but will look out for the next automated build. I'm not sure it'll be easy to tell the difference, since the recent timeframe is dominated by global context due to the relative lack of recent sequences from Asia.

PR #1106, opened 2024-05-03
Use weighted sampling for Asia builds

victorlin:
_tracked by https://github.com/nextstrain/ncov/issues/1141_

This replaces the Asia/China/India split with population-based weighted sampling (possible with https://github.com/nextstrain/augur/pull/1454).

## Previews

- open/asia [1m](https://next.nextstrain.org/staging/ncov/open/trial/victorlin-weighted-sampling-11/asia/1m), [2m](https://next.nextstrain.org/staging/ncov/open/trial/victorlin-weighted-sampling-11/asia/2m), [6m](https://next.nextstrain.org/staging/ncov/open/trial/victorlin-weighted-sampling-11/asia/6m), [all-time](https://next.nextstrain.org/staging/ncov/open/trial/victorlin-weighted-sampling-11/asia/all-time)
    - NOTE: new builds blocked on https://github.com/nextstrain/ncov-ingest/issues/469. These links are for the last successful build.
- gisaid/asia [1m](https://next.nextstrain.org/staging/ncov/gisaid/trial/victorlin-weighted-sampling-11/asia/1m), [2m](https://next.nextstrain.org/staging/ncov/gisaid/trial/victorlin-weighted-sampling-11/asia/2m), [6m](https://next.nextstrain.org/staging/ncov/gisaid/trial/victorlin-weighted-sampling-11/asia/6m), [all-time](https://next.nextstrain.org/staging/ncov/gisaid/trial/victorlin-weighted-sampling-11/asia/all-time)

## Analysis

I think a good comparison is gisaid/asia/all-time [before](https://next.nextstrain.org/ncov/gisaid/asia/all-time@2024-08-21?f_region=Asia) and [after](https://next.nextstrain.org/staging/ncov/gisaid/trial/victorlin-weighted-sampling-11/asia/all-time?f_region=Asia) weighted sampling. Some notes from that comparison:

1. With weighted sampling, more sequences are included for large countries such as Indonesia which did not get the same treatment as China and India prior to this PR.
2. With weighted sampling, there is a high chance that no sequences are included for small countries such as Maldives.
3. The sampling difference results in some subtle but maybe significant differences in the frequencies panel.

## Questions for reviewers

1. What can be done to check that the sampling results are appropriate?
2. ✅ Is it a good idea to weight on population size? I was thinking per-country case counts over time might be better, but not sure where to get that data. ([population size yes, case counts no](https://github.com/nextstrain/ncov/pull/1106#issuecomment-2151177258))

## Notable comment threads

- [x] [Weight using case count data?](https://github.com/nextstrain/ncov/pull/1106#discussion_r1596112762) No
- [x] [Where to define the weights file?](https://github.com/nextstrain/ncov/pull/1106#discussion_r1661658187) `config.subsampling.<sampling_scheme>.<sample>.group_by_weights`
- [x] [Use UN data source?](https://github.com/nextstrain/ncov/pull/1106#discussion_r1700877314) Yes

## Checklist

- [x] Sampling results look ok? [Yes](https://github.com/nextstrain/ncov/pull/1106#pullrequestreview-2236968664)
- [x] Drop 4e5ee678e3e4f6a8256955c8c1ab99d2fccecd60
- [x] Merge+release https://github.com/nextstrain/augur/pull/1454
- [x] Add Augur version info on 7a3cb766df50e76e82deb1416e34b41e4d412b4d
- [ ] Post-merge: update minimum Augur version on https://github.com/nextstrain/ncov/issues/1118

## Release checklist


victorlin:
Not ready to be merged but opening for initial review and feedback to shape implementation in https://github.com/nextstrain/augur/pull/1454.

nextstrain_profiles/nextstrain-open/builds.yaml
```diff
@@ -387,30 +246,18 @@ subsampling:
 
   # Custom subsampling logic for region Asia over 6m
   # Grouping by division
-  # Separating three buckets for China, India and elsewhere
+  # Grouping by country weighted by population size
   # 4375 total
```
genehack, line 250: is this comment still correct?

genehack (approved):

nextstrain_profiles/nextstrain-open/builds.yaml
victorlin: Yes, it's still 4375 total (700+175+2800+700). Note that this is the **target** size and doesn't reflect number of samples that end up in the Auspice dataset. That can be less due to the caveat in the PR description and other steps downstream of subsampling such as sequence length filter, clock filter, etc.

genehack: thanks -- i was missing the 175 because of a code fold in the PR >_<

joverlee521:
> I was thinking per-country case counts over time might be better, but not sure where to get that data.

We ingest country case counts from [OWID](https://covid.ourworldindata.org/data/owid-covid-data.csv) in forecasts-ncov and upload them to `s3://nextstrain-data/files/workflows/forecasts-ncov/cases/global.tsv.gz`.

workflow/snakemake_rules/main_workflow.smk
```diff
@@ -265,6 +267,14 @@ rule index_sequences:
             --output {output.sequence_index} 2>&1 | tee {log}
         """
 
+rule get_weights:
+    output: "data/country_week_weights.tsv"
```
victorlin, line 271: Re: [using country case counts from OWID](https://github.com/nextstrain/ncov/pull/1106#issuecomment-2101636473)

Trying this with 5ca1afa129268f743e0a76c3337d00364116353e. Trial runs: [open](https://github.com/nextstrain/ncov/actions/runs/9025041404/job/24800039148), [gisaid](https://github.com/nextstrain/ncov/actions/runs/9025045965/job/24800051957)

There's some missing case count data, i.e. sequences are present despite country/week combinations having a case count of 0. Reasons and how these were addressed:

1. Things listed out here:
 
    https://github.com/nextstrain/ncov/blob/5ca1afa129268f743e0a76c3337d00364116353e/scripts/get_weights.py#L33-L44

2. On top of (1), there are also sequences present for weeks and countries missing entirely from case count data. To address this, I added a [patch](https://github.com/nextstrain/augur/commit/60a0f3ed2207c5746aa6fc1aa29ab3f75990cb9f) in the Augur PR that simply drops these sequences rather than erroring out.

Quick tests on the 100k subsampled dataset make me skeptical about this approach. The high case counts attempt to capture more than the available amount of sequences, while at the same time making weights for weeks with small case counts very small and thus not sample from those at all (despite their potential importance in phylogenetic analysis). Hopefully the trial run results will make it clear whether to pursue this direction.

victorlin: Trial runs are complete. Results seem significantly different – it'd be good to have a phylo person take a look. Here's quick links to `gisaid/asia/all-time` for:

1. Current build on `master`: [uniform sampling by division+month](https://next.nextstrain.org/ncov/gisaid/asia/all-time)
2. Original PR implementation: [weighted sampling by country population, uniform sampling by month](https://next.nextstrain.org/staging/ncov/gisaid/trial/victorlin-weighted-sampling-2/asia/all-time)
3. New PR implementation: [weighted sampling by country+week case counts](https://next.nextstrain.org/staging/ncov/gisaid/trial/victorlin-weighted-sampling-3/asia/all-time)

Note from `builds.yaml` that (2) and (3) only apply to the focal Asia samples. Context (non-Asia) samples still conform to (1), but that could be changed.

huddlej: @victorlin First, one big picture comment related to [this comment in the augur PR](https://github.com/nextstrain/augur/pull/1454#issuecomment-2121165641):

> I want to make sure we have at least one solid use case for this feature.

I see example 2 above ("Original PR implementation") as exactly the use case for this feature where we want to minimize [the kind of manual counting and hardcoded subsampling blocks](https://github.com/nextstrain/ncov/commit/9fd01c64f9471638757ab6ce2dd6a43fe090bfeb) that @trvrb had to implement to support proper weighting of India and China relative to the rest of Asia. If we can replace that subsampling YAML approach with a clearer, simpler table of weights by country or region, that's a big win.

Digging into the specific trees you linked to above, there are two main points that came to mind:

1. Weights by country population, uniform sampling by month looks remarkably like the original implementation, but with a simpler config in the ncov repo and slightly better representation of countries like India and China than even the original implementation. Below I've included example views of the map colored by clade per country for each of the three builds linked above where I've also filtered to "region=Asia".

The original view (sampling by division and month):
image

The weighted by population view (sampling by weighted country and uniform month):
image 2

The weighted by case counts view (sampling by weighted country and case counts):
image 3

The first two views show how similar the two builds are qualitatively, but the second view has more data for China, India, Indonesia, and Pakistan than the first. The second view has correspondingly fewer samples for Japan, Thailand, and Malaysia. This seems correct. For example, Indonesia and Malaysia have similar numbers of samples in the first view, but Indonesia has 8 times as many people as Malaysia. The second view more accurately represents that difference in population with 194 samples for Indonesia and 23 for Malaysia.

From 31e1099's commit message:

> This requires changing the geographical grouping resolution from division to country, but I assume it was only grouped by division in an attempt to have varying group sizes per country, and that population-based weighting is an acceptable replacement.

I am curious whether @trvrb chose to group by division for this reason only (to get varying group sizes per country) or if he really preferred even sampling by division. Anecdotally (and perhaps I've misremembered the specifics here), I thought he and Marlin were concerned about overrepresentation of sequences from specific divisions in India as new variants were emerging. To explore this case, I made example map views from each of your builds above at the division resolution and filtering to "country=India".

Original view of India by division:
image 4

India by division with population weights by country:
image 5

India by division with weights by country and case counts:
image 6

The third view confirms what we already saw above, that sampling by case counts enriches for older clades, which I discuss more in my second point below.

The first view shows a similar representation of each state in India, but the second view (which only groups by country) shows an overrepresentation of samples from a smaller set of states. I can't say that either one of these views is "the correct" implementation, since that judgement depends on the question we're trying to answer with these trees. However, this difference in results does raise the question for me of whether would could have weighted subsampling by country _and_ uniform sampling by division in the same group by logic. In other words, could we have `group_by: "country division year month"` and still use the weights per country? If so, we get the best of both worlds for this use case.

2. Weights by case counts do not seem as useful for this specific ncov use case. The tree from example 3 above overrepresents samples from older clades circulating around the start of 2023 (22B) at the expense of including more samples from recent clades (e.g., 22F, 23I). Your first comment in this thread highlights all of the issue with case counts which I think could be summarized as: case counts are noisy data. They are also less reliable data now that there is less funding for surveillance. While population counts per country are also inaccurate and will change over time, the time scale of those changes is decades and you still get useful relative comparisons between countries independent of their genomic surveillance budgets.

This comment above specifically made me worry about using case counts for subsampling in the ncov builds:

> there are also sequences present for weeks and countries missing entirely from case count data. To address this, I added a [patch](https://github.com/nextstrain/augur/commit/60a0f3ed2207c5746aa6fc1aa29ab3f75990cb9f) in the Augur PR that simply drops these sequences rather than erroring out.

When case count data are missing but sequences are present, I would still want to have the sequences in the tree. Even if you reverted this behavior of Augur to not drop sequences when cases were missing, I probably wouldn't reach for case counts as weights in these builds. That said, Miguel might reach for exactly this when designing a subsampling scheme for SARS-CoV-2 in King County where the case counts during a specific time span are more reliable and the analysis is more sensitive to geographic sampling and case load during that time. I would show this example approach to weighting by case counts to him and get his feedback on how useful it would have been for him.

victorlin: Thanks @huddlej, this is really valuable. You confirmed my suspicious - I'll revert 6b6cd2b5cfb96b5a4eaa3a7fdfc9585a6a367e8c.

> could we have `group_by: "country division year month"` and still use the weights per country?

Trying this with 835481da34d7bcda0eda690607730d72eb8763a8 ([GISAID trial run](https://github.com/nextstrain/ncov/actions/runs/9166507128/job/25202025886)). Will update here when the run is complete.

huddlej: Nice! That looks pretty good to me at first glance. The increased number of sequences for India as a whole makes sense from the perspective that each division has to be equally represented by a fixed number of sequences per group within the country and there are a lot of divisions. Looking at the Malaysia/Indonesia representation, I see the same effect where Indonesia has more divisions than Malaysia and ends up with 12 times as many sequences instead of 8 times as many that we'd expect from country population weights alone. This seems like an expected behavior of the group by country and division, though, so I might not worry about this much...

trvrb:
I see
>Is it a good idea to weight on population size? I was thinking per-country case counts over time might be better, but not sure where to get that data.

Case counts no longer mean anything and haven't really meant anything since early 2022. We can see this today in the parsed case counts file `country_week_weights.tsv` where Afghanistan with a population of 41M reported 638 cases in the week of (2024, 20), while Brunei with a population of 500k reported 175 cases in this same time period. This is 1.5 cases per 100k in Afganistran and 35 cases per 100k in Brunei. This 23-fold difference is going to be almost entirely reporting rate differences.

If you do subsampling based on reported cases it will strongly bias included samples to (wealthier) countries with better surveillance.

For ongoing ncov builds, we should be weighting on population size. I wouldn't worry about admin division population size and would just worry about country-level population sizes in terms of weighting.

I'd start by rolling back 6b6cd2b5cfb96b5a4eaa3a7fdfc9585a6a367e8c.

trvrb:
Following from this, I worked from 1afb6d7e5278bad4a69801411d791d5b2cd7d5e8 and tried:
```
python3 scripts/get_population_sizes.py --output data/country_population_weights.tsv
``` 
from a freshly updated `nextstrain shell` command and got the following error
```
  File "/nextstrain/build/scripts/get_population_sizes.py", line 67, in <module>
    export_imf(args.output)
  File "/nextstrain/build/scripts/get_population_sizes.py", line 13, in export_imf
    workbook = xlrd.open_workbook(file_contents=excel_contents, ignore_workbook_corruption=True)
TypeError: open_workbook() got an unexpected keyword argument 'ignore_workbook_corruption'
```
and if I drop `ignore_workbook_corruption=True` in line 13 of the script I get an error about a corrupted workbook:
```
xlrd.compdoc.CompDocError: Workbook corruption: seen[2] == 4
```

https://www.cia.gov/the-world-factbook/field/population/country-comparison/#:~:text=DOWNLOAD-,DATA,-Rank looks potentially cleaner.

victorlin:
I've dropped 6b6cd2b5cfb96b5a4eaa3a7fdfc9585a6a367e8c and updated f038890d04863b92d0da2db44fc798b512510e8f to use "the world factbook" as source of population data. Also moved the script and data files into version control (great suggestion).

I'm going to run this through the trial builds to check if the updated data source contains the right country names.

workflow/snakemake_rules/main_workflow.smk
victorlin: Dropped 835481da34d7bcda0eda690607730d72eb8763a8 per https://github.com/nextstrain/ncov/pull/1106#issuecomment-2151177258:

> For ongoing ncov builds, we should be weighting on population size. I wouldn't worry about admin division population size and would just worry about country-level population sizes in terms of weighting.

victorlin:
Cleaned up the commits and updated trial build links in the PR description. Everything is still draft with a few lingering FIXMEs in the changes. Next week I plan to shift back to https://github.com/nextstrain/augur/pull/1454 and tweak the implementation logic before coming back here.

nextstrain_profiles/nextstrain-gisaid/builds.yaml
```diff
@@ -273,30 +273,18 @@ subsampling:
...
-      group_by: "division year month"
-      max_sequences: 200
-      max_date: "--max-date 1M"
-      exclude: "--exclude-where 'country!=China'"
-    # Early focal samples for India
-    india_early:
-      group_by: "division year month"
-      max_sequences: 200
+      group_by: "country year month"
+      group_by_weights: "defaults/population_weights.tsv"
```
victorlin, line 189: Implementation tweak is looking good and I think the current preview results are what it's going to be.

The next step here is to figure out how to define the weights file. Currently it's saved as `defaults/population_weights.tsv`, which I think is the best place to put the file. However, unlike other files in `defaults`, this isn't used by a single config key and practically not meant to be overridden (to do so would require overriding all 7 references to it in the subsampling YAML).

No better ideas are coming to mind, so suggestions are welcome.

genehack (approved):

defaults/population_weights.tsv
huddlej, line 1: I wish there was a comment in this file that mentioned how the weights were constructed. If that doesn't work with the weights file format, maybe an entry in the ncov workflow README with the example commands to generat this file would be better?

huddlej (commented):
This looks pretty good to me, @victorlin! Regarding your question:

> What can be done to check that the sampling results are appropriate?

The main pattern I'm looking for is that the pie chart sizes for Asian countries in the map view for the GISAID builds are proportional to the actual country populations. [The all-time GISAID map view](https://next.nextstrain.org/staging/ncov/gisaid/trial/victorlin-weighted-sampling-6/asia/all-time?d=map&f_region=Asia&p=full) looks consistent with the population sizes. For example, Indonesia has more samples than Japan and Malaysia but fewer than China and India. The Philippines and Vietnam have similar numbers of samples.

Regarding the workflow implementation, I really like how simple you've made this new option to group by weights. My only minor comment would be that additional documentation could help developers and users better understand how to generate these weights and use them in the workflow. For developers, a small section in the ncov README could be enough. For users, an entry in the config reference guide would be most useful.

Altogether, this is excellent, though! Thank you for the hard work you've put into pushing this along over many many months!

trvrb:
I really like where this is going! However, I ran into an error when trying to run this locally. In this filter step:

```
augur filter --metadata nextstrain-ncov-private/100k/metadata.tsv.xz
  --include defaults/include.txt
  --exclude defaults/exclude.txt
  --max-date 6M
  --exclude-where 'region!=Asia'
  --group-by country year month
  --group-by-weights defaults/population_weights.tsv
  --subsample-max-sequences 700
  --output-strains results/asia_6m/sample-asia_early.txt
WARNING: Asked to provide at most 700 sequences, but there are 1601 groups.
Sampling with weights defined by defaults/population_weights.tsv.
NOTE: Skipping 101242 groups due to lack of entries in metadata.
NOTE: Weights were not provided for the columns 'year', 'month'. Using equal weights across values in those columns.
ERROR: 27 groups appear in the metadata but are missing from the weights file. Re-run with --output-group-by-missing-weights to continue.
```
If I then run with adding `--output-group-by-missing-weights missing-weights.txt` it completes successfully with standard out:
```
WARNING: Asked to provide at most 700 sequences, but there are 1601 groups.
Sampling with weights defined by defaults/population_weights.tsv.
NOTE: Skipping 101242 groups due to lack of entries in metadata.
NOTE: Weights were not provided for the columns 'month', 'year'. Using equal weights across values in those columns.
WARNING: 27 groups appear in the metadata but are missing from the weights file. Sequences from these groups will be dropped.
All missing groups added to 'missing-weights.txt'.
91656 strains were dropped during filtering
	69341 were dropped because of 'region!=Asia'
	4791 were dropped because they were later than 2024.09 or missing a date
	1 was added back because it was in defaults/include.txt
	17525 were dropped because of subsampling criteria
695 strains passed all filters
```

The missing weights file looks like:
```
country	year	month	weight
Myanmar	2020	(2020, 4)	
Myanmar	2020	(2020, 8)	
Myanmar	2021	(2021, 1)	
```
I can see that Myanmar is missing from `population_weights.tsv`. Two things:
1. I think that ncov needs to have `--output-group-by-missing-weights`
2. Double check entries in `population_weights.tsv` to make sure all countries in our metadata are present.

I think the behavior off erroring out is maybe not the best, but I can discuss this over in https://github.com/nextstrain/augur/pull/1454.

victorlin:
@trvrb thanks for taking it for a spin, sorry that you ran into another error 😞

> the behavior off erroring out is maybe not the best, but I can discuss this over in https://github.com/nextstrain/augur/pull/1454.

I think appropriate to discuss here since your scenario is a good example.

The erroring behavior is what I intended. The weights file should be comprehensive, otherwise sequences get dropped due to lack of weights.

For this PR, that means `population_weights.tsv` should have weights for everything under `region=Asia`. I assumed running rebuild-gisaid.yml would be sufficient to test this, but that didn't include sequences from Myanmar (not sure why, I'll investigate separately).

The proper fix here is to update the country name mapping in get_population_weights with `"Burma": "Myanmar"` (your 2nd point).

> I think that ncov needs to have `--output-group-by-missing-weights`

I think differently. `--output-group-by-missing-weights` is meant more as an immediate workaround rather than something to use by default. If the ncov workflow uses this option by default, it's less likely that we'd catch any discrepancies since WARNINGs in output are not something we go back and check regularly.

Example scenario: right now we know `population_weights.tsv` is comprehensive, but sometime in the future new sequences are added under a country name that differs from what's in `population_weights.tsv`. This would still produce "successful" automated builds and one would only notice that those new sequences are being dropped if the Snakemake logs are inspected.

victorlin:
> I assumed running rebuild-gisaid.yml would be sufficient to test this, but that didn't include sequences from Myanmar (not sure why, I'll investigate separately).

I [re-ran](https://github.com/nextstrain/ncov/actions/runs/10257322062/attempts/2) rebuild-gisaid.yml and it failed with the same error. I suspect that I ran the last trial build before pushing error handling updates so it silently succeeded by dropping all Myanmar sequences.

The error message has been improved:

```
ERROR: The input metadata contains these values under the following columns that are not covered by 'defaults/population_weights.tsv':
- 'country': ['Myanmar']
Re-run with --output-group-by-missing-weights to continue.
```

This confirms that Myanmar is the only country missing from `population_weights.tsv`. I'll push a fix to add Myanmar weight, re-run trial build, and update preview links in the PR description.

defaults/population_weights.tsv
victorlin: [Updated](https://github.com/nextstrain/augur/compare/827844ff7e20cbad0c24217d3fc52d580bc5e0fa..18dbff0f06d02a757196cd38c70d266c9acdb677) the Augur PR to support comments in the weights file and [added](https://github.com/nextstrain/ncov/compare/4e5ee678e3e4f6a8256955c8c1ab99d2fccecd60..fb1a36e4bbb268236c5b042b846e9436b561a39e) comments to this file:

```
# [DO NOT EDIT] This file was generated by scripts/developer_scripts/get_population_weights
# Last updated on 2024-08-06
```

scripts/developer_scripts/get_population_weights
```diff
@@ -0,0 +1,65 @@
...
+from pathlib import Path
+import pandas as pd
+
+
+def export_world_factbook(output):
+    # Retrieve this file by going to <https://www.cia.gov/the-world-factbook/field/population/country-comparison/>,
+    # and selecting "DOWNLOAD DATA" (no filters).
+    # Ideally, this would be an automated web request. But "DOWNLOAD DATA" calls
+    # a JavaScript function to generate the download contents instead of serving
+    # the contents directly via an API.
```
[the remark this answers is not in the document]
victorlin: Thanks! I've pushed some fixes so I'm going to re-run the trial build once more with the CIA data, then try replacing it with the UN data.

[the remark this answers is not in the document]
victorlin: Added a commit to switch to UN data: 3e0945d5706aacd1a2692dde4a79cc32f2035f5a

Running trials for [open](https://github.com/nextstrain/ncov/actions/runs/10274428960) and [gisaid](https://github.com/nextstrain/ncov/actions/runs/10274433159). Results will be available at these links:

- open/asia [1m](https://next.nextstrain.org/staging/ncov/open/trial/victorlin-weighted-sampling-8/asia/1m), [2m](https://next.nextstrain.org/staging/ncov/open/trial/victorlin-weighted-sampling-8/asia/2m), [6m](https://next.nextstrain.org/staging/ncov/open/trial/victorlin-weighted-sampling-8/asia/6m), [all-time](https://next.nextstrain.org/staging/ncov/open/trial/victorlin-weighted-sampling-8/asia/all-time)
- gisaid/asia [1m](https://next.nextstrain.org/staging/ncov/gisaid/trial/victorlin-weighted-sampling-8/asia/1m), [2m](https://next.nextstrain.org/staging/ncov/gisaid/trial/victorlin-weighted-sampling-8/asia/2m), [6m](https://next.nextstrain.org/staging/ncov/gisaid/trial/victorlin-weighted-sampling-8/asia/6m), [all-time](https://next.nextstrain.org/staging/ncov/gisaid/trial/victorlin-weighted-sampling-8/asia/all-time)

> you would need to map LocID to the human-readable names

I was able to use the `Location` column for human-readable names:

> Location (string): name of the region, subregion, country or area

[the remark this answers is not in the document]
victorlin: This seems to be better. I'll use it in this PR going forwards.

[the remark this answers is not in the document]
huddlej: Nice! Bummer that the automated URL fetching still doesn't work though..."computers are terrible".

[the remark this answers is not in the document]
victorlin: Actually I realized later that it's do-able, but was hacky enough to deserve its own function:

https://github.com/nextstrain/ncov/blob/2d4af98edb6f3b1338a04aa9a26d46a58d036a87/scripts/developer_scripts/get_population_weights#L13-L27

trvrb:
Thanks for the feedback @victorlin. I have a couple thoughts:

1. In decent fraction of scenarios, I believe I'm going to want a "default" value for `population_weights.tsv`. As a simple example, I might want specify `--group-by country` and `--group-by-weights weights.tsv` with entries of
```
country weight
USA     100
default 1
```
where I don't want to have to semi-manually make sure to include every possible country as 1, just to be able to specify that I want sampling more intensively from the USA than other countries. In many situations you'll have a category like `host` where you won't be sure what exactly will come through and you can't enumerate ahead of time. If we were to add weighted sampling here https://next.nextstrain.org/avian-flu/h5n1-cattle-outbreak/genome we'd run into random new hosts continually breaking the build. I'd want a simple default to avoid this erroring.

However, I do totally see why you'd want to require an explicit `default` value rather than having an implicit default weight for entry in metadata that's not present in `weights.tsv`.

2. It feels a little strange to tie "don't error out on missing values" to `--output-group-by-missing-weights`. They seem semantically distinct to me. By giving me the option of not erroring out by including `--output-group-by-missing-weights missing-weights.txt` I wanted to set this as default.

My suggestion would be to just drop `--output-group-by-missing-weights` as an option. If there are missing values they should always be printed. I really like:
```
ERROR: The input metadata contains these values under the following columns that are not covered by 'defaults/population_weights.tsv':
- 'country': ['Myanmar']
```

This also slightly slims down the very long list of `augur filter` command line arguments. There's already a bunch of `output-` options. This also stops the user from having to run the command again with an additional argument.

So, scenarios would be
- Missing values in `weights.tsv` with no `default` set  – Error out, print to standard out the missing values that need to be specified
- Missing values in `weights.tsv` with `default` set – Proceed, but still print to standard out the missing values

victorlin:
@trvrb thanks for the example use case and suggestion – both were insightful. I've implemented in the Augur PR as https://github.com/nextstrain/augur/commit/a00d3b52037887da49fe41d15de4430c507ed267.

For the example here, the scenarios now look like:

-   ```
    ERROR: The input metadata contains these values under the following columns that are not covered by 'defaults/population_weights.tsv':
    - 'country': ['Myanmar']
    To fix this, either:
    (1) specify weights explicitly - add entries to 'defaults/population_weights.tsv' for the values above, or
    (2) specify a default weight - add an entry to 'defaults/population_weights.tsv' with the value 'default' for all columns
    ```

-   ```
    WARNING: The input metadata contains these values under the following columns that are not directly covered by 'defaults/population_weights.tsv':
    - 'country': ['Myanmar']
    The default weight of 1 will be used for all groups defined by those values.
    ...
    ```

trvrb (approved):
Really nice work here. Thanks for humoring me with the requested changes. This looks good to me from ncov perspective. I'd suggest a trial build just to make sure nothing is broken. And then maybe doing a second PR to extend weighted sampling to other regions?

victorlin:
> I'd suggest a trial build just to make sure nothing is broken.

Rebuilding the docker image to run trial builds now. Once that's done, I'll update the links in the PR description. I'll also take a look and note if anything seems off compared to the previous run.

> And then maybe doing a second PR to extend weighted sampling to other regions?

I've created an issue to track rollout of weighted sampling in this workflow: #1141

PR #1129, opened 2024-07-25
Update time based sampling

trvrb:

Currently, we focus on a recent time window in many of our ncov analyses. For example, [ncov/gisaid/north-america/6m](https://nextstrain.org/ncov/gisaid/north-america/6m), does more intensive sampling of the previous 6 months (aiming for ~4000 "recent" samples and ~1000 "early" samples). However, as the past continues to recede (as it does) we're getting more and more early context that isn't so relevant for understanding circulating diversity. Here's the live 6m tree for example:

Screenshot 2024-07-25 at 11 52 12 AM

Once selective sweeps occur, we can largely forget about past evolution when looking at current diversity. This had previously prompted us to make the ["21L" builds that root to clade 21L / lineage BA.2](https://github.com/nextstrain/ncov/pull/1029).

At this point, we're getting closer and closer to wanting the same thing with a clade 24A / lineage JN.1 rooting. However, this strategy is clearly not sustainable. In 4 more years we don't want to have have 4 different rootings, all of which require  updating and it not being clear to users what they should be looking at.

This PR addresses the issue in a simple fashion, basically making ncov work more like seasonal-flu or avian-flu where there is recent focal samples and older contextual samples, but the contextual samples only go back a year rather than many.

Here's the resulting global 6m tree:
 
Screenshot 2024-07-25 at 12 24 56 PM

Results from running this PR can be seen at:

- [global/6m](https://nextstrain.org/staging/ncov/gisaid/trial/time-based-sampling-v2/global/6m)
- [global/2m](https://nextstrain.org/staging/ncov/gisaid/trial/time-based-sampling-v2/global/2m)
- [global/1m](https://nextstrain.org/staging/ncov/gisaid/trial/time-based-sampling-v2/global/1m)
- [africa/6m](https://nextstrain.org/staging/ncov/gisaid/trial/time-based-sampling-v2/africa/6m)
etc...

This is using a 10:1 ratio of recent to early samples and doing a +1 year back for early samples. For for the 6m analysis, it's 0m to 6m back as recent focal samples and 6m to 18m back as early contextual samples.

I had tried a +2 year context as well, but it didn't seem to add much understanding while taking up additional screen real estate and additional color ramp. You can compare here however: [global/6m](https://nextstrain.org/staging/ncov/gisaid/trial/time-based-sampling/global/6m)

The the biggest worry I see here is that people currently landing at [ncov/gisaid/global/6m](https://nextstrain.org/ncov/gisaid/global/6m) can see what's currently circulating and get all the context that they may need going back to the beginning of the pandemic (with well known VOCs, etc...).

If we did merge this, we should make two showcase cards on the splash page to direct to `6m` vs `all-time` to (partially) address this. Also, if we did merge this, I'd imagine deprecating the 21L builds, where we'd remove them from the automated GitHub Actions rebuild, remove them from the manifest and add redirects to go from [ncov/gisaid/21L/global/6m](https://nextstrain.org/ncov/gisaid/21Lglobal/6m) to [ncov/gisaid/global/6m](https://nextstrain.org/ncov/gisaid/global/6m).

The other approach to this same issue would be take older clades (and conceivably Pango lineages) and make these gray while keeping the color ramp only for more recent clades. This strategy is also not mutually exclusive and we do could do both or neither. I'll try to put together a separate PR for the clade colors idea. Though even if colors update fixes things enough for the time being, I do think we'll eventually want to do something like this strategy. But it's possible this is a couple years down the road.

In addition to code review, I'd appreciate 👍 / 👎 feedback on whether you prefer this to the current sampling strategy.

## Testing

Tested locally and via GitHub Action trial builds.

## Release checklist

### DIFF
--- a/nextstrain_profiles/nextstrain-gisaid/builds.yaml
+++ b/nextstrain_profiles/nextstrain-gisaid/builds.yaml
@@ -165,19 +165,21 @@ subsampling:
   # Custom subsampling logic for regions over 1m
   # Grouping by division for North America and Oceania
   # 4000 total
-  # 4:1 ratio of recent to early
+  # 10:1 ratio of recent to early
   # 4:1 ratio of focal to context
   nextstrain_region_grouped_by_division_1m:
     # Early focal samples for region
     focal_early:
       group_by: "division year month"
-      max_sequences: 640
+      max_sequences: 256
+      min_date: "--min-date 25M"
       max_date: "--max-date 1M"
       exclude: "--exclude-where 'region!={region}'"
     # Early contextual samples from the rest of the world
     context_early:
       group_by: "country year month"
-      max_sequences: 160
+      max_sequences: 64
+      min_date: "--min-date 25M"
       max_date: "--max-date 1M"
       exclude: "--exclude-where 'region={region}'"
     # Recent focal samples for region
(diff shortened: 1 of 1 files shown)


jameshadfield:
As an experiment I cut up the global/6m tree to split out the recombinant XBB clade and drop the 22B & 22D clades which had no samples from the 6 month focus window. It wasn't as much of an improvement as I had hoped, but still better I think. Haven't sketched out how hard it would be to automate such a cut, and if we want to pursue it we can of course do it afterwards.

image

genehack (approved):

trvrb:
>The other approach to this same issue would be take older clades (and conceivably Pango lineages) and make these gray while keeping the color ramp only for more recent clades. 

After working through the coloring option in PR https://github.com/nextstrain/ncov/pull/1132 I'm definitely more of a fan of the color update. Unless there's conflicting preferences, I'll plan to just close this PR.

trvrb:
I just added the "revisit sometime" label. As time accrues leaving the older context all the way back to Wuhan will get increasingly clunky. In perhaps 6 months or a year we should implement something like this PR and include a `2y` temporal window as well.

PR #1130, opened 2024-07-26
Allow build-specific --narrow-bandwidth param in frequencies

trvrb:

This is a bug fix PR. There was a bug in the current behavior where the rule `frequencies` was calling `config["frequencies"]["narrow_bandwidth"]`. This resulted in `--narrow-bandwidth` to be fixed to whatever was specified in `parameters.yaml`. However, we were relying on build-specific settings to differentiate behavior in the `all-time` vs `6m` vs `2m` vs `1m` builds, ala:

```
frequencies
    global_1m:
      narrow_bandwidth: 0.019
```

This PR fixes this issue and allows overriding of `narrow_bandwidth` in `parameters.yaml` by build-specific settings. It also provides a genuine default (equal to augur default) so that `parameters.yaml` doesn't have to always specify `narrow_bandwidth`.

Additionally, this PR removes the build-specific setting for `recent_days_to_censor`. This setting wasn't actually being applied in a build-specific fashion and so we had been always using the workflow default of `0`. I decided to stick with this.

## Testing

I've tested locally and via trial build. Trail builds are visible at:
- [global/6m](https://next.nextstrain.org/staging/ncov/gisaid/trial/build-specific-narrow-bandwidth/global/6m)
- [global/2m](https://next.nextstrain.org/staging/ncov/gisaid/trial/build-specific-narrow-bandwidth/global/2m)
etc...

### DIFF
--- a/defaults/parameters.yaml
+++ b/defaults/parameters.yaml
@@ -137,8 +137,21 @@ ancestral:
 # Frequencies settings
 frequencies:
 
-  # min_date is set by default to 1 year before present
-  # but can be explicitly set if desired
+  # default settings that can be over-ridden for specific builds
+  default:
+
+    # min_date is set by default to 1 year before present
+    min_date: "1Y"
+
+    # max_date is set by default to present date - recent_days_to_censor
+
+    # KDE bandwidths in proportion of a year to use per strain.
+    # using 1M bandwidth by default
+    narrow_bandwidth: 0.0833
+
+
+  # settings that can be over-ridden across all builds, but not for specific builds
+  recent_days_to_censor: 0
 
   # Number of weeks between pivots
   pivot_interval: 1
@@ -146,9 +159,7 @@ frequencies:
   # Measure pivots in weeks
   pivot_interval_units: "weeks"
 
-  # KDE bandwidths in proportion of a year to use per strain.
-  # using 15 day bandwidth
-  narrow_bandwidth: 0.041
+  # Weight of KDE that uses wide bandwidth
   proportion_wide: 0.0
 
   # Diffusion frequency settings
--- a/docs/src/reference/workflow-config-file.rst
+++ b/docs/src/reference/workflow-config-file.rst
@@ -983,12 +983,29 @@ columns
 
 frequencies
 -----------
-- Valid attributes:
+-  type: object
+-  description: Parameters for specifying tip frequency calculations via ``augur frequencies``
+-  examples:
+
+.. code:: yaml
+
+   frequencies:
+     pivot_interval_units: "weeks"
+     default:
+       min_date: "6M"
+       narrow_bandwidth: 0.038
+     global_1m:
+       min_date: "1M"
+       narrow_bandwidth: 0.019
+     global_2020_to_2022:
+       min_date: "2020-01-01"
+       max_date: "2022-01-01"
+       narrow_bandwidth: 0.076
+
+Each named traits configuration (``default`` or build-named) supports specification of ``min_date``, ``max_date`` and ``narrow_bandwidth``. Other parameters can only be specified across all builds.
 
 .. contents::
    :local:
-
-.. _min_date-1:
 
 min_date
 ~~~~~~~~
--- a/nextstrain_profiles/nextstrain-gisaid/builds.yaml
+++ b/nextstrain_profiles/nextstrain-gisaid/builds.yaml
@@ -890,116 +890,87 @@ frequencies:
   reference:
     min_date: "2020-01-01"
     narrow_bandwidth: 0.038
-    recent_days_to_censor: 7
   global_1m:
     min_date: "1M"
     narrow_bandwidth: 0.019
-    recent_days_to_censor: 7
   global_2m:
     min_date: "2M"
     narrow_bandwidth: 0.019
-    recent_days_to_censor: 7
   global_6m:
     min_date: "6M"
     narrow_bandwidth: 0.038
-    recent_days_to_censor: 7
   global_all-time:
     min_date: "2020-01-01"
     narrow_bandwidth: 0.038
-    recent_days_to_censor: 7
   africa_1m:
     min_date: "1M"
     narrow_bandwidth: 0.019
-    recent_days_to_censor: 7
   africa_2m:
     min_date: "2M"
     narrow_bandwidth: 0.019
-    recent_days_to_censor: 7
   africa_6m:
     min_date: "6M"
     narrow_bandwidth: 0.038
-    recent_days_to_censor: 7
   africa_all-time:
     min_date: "2020-01-01"
     narrow_bandwidth: 0.038
-    recent_days_to_censor: 7
   asia_1m:
     min_date: "1M"
     narrow_bandwidth: 0.019
-    recent_days_to_censor: 7
   asia_2m:
     min_date: "2M"
     narrow_bandwidth: 0.019
-    recent_days_to_censor: 7
   asia_6m:
     min_date: "6M"
     narrow_bandwidth: 0.038
-    recent_days_to_censor: 7
   asia_all-time:
     min_date: "2020-01-01"
     narrow_bandwidth: 0.038
-    recent_days_to_censor: 7
   europe_1m:
     min_date: "1M"
     narrow_bandwidth: 0.019
-    recent_days_to_censor: 7
   europe_2m:
     min_date: "2M"
     narrow_bandwidth: 0.019
-    recent_days_to_censor: 7
   europe_6m:
     min_date: "6M"
     narrow_bandwidth: 0.038
-    recent_days_to_censor: 7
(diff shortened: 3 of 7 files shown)


nextstrain_profiles/nextstrain-gisaid/builds.yaml
```diff
@@ -973,116 +973,87 @@ frequencies:
   reference:
     min_date: "2020-01-01"
     narrow_bandwidth: 0.038
-    recent_days_to_censor: 7
```
victorlin, line 976: > Additionally, this PR removes the build-specific setting for `recent_days_to_censor`. This setting wasn't actually being applied in a build-specific fashion and so we had been always using the workflow default of `0`. I decided to stick with this.

Good catch. These should also be removed from nextstrain_profiles/nextstrain-open/builds.yaml.

Relatedly, it would be good to update [Workflow config file reference](https://docs.nextstrain.org/projects/ncov/en/latest/reference/workflow-config-file.html) to distinguish between settings that are "global" (directly under `frequencies`) vs. build-specific. Currently one has to inspect this Snakemake logic to determine where it can be set:

https://github.com/nextstrain/ncov/blob/4477b9654a529c4840dbe5d006113bd8c33f73c2/workflow/snakemake_rules/common.smk#L203

trvrb: Thanks for the catch with other profiles. 05b4f0f3d6191be5a0fe4a888aaf5163f502db11 removes `recent_days_to_censor` for GISAID/21L and for open.

trvrb: With 64583576fb14120688210bfa62c45fe0c61b5a05, I've updated `workflow-config-file.rst` to describe what can be over-ridden via `default` and what can't.

trvrb:
Okay. I've updated this PR to follow "strategy 2" from https://github.com/nextstrain/ncov/issues/1131 and use `default` to specify build-specific defaults. I've updated `parameters.yaml` and `workflow-config-file.rst` to spell out what parameters can go under `default` and what are global.

I've also re-triggered trial builds with https://github.com/nextstrain/ncov/actions/runs/11062321363.

@victorlin: Could you review this when you have a moment? I think it should be ready to be merged.

workflow/snakemake_rules/common.smk
```diff
@@ -207,6 +207,17 @@ def _get_max_date_for_frequencies(wildcards):
...
+def _get_narrow_bandwidth_for_wildcards(wildcards):
+    # check if builds.yaml contains frequencies:{build_name}:narrow_bandwidth
+    if wildcards.build_name in config["frequencies"] and 'narrow_bandwidth' in config["frequencies"][wildcards.build_name]:
+        return config["frequencies"][wildcards.build_name]["narrow_bandwidth"]
+    # check if parameters.yaml contains frequencies:default:narrow_bandwidth
+    elif "frequencies" in config and "narrow_bandwidth" in config["frequencies"]["default"]:
+        return config["frequencies"]["default"]["narrow_bandwidth"]
+    # else return augur frequencies default value
+    else:
+        return 0.0833
```
victorlin, line 219: _(non-blocking)_


```suggestion
```

This shouldn't be necessary since `defaults/parameters.yaml` already defines the default value and should always be used¹.

Non-blocking because `_get_min_date_for_frequencies` already has similar redundancy.

¹ under [expected usage](https://docs.nextstrain.org/projects/ncov/en/latest/guides/workflow-config-file.html#changing-parameters)

docs/src/reference/workflow-config-file.rst
```diff
@@ -983,13 +983,30 @@ columns
...
+       narrow_bandwidth: 0.038
+     global_1m:
+       min_date: "1M"
+       narrow_bandwidth: 0.019
+     global_2020_to_2022:
+       min_date: "2020-01-01"
+       max_date: "2022-01-01"
+       narrow_bandwidth: 0.076
+
+Each named traits configuration (``default`` or build-named) supports specification of ``min_date``, ``max_date`` and ``narrow_bandwidth``. Other parameters can only be specified across all builds.
```
victorlin, line 1005: _(non-blocking)_

Would it be worth allowing build-specific `recent_days_to_censor`, `pivot_interval`, etc. as well? If done right, this could reduce the duplication of functions `_get_min_date_for_frequencies`/`_get_max_date_for_frequencies`/`_get_narrow_bandwidth_for_wildcards`.

Obviously beyond the scope this PR – I could take this or at least make an issue to start it off.

victorlin (changes requested):
Looks good – one small thing and a couple non-blocking comments. I can make these changes if it helps.

victorlin (approved):

defaults/parameters.yaml
```diff
@@ -137,18 +137,29 @@ ancestral:
...
+
+    # max_date is set by default to present date - recent_days_to_censor
+
+    # KDE bandwidths in proportion of a year to use per strain.
+    # using 1M bandwidth by default
+    narrow_bandwidth: 0.0833
+
+
+  # settings that can be over-ridden across all builds, but not for specific builds
+  recent_days_to_censor: 0
```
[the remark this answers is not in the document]
trvrb: In general, I liked having workflow default in addition to the parameters.yaml default. As you point out this redundancy also exists for `frequencies.narrow_bandwidth`. 

But perhaps I understand your concern, where by having multiple layers of "defaults" it might add confusion. It would be nice if people could look just at `parameters.yaml` to understand defaults rather than digging into the workflow.

I think the original push for adding workflow redundancy here was that `parameters.yaml` (still) doesn't include `recent_days_to_censor` and if we start looking for it in the workflow and people are stilling using old `parameters.yaml` files then things will break.

I think we need a broader pattern to adhere to here (kind of like the conversion in issue https://github.com/nextstrain/ncov/issues/1131). I'm not immediately sure what's best for this specific PR.

docs/src/reference/workflow-config-file.rst
trvrb: Totally. Ideally, every entry in `parameters.yaml` would be able to be over-ridden with build-specific entries in `builds.yaml`. As you say, the current Snakemake workflow strategy here however would require padding out individual functions for each parameter. It would be wonderful to have a generic strategy that makes for automatic over-rides.

I think maybe more important that making this work for `ncov` is thinking through how to do this right across pathogens. Everyone requests the `ncov` style parameter-overrides (this came up for `mpox` most obviously). cc @joverlee521

defaults/parameters.yaml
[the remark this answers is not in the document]
victorlin: > if we start looking for it in the workflow and people are stilling using old `parameters.yaml` files then things will break.

I'm not familiar with external usage of ncov so I appreciate the thoughts here.

Since `parameters.yaml` is under version control, shouldn't it be sufficient to have workflow changes + `parameters.yaml` changes (adding `recent_days_to_censor`) in the same PR? If someone pulls the workflow changes, they'll also pull the default `recent_days_to_censor` in `parameters.yaml`. I can think of a few possibilities where breakage might happen:

1. User has made custom changes to `parameters.yaml`. `git pull` will try to auto-merge and and flag any merge conflicts.
2. User has explicitly removed this line from Snakefile: https://github.com/nextstrain/ncov/blob/82ca8d48f671349c5688a055bcc67e108c153c54/Snakefile#L42
3. User is using a profile that does not source from `parameters.yaml`, i.e. missing this line: https://github.com/nextstrain/ncov/blob/82ca8d48f671349c5688a055bcc67e108c153c54/nextstrain_profiles/nextstrain-open/config.yaml#L1-L2

    This seems the most likely for external usage, and I believe it's one of the reasons why we now recommend `--configfile` over `--profile` nowadays.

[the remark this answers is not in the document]
victorlin: > I think we need a broader pattern to adhere to here

The broader pattern seems to be having a default in `parameters.yaml` + direct access in Snakemake. This is how it's done for other config:

https://github.com/nextstrain/ncov/blob/83992903c54496513dcdf0b5acdef17d834c0ad5/workflow/snakemake_rules/main_workflow.smk#L1165-L1168

https://github.com/nextstrain/ncov/blob/83992903c54496513dcdf0b5acdef17d834c0ad5/defaults/parameters.yaml#L143-L152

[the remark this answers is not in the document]
trvrb: Thanks for the thoughts. I agree with your logic. Though I would like to push the fixes to a separate PR.

docs/src/reference/workflow-config-file.rst
joverlee521: There's definitely been a lot of discussions around build configs outside of `ncov`. I'll take some time to wrangle previous discussions and write up a summary outside of this PR (probably in pathogen-repo-guide).

Issue #1131, opened 2024-07-26
Improve / standardize the way that we pass through default parameters and build-specific parameters


trvrb:
## Context
Currently, the workflow is a bit of mess when it comes to how one can specify parameters. There seems to be 3 different approaches:

### 1 – parameters.yaml only

For the `frequencies` rule, the parameter `pivot_interval` is retrieved via `config["frequencies"]["pivot_interval"]`. This makes it grab exactly what's in `parameters.yaml` under `frequencies:pivot_interval`. If someone tries to specify a 4-week build-specific `pivot_interval` via
```
frequencies:
  global_all-time:
    pivot_interval: 4
```
The workflow will secretly just keep using the `pivot_interval` specified in `paramaters.yaml`.

Most parameters in the workflow work like this.

### 2 – builds.yaml override with necessity of `default`

For the `traits` rule, the parameter `sampling_bias_correction` is retrieved via `_get_sampling_bias_correction_for_wildcards` which is defined as
```
def _get_sampling_bias_correction_for_wildcards(wildcards):
    if wildcards.build_name in config["traits"] and 'sampling_bias_correction' in config["traits"][wildcards.build_name]:
        return config["traits"][wildcards.build_name]["sampling_bias_correction"]
    else:
        return config["traits"]["default"]["sampling_bias_correction"]
```
Ie it first looks for a build-specific list of `traits:{build_name}:sampling_bias_correction` and if doesn't find it, it expects `traits:default:sampling_bias_correction` in `builds.yaml` or `parameters.yaml`. This is described in the [docs](https://docs.nextstrain.org/projects/ncov/en/latest/reference/workflow-config-file.html#traits) as

```
   traits:
     default:
       sampling_bias_correction: 2.5
       columns: ["country"]
     washington:
       # Override default sampling bias correction for
       # "washington" build and continue to use default
       # trait columns.
       sampling_bias_correction: 5.0
```

This works, but requires `parameters.yaml` to look like:
```
traits:
  default:
    sampling_bias_correction: 2.5
    columns: ["country"]
```
with an extra `default` key compared to other entries in `parameters.yaml`. 

This strategy is only used for the `traits` rule.

### 3 – builds.yaml override without `default`

For the `frequencies` rule, the parameter `min_date` is retrieved via `_get_min_date_for_frequencies` which is defined as
```
def _get_min_date_for_frequencies(wildcards):
[...]

PR #1132, opened 2024-07-27
Gray out early clades

trvrb:

This PR is attempting to solve the same problem as #1129, but in a different fashion. In this case, we keep the subsampling in the 6m, 2m and 1m builds the same as before and so keep the long contextual tail going from 6m / 2m / 1m back to the beginning of the pandemic. However, I've added new functionality to the `assign-colors.py` script that takes in a `--clade-recency` parameter. This specifies the number of months back to look for circulating clades. So for example, `--clade-recency 6` would use tips in the previous 6 months and only keep `clade_membership` in this time period. I added parameter logic to specify `--clade-recency` of `6` vs `2` vs `1` for `6m`, `2m` and `1m` builds. The `all-time` builds include all clades.

Here's some comparisons.

#### [global/6m](https://nextstrain.org/staging/ncov/gisaid/trial/gray-out-early-clades/global/6m)

_Live_

Screenshot 2024-07-26 at 6 24 11 PM

_PR_

Screenshot 2024-07-26 at 6 25 26 PM

#### [global/2m](https://nextstrain.org/staging/ncov/gisaid/trial/gray-out-early-clades/global/2m)

_Live_

Screenshot 2024-07-26 at 6 26 20 PM

_PR_

Screenshot 2024-07-26 at 6 26 50 PM

#### [global/1m](https://nextstrain.org/staging/ncov/gisaid/trial/gray-out-early-clades/global/1m)

_Live_

Screenshot 2024-07-26 at 6 27 46 PM

_PR_

Screenshot 2024-07-26 at 6 28 20 PM

------------------------------------

I think this is a big improvement. We no longer have 3 highly similar similar shades of red for currently circulating clades. It also seems semantically appropriate to gray out these early clades to correspond to "context". This the same thing we do for regions outside of the focal region ala [south-america/6m?c=country](https://next.nextstrain.org/ncov/gisaid/south-america/6m?c=country)

My preference would be to merge this PR (after code review) and reconsider the previous time-based sampling PR #1129. I think we'll eventually need to move the `min_date` up from 2020, but with clade colors fixed it's not so urgent. Additionally, having this strategy of gray clades outside of focal window would work with alternative time-based sampling windows.

## Testing

Tested locally and via trial build (URLs linked above). However, this adds a new `parameters.yaml` option. I tested and it shouldn't break existing builds that don't specify `colors:clade_recency`, but attention here wouldn't hurt.

## Release checklist

### DIFF
--- a/docs/src/reference/workflow-config-file.rst
+++ b/docs/src/reference/workflow-config-file.rst
@@ -941,6 +941,33 @@ no_timetree
 -  description: Do not produce a time tree.
 -  default: ``false``
 
+colors
+------
+
+-  type: object
+-  description: Parameters for assigning colors in ``scripts/assign-colors.py``
+-  examples:
+
+.. code:: yaml
+
+   colors:
+     default:
+       clade_recency: ""
+     global-6m:
+       # Override clade recency colors for "global-6m" build
+       clade_recency: 6
+
+Each named traits configuration (``default`` or build-named) supports the following attributes:
+
+.. contents::
+   :local:
+
+clade_recency
+~~~~~~~~~~~~~
+
+-  type: integer
+-  description: if provided, restrict to clades found in tree within X months of present
+-  default: ````
 
 traits
 ------
--- a/docs/src/reference/workflow-config-file.rst
+++ b/docs/src/reference/workflow-config-file.rst
@@ -952,7 +952,7 @@ colors
 
    colors:
      default:
-       clade_recency: ""
+       clade_recency: "all"
      global-6m:
        # Override clade recency colors for "global-6m" build
        clade_recency: 6
@@ -966,8 +966,8 @@ clade_recency
 ~~~~~~~~~~~~~
 
 -  type: integer
--  description: if provided, restrict to clades found in tree within X months of present
--  default: ````
+-  description: if integer value is provided, restrict to clades found in tree within X months of present
+-  default: ``all``
 
 traits
 ------
--- a/docs/src/reference/workflow-config-file.rst
+++ b/docs/src/reference/workflow-config-file.rst
@@ -955,7 +955,7 @@ colors
        clade_recency: "all"
      global-6m:
        # Override clade recency colors for "global-6m" build
-       clade_recency: 6
+       clade_recency: "6M"
 
 Each named traits configuration (``default`` or build-named) supports the following attributes:
 
@@ -965,9 +965,10 @@ Each named traits configuration (``default`` or build-named) supports the follow
 clade_recency
 ~~~~~~~~~~~~~
 
--  type: integer
--  description: if integer value is provided, restrict to clades found in tree within X months of present
--  default: ``all``
+-  type: string
+-  format: `ISO 8601 <https://en.wikipedia.org/wiki/ISO_8601#Durations>`__ duration with optional ``P`` prefix (e.g. ``2M``, ``18M``, ``1Y6M``)
+-  description: restrict to clades found in tree within this duration from present
+-  default: ``all`` (no restriction)
 
 traits
 ------
--- a/nextstrain_profiles/nextstrain-gisaid/builds.yaml
+++ b/nextstrain_profiles/nextstrain-gisaid/builds.yaml
@@ -332,47 +332,47 @@ colors:
   default:
     clade_recency: "all"
   global_1m:
-    clade_recency: 1
+    clade_recency: "1M"
   global_2m:
-    clade_recency: 2
+    clade_recency: "2M"
   global_6m:
-    clade_recency: 6
+    clade_recency: "6M"
   africa_1m:
-    clade_recency: 1
+    clade_recency: "1M"
   africa_2m:
-    clade_recency: 2
+    clade_recency: "2M"
   africa_6m:
-    clade_recency: 6
+    clade_recency: "6M"
   asia_1m:
-    clade_recency: 1
+    clade_recency: "1M"
   asia_2m:
-    clade_recency: 2
+    clade_recency: "2M"
   asia_6m:
-    clade_recency: 6
+    clade_recency: "6M"
   europe_1m:
-    clade_recency: 1
+    clade_recency: "1M"
   europe_2m:
-    clade_recency: 2
+    clade_recency: "2M"
   europe_6m:
-    clade_recency: 6
+    clade_recency: "6M"
   north-america_1m:
-    clade_recency: 1
+    clade_recency: "1M"
   north-america_2m:
-    clade_recency: 2
+    clade_recency: "2M"
   north-america_6m:
-    clade_recency: 6
+    clade_recency: "6M"
   oceania_1m:
-    clade_recency: 1
+    clade_recency: "1M"
   oceania_2m:
-    clade_recency: 2
+    clade_recency: "2M"
   oceania_6m:
-    clade_recency: 6
+    clade_recency: "6M"
   south-america_1m:
-    clade_recency: 1
+    clade_recency: "1M"
   south-america_2m:
-    clade_recency: 2
+    clade_recency: "2M"
   south-america_6m:
-    clade_recency: 6
+    clade_recency: "6M"
 
 # if different traits should be reconstructed for some builds, specify here
 # otherwise the default trait config in defaults/parameters.yaml will used
--- a/nextstrain_profiles/nextstrain-gisaid/builds.yaml
+++ b/nextstrain_profiles/nextstrain-gisaid/builds.yaml
@@ -329,7 +329,7 @@ subsampling:
 # if different traits should be reconstructed for some builds, specify here
 # otherwise the default trait config in defaults/parameters.yaml will used
 colors:
-  reference:
+  default:
     clade_recency: "all"
   global_1m:
     clade_recency: 1
@@ -337,56 +337,42 @@ colors:
     clade_recency: 2
   global_6m:
     clade_recency: 6
-  global_all-time:
-    clade_recency: "all"
   africa_1m:
     clade_recency: 1
   africa_2m:
     clade_recency: 2
   africa_6m:
     clade_recency: 6
-  africa_all-time:
-    clade_recency: "all"
   asia_1m:
     clade_recency: 1
   asia_2m:
     clade_recency: 2
   asia_6m:
     clade_recency: 6
-  asia_all-time:
-    clade_recency: "all"
   europe_1m:
     clade_recency: 1
   europe_2m:
     clade_recency: 2
   europe_6m:
     clade_recency: 6
-  europe_all-time:
-    clade_recency: "all"
   north-america_1m:
     clade_recency: 1
   north-america_2m:
     clade_recency: 2
   north-america_6m:
     clade_recency: 6
-  north-america_all-time:
-    clade_recency: "all"
   oceania_1m:
     clade_recency: 1
   oceania_2m:
     clade_recency: 2
   oceania_6m:
     clade_recency: 6
-  oceania_all-time:
-    clade_recency: "all"
   south-america_1m:
     clade_recency: 1
   south-america_2m:
     clade_recency: 2
   south-america_6m:
     clade_recency: 6
-  south-america_all-time:
-    clade_recency: "all"
 
 # if different traits should be reconstructed for some builds, specify here
 # otherwise the default trait config in defaults/parameters.yaml will used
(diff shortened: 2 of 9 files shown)


docs/src/reference/workflow-config-file.rst
```diff
@@ -936,6 +936,33 @@ no_timetree
...
+
+-  type: object
+-  description: Parameters for assigning colors in ``scripts/assign-colors.py``
+-  examples:
+
+.. code:: yaml
+
+   colors:
+     default:
+       clade_recency: ""
```
huddlej, line 950: I think there is a mismatch between these docs and the code implementation. In the workflow code, the `_get_clade_recency_for_wildcards` function looks for `config["colors"]["clade_recency"]` as the default value instead of a `default` key.

I slightly prefer the `default` key over `clade_recency`.

scripts/assign-colors.py
```diff
@@ -15,6 +25,7 @@
     parser.add_argument('--color-schemes', type=str, required=True, help="input color schemes file")
     parser.add_argument('--metadata', type=str, help="if provided, restrict colors to only those found in metadata")
     parser.add_argument('--clade-node-data', type=str, help="if provided, restrict to only those clades found in tree")
+    parser.add_argument('--clade-recency', type=int, nargs='?', const=None, help="if provided, restrict to clades found in tree within X months of present")
```
huddlej, line 28: If you opted for the approach I described above of building a `params.clade_recency_argument` in the workflow, you could drop the `nargs='?', const=None,` arguments here, since that's already the default behavior for optional arguments.

huddlej (commented):
I agree this looks like a great improvement! I'm excited to try the same approach in the seasonal flu workflow.

My main comment is that I think the default value should be an explicit value that ties to the meaning of the `clade_recency` functionality instead of an empty string. For example, a string value of “all” or “all-time” would more clearly communicate what the config option is doing. An empty string doesn’t immediately communicate “assign colors to all clades” and looks more like a typo in the config YAML (like maybe someone forgot to enter a value?).

The clade recency argument appears in two different parts of the workflow: the workflow configuration/rules and the `assign-colors.py` script. We could modify either of these parts to support a default argument of "all".

In the context of the script, we could keep the default behavior of “assign colors to all clades” when the `--clade-recency` argument is not provided and that would not be surprising behavior. Providing that argument would be a way to opt into the behavior of filtering by date. When the user provides that argument, the script would require that the value is an integer representing number of months in the past. In the context of the workflow, this approach would require us to build a clade recency argument string to pass to the script. So, instead of this:

```
--clade-recency {params.clade_recency}
```

we would build this:

```
{params.clade_recency_argument}
```

When the user provides a value of “all” or the workflow can’t find a clade recency in the config, the clade recency argument is an empty string. When the user provides an integer value, the clade recency argument is `--clade-recency 3`, for example. We already use this pattern in the ncov workflow (like [the get_priority_argument function](https://github.com/nextstrain/ncov/blob/4477b9654a529c4840dbe5d006113bd8c33f73c2/workflow/snakemake_rules/main_workflow.smk#L153-L163)).

Another pattern we use elsewhere is a custom argparse type that accepts a specific string value (“all”) or an integer [like we do for Augur’s nthreads arguments](https://github.com/nextstrain/augur/blob/2c882986bc9ed7cf683546ebab96250e07b9321d/augur/utils.py#L525-L536). We could use this pattern to modify the `assign-colors` script to accept `--clade-recency all` as a valid input. Then the workflow can keep the same pattern in this PR where the `colors` rule hardcodes the `--clade-recency` argument with a `{params.clade_recency}` reference instead of building the argument as described above. When the user provides a value of `all`, the script happily performs the default behavior of assigning colors to all clades.

I lean slightly toward the `params.clade_recency_argument` pattern, since it doesn’t add complexity to the `assign-colors` script that would only exist for the sake of the workflow, but either approach would work.

scripts/assign-colors.py
trvrb: Awesome! Resolved in 497cb9ce63bbf92dc400994463b367dfac7df6bd.

trvrb:
Thanks so much for the detailed feedback @huddlej. This was extremely helpful. With 497cb9ce63bbf92dc400994463b367dfac7df6bd, I've implemented your first suggestion of swapping `--clade-recency {params.clade_recency}` to `{params.clade_recency_argument}`.

docs/src/reference/workflow-config-file.rst
trvrb: In 82ca8d48f671349c5688a055bcc67e108c153c54 I've switched behavior to use `config["colors"]["default"]["clade_recency"]` when looking for a default value. @huddlej: Your preference of `default` over `clade_recency` matches @joverlee521, @victorlin and @jameshadfield's preferences expressed in https://github.com/nextstrain/ncov/issues/1131.

trvrb:
I've (finally) attended to @huddlej's helpful comments. This included swapping to `{params.clade_recency_argument}` and swapping to `config["colors"]["default"]["clade_recency"]`.

I'm re-running trial builds in https://github.com/nextstrain/ncov/actions/runs/11059102557. This should update the URLs referenced in the original description like https://nextstrain.org/staging/ncov/gisaid/trial/gray-out-early-clades/global/6m.

The resulting dataset coloring should be the same as the original PR, but just with a cleaner interface.

@victorlin: I've tagged you for review here because I know that @huddlej is very busy with VCM and Options. If you could look things over, I'd very much appreciate it.

scripts/assign-colors.py
```diff
@@ -15,6 +25,7 @@
     parser.add_argument('--color-schemes', type=str, required=True, help="input color schemes file")
     parser.add_argument('--metadata', type=str, help="if provided, restrict colors to only those found in metadata")
     parser.add_argument('--clade-node-data', type=str, help="if provided, restrict to only those clades found in tree")
+    parser.add_argument('--clade-recency', type=int, help="if provided, restrict to clades found in tree within X months of present")
```
victorlin, line 28: _(non-blocking suggestion)_

How about encoding the unit somewhere in the option name or value? Examples:

1. `clade_recency_months: 6`
2. `clade_recency: 6M`

I'd lean towards (2) since a proper implementation would also allow different units e.g. `clade_recency: 1Y`.

If that sounds reasonable, I'd be happy to add a commit that does this – it'd be similar to https://github.com/nextstrain/augur/pull/740/commits/873a70a34105fd6579d2780d4dba946a1524363c

victorlin (commented):
Didn't get to review much but sending in a comment before I sign off for the day.

Will plan to continue review on Monday.

scripts/assign-colors.py
trvrb: Ah! Option 2 is a much nicer interface and would fit with usage of `min_date: "6M"`, etc... elsewhere in the config. If you're up for adding this commit that would be amazing.

victorlin: Added as 2bf39d32ceae17d13a73373314dc3f4c9b3bad68 and started a new [trial build](https://github.com/nextstrain/ncov/actions/runs/11113055010).

nextstrain_profiles/nextstrain-gisaid/builds.yaml
victorlin, line 1: Should the same changes be applied to the open and gisaid-21L profiles?

trvrb: Good call. I just added these in d517a5ff2a75be871c9a2778c8aebe6c07393560.

scripts/assign-colors.py
trvrb: Awesome. Thank you!

victorlin: Trial build did not add gray clades because I forgot an important part: 024ae9474ce79f584fa75efb70338250e1d00bf1

Started a new [trial build](https://github.com/nextstrain/ncov/actions/runs/11116193172).

trvrb: Hmm.... It looks like it's still not working quite correctly. https://nextstrain.org/staging/ncov/gisaid/trial/gray-out-early-clades-2/global/6m. Not exactly sure what went wrong.

victorlin: I noticed this too. Will investigate locally.

victorlin: 🤦 I updated `assign-colors.py` but not the actual workflow. Updated with ea85954ffeb6eb77c962fd06f1b311380cdb26e6 and [re-running](https://github.com/nextstrain/ncov/actions/runs/11133100504) with the same trial name `gray-out-early-clades-2`.

victorlin:
I pushed a couple more fixes: b685fbbb93ec157176d5ffad199c6071485b36fc, 3273dcc9cba8c5de0c167c245e99ff90db4d6d4d

I also contemplated renaming `clade_recency` to `clade_min_date` to align with other `min_date` options that take the same duration format, but `clade_min_date: "all"` doesn't sound right. The special value `"all"` could be replaced by a blank/`null` value (i.e. `clade_min_date:`) but that goes against @huddlej's point in https://github.com/nextstrain/ncov/pull/1132#pullrequestreview-2211150127.

scripts/assign-colors.py
victorlin: Still didn't work. I finally figured out how to test locally to check that `--clade-recency` is being passed through properly (should have done this much earlier):

```console
$ nextstrain build . --profile nextstrain_profiles/nextstrain-gisaid results/global_6m/colors.tsv --dry-run
...
        python3 scripts/assign-colors.py \
            --ordering defaults/color_ordering.tsv \
            --color-schemes defaults/color_schemes.tsv \
            --output results/global_6m/colors.tsv \
            --clade-node-data results/global_6m/clades.json \
            --clade-recency 6M \
            --metadata results/global_6m/metadata_adjusted.tsv.xz 2>&1 | tee logs/colors_global_6m.txt
...
```

Pushed the fix as 0b907ff3216ea8d76bb214dddc089a2ad1e2df41 and [re-running](https://github.com/nextstrain/ncov/actions/runs/11149144810) with the same trial name `gray-out-early-clades-2`.

victorlin: Thanks, noted for the future. The trial build completed and seems to be good: https://nextstrain.org/staging/ncov/gisaid/trial/gray-out-early-clades-2/global/6m

workflow/snakemake_rules/common.smk
```diff
@@ -168,6 +186,27 @@ def _get_metadata_by_wildcards(wildcards):
...
+def _get_clade_recency_for_wildcards(wildcards):
+    # check if builds.yaml contains colors:{build_name}:clade_recency
+    if wildcards.build_name in config["colors"] and 'clade_recency' in config["colors"][wildcards.build_name]:
+        return config["colors"][wildcards.build_name]["clade_recency"]
+    # check if builds.yaml or parameters.yaml contains colors:default:clade_recency
+    elif "colors" in config and "clade_recency" in config["colors"]["default"]:
+        return config["colors"]["default"]["clade_recency"]
+    # else return sensible default value
+    else:
+        return "all"
```
victorlin, line 198: _non-blocking: I've already picked on this [elsewhere](https://github.com/nextstrain/ncov/pull/1130#discussion_r1781807136) and don't want implementation details to get in the way of having improved coloring_

In #1131 we settled on "strategy 2" which would translate to:

```py
def _get_clade_recency_for_wildcards(wildcards):
    # check if builds.yaml contains colors:{build_name}:clade_recency
    if wildcards.build_name in config["colors"] and 'clade_recency' in config["colors"][wildcards.build_name]:
        return config["colors"][wildcards.build_name]["clade_recency"]
    # use workflow default
    else:
        return config["colors"]["default"]["clade_recency"]
```

victorlin (approved):
Looks good. I think all the commits can get squashed into one, or at least all my commits squashed into one along the lines of "Use ISO duration format for clade recency, error on invalid values".

workflow/snakemake_rules/common.smk
trvrb: Thanks for the catch here. I agree that cleaning up default values to source directly from `parameters.yaml` is the right way to do things. But that said, I think I'd like to tackle this cleanup in a separate PR after this PR and #1130 are merged.

trvrb:
Thanks for the review @victorlin! Could you squash your commits however you'd like and I'll then update `change_log.md` and merge.

victorlin:
@trvrb done. Also rebased onto latest `master` to get the latest change log.

trvrb:
Thanks again @victorlin!

PR #1151, opened 2024-08-26
Use weighted sampling other builds

### LINKED ISSUE #1141: Use weighted sampling
> [!NOTE]
> https://github.com/nextstrain/ncov/pull/1106 came first. This is a higher level summary written after some design discussions happened in that PR.

## Context

There is much sampling bias in SARS-CoV-2 data. One of the goals of this workflow is to produce datasets that are representative of real-world incidence, stripping away as much sampling bias as possible.

Currently, this is approximated by sampling with various `group_by`s - a combination of geographic (`division`/`country`) and temporal (`month`/`week`) attributes - to define groups that are then uniformly sampled based on a target `max_sequences`.

The need for uniform sampling at the group level is an inherent limitation of `augur filter`. It has prompted workarounds in this workflow such as https://github.com/nextstrain/ncov/pull/1074.

## Proposal

There is a proposal to remove the limitation of `augur filter`: https://github.com/nextstrain/augur/issues/1318. The option to specify sampling weights could be directly used in this workflow. Population-based weighted sampling would bring this workflow one step closer to representing real-world incidence, though there will still be some inherent sampling bias¹.

Case count data was also considered as a potential source of weights, however it was determined that population data would be a better source. See discussion: https://github.com/nextstrain/ncov/pull/1106#discussion_r1596112762

¹ weighted target sizes are calculated without taking into account the actual number of sequences available per group. This means under-sampled countries would still be under-sampled, resulting in fewer total sequences than requested by `max_sequences`. This is already the case with current uniform sampling, but it may be more noticeable under population-based weighted sampling for large countries that are under-sampled.

## Progress

- [x] https://github.com/nextstrain/augur/issues/1318
- [x] https://github.com/nextstrain/ncov/pull/1106
    - I chose these builds to directly address the workaround applied in https://github.com/nextstrain/ncov/pull/1074
- [ ] https://github.com/nextstrain/ncov/pull/1150
- [ ] https://github.com/nextstrain/ncov/pull/1151

victorlin:

Extend the weighted sampling approach from Asia builds to all other builds.

## Related issue(s)

Closes #1141

## Testing

- [ ] Run trial builds

## Release checklist


trvrb:
This is great @victorlin! Please go ahead and plan to rebase and merge whenever you'd like. I did want to note for the future that the switch to country-level only for North America has resulted in US states with more samples getting over-represented relative to population size. Here's trail build: https://nextstrain.org/staging/ncov/gisaid/trial/victorlin-all-builds-weighted/north-america/6m?f_region=North%20America

Screenshot 2024-09-26 at 4 32 31 PM

Here's the comparison to previous behavior:

Screenshot 2024-09-26 at 4 32 59 PM

Overall, I think it's a big improvement to have country like Costa Rica with many admin divisions get down-weighted to balance its population size. 

We could consider implementing division population weights at some point in the future, but I wouldn't worry too much about it immediately. As it stands, states with large population sizes like CA, NY and TX are tending to submit more sequences anyway so the resulting sampling maybe more representative anyway.

trvrb (approved):

Issue #1157, opened 2024-10-17
Migrate to Nextclade v3


joverlee521:
[Migration to Nextclade v3 guide](https://docs.nextstrain.org/projects/nextclade/en/stable/user/migration-v3.html) for reference.

## TODOs

- [x] remove `use_nextalign` config from build.yaml files (the use of the config was removed in https://github.com/nextstrain/ncov/pull/630)
- [x] ~update [mutation_summary.py](https://github.com/nextstrain/ncov/blob/7e594bc85ac82c4d499555f5288ffd6628dbd141/scripts/mutation_summary.py) to work with Nextclade v3 outputs~ Remove mutation_summary.py and related rules
- [x] update nextclade2/nextalign commands in workflow rules
- [x] remove references to "nextalign" in docs
- [x] update nextclade dependencies [workflow/envs/nextstrain.yaml](https://github.com/nextstrain/ncov/blob/7e594bc85ac82c4d499555f5288ffd6628dbd141/workflow/envs/nextstrain.yaml#L10-L11)
- [ ] major release of the workflow

PR #1160, opened 2024-10-23
Migrate to Nextclade v3

joverlee521:

Main goal is to migrate to Nextclade v3. I discovered Nextclade related params and rules that were unused, so this PR also removes them to make the migration easier. 

See commits for details. 

## Related issue(s)

Resolves https://github.com/nextstrain/ncov/issues/1157

## Release checklist


workflow/snakemake_rules/main_workflow.smk
```diff
@@ -503,13 +493,12 @@ rule build_align:
...
             --jobs {threads} \
             --input-dataset {input.nextclade_dataset} \
             --output-tsv {output.nextclade_qc} \
             --output-fasta {output.alignment} \
-            --output-translations {params.output_translations} \
-            --output-insertions {output.insertions} 2>&1 | tee {log}
+            --output-translations {params.output_translations} 2>&1 | tee {log}
         """
 
 rule join_metadata_and_nextclade_qc:
```
joverlee521, line 504: I was tempted to update this rule to use `augur merge` to join the metadata and Nextclade data. However, the custom `join-metadata-and-clades.py` script includes [calculations for `clock_deviation`](https://github.com/nextstrain/ncov/blob/66aa9be182df05f04acc5a2cbb4ad750b39eb0ec/scripts/join-metadata-and-clades.py#L125-L143) that I don't really want to touch on here. Will leave that for another time...

joverlee521:
> Emma used mutation summary in covariants back in the day. I am not sure if it is still used. If file upload was removed, then probably not. Or she could be running it outside of the official AWS-based flow.

Thanks for the reminder! I found an old [Slack thread](https://bedfordlab.slack.com/archives/C01LCTT7JNN/p1647538476963439?thread_ts=1647353315.650509&cid=C01LCTT7JNN) that confirms @emmahodcroft no longer used the mutation summary produced by ncov-ingest, but I could not find anything regarding the mutation summary produced here. 

> Without annotation it might produce lower quality alignments and change what mutations are detected, so I think it's better to keep annotations, even if translation is not being run. If so decided, then might also add a comment or a description that even though annotation appears unused, it is in fact used for codon-aware nucleotide alignment (and perhaps a link to the docs):
> 
> https://docs.nextstrain.org/projects/nextclade/en/stable/user/algorithm/01-sequence-alignment.html

Thanks for the pointer! That's good to know. I'll add back annotations.

docs/src/reference/workflow-config-file.rst
```diff
@@ -513,17 +513,17 @@ alignment_reference
...
 -  default: ``defaults/reference_seq.fasta``
--  used in rules: ``align``, ``proximity_score`` (subsampling), ``build_align``, ``build_mutation_summary``
+-  used in rules: ``align``, ``proximity_score`` (subsampling)
 
 annotation
 ~~~~~~~~~~
 
 -  type: string
--  description: Path to a GFF-formatted annotation of gene coordinates (e.g., a “gene map”) for use by ``nextalign`` and mutation summaries.
+-  description: Path to a GFF-formatted annotation of gene coordinates (e.g., a “gene map”) for use by ``nextalign`` for codon-aware alignment.
```
genehack, line 524: ?? 

```suggestion
-  description: Path to a GFF-formatted annotation of gene coordinates (e.g., a “gene map”) for use by ``nextclade`` for codon-aware alignment.
```

genehack (approved):

docs/src/reference/workflow-config-file.rst
joverlee521: Ah, good catch! I missed this during rebasing.

joverlee521: Fixed in [force-push](https://github.com/nextstrain/ncov/compare/c5ca8fee2c60a14786d51716ed90e8795bd536c9..f269e69b4921f93c37a0964accd4bdfc418ea1c5)

PR #1161, opened 2024-11-16
Use weighted sampling for recent samples in global builds

victorlin:
This should have been a part of #1106 but got lost in translation.

[GISAID trial build](https://github.com/nextstrain/ncov/actions/runs/11866290700) links:

- [global/1m](https://nextstrain.org/staging/ncov/gisaid/trial/victorlin-weight-global-recent/global/1m)
- [global/2m](https://nextstrain.org/staging/ncov/gisaid/trial/victorlin-weight-global-recent/global/2m)
- [global/6m](https://nextstrain.org/staging/ncov/gisaid/trial/victorlin-weight-global-recent/global/6m)

### DIFF
--- a/nextstrain_profiles/nextstrain-gisaid-21L/builds.yaml
+++ b/nextstrain_profiles/nextstrain-gisaid-21L/builds.yaml
@@ -294,6 +294,7 @@ subsampling:
       max_date: "--max-date 1M"
     recent:
       group_by: "country week"
+      group_by_weights: "defaults/population_weights.tsv"
       max_sequences: 4100
       min_date: "--min-date 1M"
 
@@ -308,6 +309,7 @@ subsampling:
       max_date: "--max-date 2M"
     recent:
       group_by: "country week"
+      group_by_weights: "defaults/population_weights.tsv"
       max_sequences: 4100
       min_date: "--min-date 2M"
 
@@ -322,6 +324,7 @@ subsampling:
       max_date: "--max-date 6M"
     recent:
       group_by: "country year month"
+      group_by_weights: "defaults/population_weights.tsv"
       max_sequences: 4100
       min_date: "--min-date 6M"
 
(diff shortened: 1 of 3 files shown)


victorlin:
Merging without review since this seems like a trivial fix.

corneliusroemer:
I think this breaks things: the number of recent samples collapsed from ~4k to ~1.2k after the merge

Before:
Google Chrome Beta 2024-12-20 13 37 15
https://next.nextstrain.org/ncov/gisaid/global/1m@2024-12-17

After:
Google Chrome Beta 2024-12-20 13 37 26
https://next.nextstrain.org/ncov/gisaid/global/1m@2024-12-18

In retrospect this isn't a trivial change, I hope you agree, trial build would have been good :)

Reverting without review, hope that makes sense given this was merged without review and trial builds and people might be heading towards Christams.

victorlin:
Thanks for catching and addressing this @corneliusroemer. I added trial build links in the PR description but should have looked at the numbers before merging.

victorlin:
I've fixed the PR description to link to the commit that this addressed. Note that prior to https://github.com/nextstrain/ncov/pull/1151/commits/ae764bb8ec845b4bd327a46175481fec6e37d10b, the number of samples for gisaid/global/1m was ~2.7k. See [before @2024-09-30](https://next.nextstrain.org/ncov/gisaid/global/1m@2024-09-30) vs. [after @2024-10-01](https://next.nextstrain.org/ncov/gisaid/global/1m@2024-10-01). The difference is that ae764bb8ec845b4bd327a46175481fec6e37d10b changed from region-level weights encoded via separate `augur filter` calls to a single `augur filter` call with uniform sampling across countries. This should have been paired with weighted sampling across countries, which is what this PR added. We're now back to what seems to be an incomplete shift towards country-level weights.

@corneliusroemer can you check whether 2024-09-30 or 2024-10-01 seems to make more sense? It would be interesting if you say 2024-10-01 – in that case we may want to consider removing weighted sampling for all 1m/2m builds.

PR #1170, opened 2025-01-28
feat: use nightly nextclade tree

ivan-aksamentov:

- [x] switch Nextclade dataset to directory format (which allows to replace dataset files)
- [x] replace reference tree in the dataset with the nightly tree from https://nextstrain.org/staging/nextclade/sars-cov-2

This allows to bypass laggy Nextclade dataset updates and use the latest data always. Which may or may not be what we want.

This aims to be a workaround until the dataset updates are sorted out.

Potential problems:
- nightly trees are not systematically reviewed and can contain bugs
- does any other parts of the dataset need to be updated along with the tree? (such as pathogen.json)
- does any other repos need to be updated to use nightly tree? (e.g. ncov-ingest) i.e. is there an assumption that the exact same dataset is used in 2 or more places?

## Related issue(s)

None I could find

## Testing

I need help with this

## Release checklist

### DIFF
--- a/workflow/snakemake_rules/main_workflow.smk
+++ b/workflow/snakemake_rules/main_workflow.smk
@@ -455,7 +455,7 @@ rule prepare_nextclade:
         Downloading reference files for nextclade (used for alignment and qc).
         """
     output:
-        nextclade_dataset = "data/sars-cov-2-nextclade-defaults",
+        nextclade_dataset = directory("data/sars-cov-2-nextclade-defaults"),
     params:
         name = config["nextclade_dataset"],
     conda: config["conda_environment"]


joverlee521:
> does any other repos need to be updated to use nightly tree? (e.g. ncov-ingest)

We would need to update ncov-ingest so that the latest lineages are included in the metadata.tsv on S3, which is used to create the inputs for model results on https://nextstrain.org/sars-cov-2/forecasts. 

There's some data provenance tracking (previously discussed in https://github.com/nextstrain/ncov-ingest/issues/458) that would get hairy with just overriding the tree.json in the dataset. The ncov-ingest caching system depends on the Nextclade + dataset versions, so if we use the nightly tree, would we drop the cache and run all sequences to Nextclade every day?

PR #1171, opened 2025-01-29
Update lineage colors

trvrb:
Ordering is taken directly from https://github.com/cov-lineages/pango-designation/blob/master/lineage_notes.txt. This should properly account for aliasing.

Trial builds from https://github.com/nextstrain/ncov/actions/runs/13038164034:

https://nextstrain.org/staging/ncov/gisaid/trial/lineage-colors/global/6m?c=Nextclade_pango

6m


trvrb:
Trial runs look good. I'm going to go ahead and merge this.

Issue #1173, opened 2025-02-26
Automated GISAID builds failed due to Slack outage


joverlee521:
Automated GISAID builds failed this morning:
- [Rebuild GISAID](https://github.com/nextstrain/ncov/actions/runs/13545237303)
- [Rebuild GISAID 21L](https://github.com/nextstrain/ncov/actions/runs/13545237307)

Most likely due to the ongoing [Slack outage](https://slack-status.com/2025-02/1b757d1d0f444c34). 
Error message shows the lack of `ts` in the Slack response. 
```
[batch] [2025-02-26T15:33:47+00:00] Error in rule deploy_single:
[batch] [2025-02-26T15:33:47+00:00]     jobid: 0
[batch] [2025-02-26T15:33:47+00:00]     input: auspice/ncov_gisaid_21L_oceania_2m.json, auspice/ncov_gisaid_21L_oceania_2m_tip-frequencies.json, auspice/ncov_gisaid_21L_oceania_2m_root-sequence.json, auspice/ncov_gisaid_21L_oceania_2m_2025-02-26.json, auspice/ncov_gisaid_21L_oceania_2m_2025-02-26_tip-frequencies.json, auspice/ncov_gisaid_21L_oceania_2m_2025-02-26_root-sequence.json
[batch] [2025-02-26T15:33:47+00:00]     output: auspice/oceania_2m.deployed
[batch] [2025-02-26T15:33:47+00:00] RuleException:
[batch] [2025-02-26T15:33:47+00:00] KeyError in file /nextstrain/build/workflow/snakemake_rules/export_for_nextstrain.smk, line 444:
[batch] [2025-02-26T15:33:47+00:00] 'ts'
[batch] [2025-02-26T15:33:47+00:00]   File "/nextstrain/build/workflow/snakemake_rules/export_for_nextstrain.smk", line 378, in __rule_deploy_single
[batch] [2025-02-26T15:33:47+00:00]   File "/nextstrain/build/workflow/snakemake_rules/export_for_nextstrain.smk", line 444, in send_slack_message
[batch] [2025-02-26T15:33:47+00:00]   File "/usr/local/lib/python3.10/concurrent/futures/thread.py", line 58, in run
```

joverlee521:
I'll let this sit for now and re-run the builds once the Slack issue is resolved. 

[...]

PR #1174, opened 2025-02-26
export_for_nextstrain.smk: catch Slack errors

joverlee521:

The Slack messages are just a nice-to-have feature and should not cause the entire workflow to crash if there is an error.

## Related issue(s)

Prompted by <https://github.com/nextstrain/ncov/issues/1173#issuecomment-2685908873>


workflow/snakemake_rules/export_for_nextstrain.smk
```diff
@@ -439,9 +439,12 @@ def send_slack_message(message, broadcast=False):
...
 
-    response = requests.post("https://slack.com/api/chat.postMessage", headers=headers, data=json.dumps(data))
-    response.raise_for_status()
-    storage.store_if_not_present("slack_thread_ts", response.json()["ts"])
+    try:
+        response = requests.post("https://slack.com/api/chat.postMessage", headers=headers, data=json.dumps(data))
+        response.raise_for_status()
+        storage.store_if_not_present("slack_thread_ts", response.json()["ts"])
+    except Exception as error:
+        print("An error occurred when sending Slack message:", error)
```
tsibley, line 447: A full traceback is probably more helpful than just the stringification of the error (which is unfortunately often very terse/uninformative).

```suggestion
        print("An error occurred when sending Slack message:", file=sys.stderr)
        traceback.print_exc()
        print("Oh well. Ignoring and moving on…", file=sys.stderr)
```

(Needs an import of `traceback` and maybe `sys` at the top.)

tsibley (approved):

workflow/snakemake_rules/export_for_nextstrain.smk
joverlee521: Updated in bf64d625eb9ffc9987816c674685bded53211d1b

PR #1178, opened 2025-06-27
Optimize `rule combine_samples`

victorlin:

This rule became [noticeably slower](https://github.com/nextstrain/augur/issues/1833) with Augur 31.2.1. I tracked down a few reasons. This PR along with [Augur #1834](https://github.com/nextstrain/augur/pull/1834) should improve the run time of this rule to be faster than with previous versions of Augur.

## Testing

Testing notes in https://github.com/nextstrain/augur/issues/1833#issuecomment-3011435623

## Checklist

- [ ] Merge+release of https://github.com/nextstrain/augur/pull/1834
- [ ] Document new minimum Augur version requirement

## Release checklist

### DIFF
--- a/workflow/snakemake_rules/main_workflow.smk
+++ b/workflow/snakemake_rules/main_workflow.smk
@@ -443,6 +443,7 @@ rule combine_samples:
         augur filter \
             --sequences {input.sequences} \
             --metadata {input.metadata} \
+            --skip-checks \
             --exclude-all \
             --include {input.include} \
             --output-sequences {output.sequences} \
--- a/workflow/snakemake_rules/main_workflow.smk
+++ b/workflow/snakemake_rules/main_workflow.smk
@@ -434,9 +434,9 @@ rule combine_samples:
         sequences = "results/{build_name}/{build_name}_subsampled_sequences.fasta.xz",
         metadata = "results/{build_name}/{build_name}_subsampled_metadata.tsv.xz"
     log:
-        "logs/subsample_regions_{build_name}.txt"
+        "logs/combine_samples_{build_name}.txt"
     benchmark:
-        "benchmarks/subsample_regions_{build_name}.txt"
+        "benchmarks/combine_samples_{build_name}.txt"
     conda: config["conda_environment"]
     shell:
         r"""
--- a/workflow/snakemake_rules/main_workflow.smk
+++ b/workflow/snakemake_rules/main_workflow.smk
@@ -438,6 +438,7 @@ rule combine_samples:
     benchmark:
         "benchmarks/combine_samples_{build_name}.txt"
     conda: config["conda_environment"]
+    threads: 2
     shell:
         r"""
         augur filter \
@@ -446,6 +447,7 @@ rule combine_samples:
             --skip-checks \
             --exclude-all \
             --include {input.include} \
+            --nthreads {threads} \
             --output-sequences {output.sequences} \
             --output-metadata {output.metadata} 2>&1 | tee {log}
         """


workflow/snakemake_rules/main_workflow.smk
```diff
@@ -440,6 +440,7 @@ rule combine_samples:
     benchmark:
         "benchmarks/combine_samples_{build_name}.txt"
     conda: config["conda_environment"]
+    threads: 2
```
joverlee521, line 443: > This change is optimized for Nextstrain's scheduled open and gisaid runs.

Unclear if this might slow down other people's workflow runs. If we only want to optimize this for the Nextstrain automated runs, we can set the job threads per profile (e.g. [custom thread for tree](https://github.com/nextstrain/ncov/blob/137ffc479a1bfe4e2c6ae1bbcd1b22e3afc5f8dd/nextstrain_profiles/nextstrain-gisaid/config.yaml#L10))

workflow/snakemake_rules/main_workflow.smk
```diff
@@ -445,6 +445,7 @@ rule combine_samples:
         augur filter \
             --sequences {input.sequences} \
             --metadata {input.metadata} \
+            --skip-checks \
```
joverlee521, line 449: > (1) is not useful since sanitize_sequences.py already drops duplicates.

Ah, took me a while to realize that the sequences _always_ run through `sanitize_sequences.py` in the `build_align` rule _after_ this filter step. 

Aside, small nit to update `combine_samples` rule's message to reflect this change.

joverlee521 (approved):
Thanks for tracking down the slowness and documenting all of the test runs!

workflow/snakemake_rules/main_workflow.smk
victorlin: Good idea, done in d53568d3b634b036410518e6e32ba0e3f6894bfe.

workflow/snakemake_rules/main_workflow.smk
victorlin: > Ah, took me a while to realize that the sequences _always_ run through `sanitize_sequences.py` in the `build_align` rule _after_ this filter step.

Oh whoops, I only added an explanation for one of the code paths (where `skip_sanitize_metadata` is false). In practice we have it set to true everywhere, and in that case ncov-ingest already drops duplicates ([e.g.](https://github.com/nextstrain/ncov-ingest/blob/5a41e381db0522e3da6bf97406bc9b02a608783f/bin/transform-rki-data-to-ndjson#L26)). Updated the commit message.

> small nit to update `combine_samples` rule's message to reflect this change.

I've removed the rule message in 01d378a619dbab67fe831d7ae6e9f63d801917d3.

joverlee521: > (1) is not useful since inputs are already deduplicated prior to running
this rule (either by ncov-ingest or sanitize_sequences.py).

Hmm, now I'm wondering if this is true. This rule is combining the results of multiple subsampling results, so even if the inputs are deduplicated, a single record could have been selected in two different subsample groups and need to be deduplicated.

victorlin: `--skip-checks` applies on the [input directly](https://github.com/nextstrain/augur/blob/c5c1af171c15f1994d7e6c28aed7092d3b5262b5/augur/filter/_run.py#L82-L84), before `--include`. If the same id appears in multiple `--include` files, it will only appear once in the output because the force-included ids are [stored as a set](https://github.com/nextstrain/augur/blob/c5c1af171c15f1994d7e6c28aed7092d3b5262b5/augur/filter/include_exclude_rules.py#L813) which automatically deduplicates.

I know it's confusing, but let me know if it's still not clear! Good to make sure that this change does not introduce any new cracks.

joverlee521: Ooh, okay that makes sense now. 

Bear with me for one more edge case: could this cause errors if a user starts with an [aligned sequence input](https://docs.nextstrain.org/projects/ncov/en/latest/reference/workflow-config-file.html#aligned) that is not produced by us that includes duplicates? `_get_unified_alignment` will directly use an `aligned` sequence input if [there is only one input](https://github.com/nextstrain/ncov/blob/137ffc479a1bfe4e2c6ae1bbcd1b22e3afc5f8dd/workflow/snakemake_rules/common.smk#L167-L168).

victorlin:
I've started a new [ncov/gisaid trial run](https://github.com/nextstrain/ncov/actions/runs/16483258260). If all goes well, I'll plan to merge tomorrow.

PR #1195, opened 2026-01-28
Update read_metadata usage

victorlin:

This is a small change to take advantage of a recent Augur update.

## Release checklist

 - [x] Replace TBD in changelog

### DIFF
--- a/scripts/sanitize_metadata.py
+++ b/scripts/sanitize_metadata.py
@@ -187,6 +187,7 @@ def get_database_ids_by_strain(metadata_file, metadata_id_columns, database_id_c
             metadata_file,
             id_columns=metadata_id_columns,
             chunk_size=metadata_chunk_size,
+            keep_id_as_column=True,
         )
     except Exception as error:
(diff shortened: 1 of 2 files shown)


scripts/sanitize_metadata.py
joverlee521, line 1: _non-blocking tangent_

This made me wonder how `read_metadata` is used in other scripts. I briefly considered if `combine_metadata.py` should be using `keep_id_as_column` as well since it's adding a `strain` column

https://github.com/nextstrain/ncov/blob/268dda8d33a380ac648252812a03c9ac486bd584/scripts/combine_metadata.py#L47-L48

However, I think that `combine_metadata.py` should just be replaced by `augur merge`?

joverlee521 (approved):

scripts/sanitize_metadata.py
victorlin: Yeah, that makes sense, I won't touch `combine_metadata.py` here. Opened a separate issue for swapping with `augur merge`: #1196

PR #1197, opened 2026-03-24
Remove Nextstrain profiles

victorlin:
This PR contains 5 prep commits and 1 main commit with the following message:

This is a comprehensive shift from `--profile` to `--configfile`, standardizing the preference started in "Remove my_profiles/" (2b880706) and already used by the "100k" workflow (which was misleadingly under "nextstrain_profiles").

Workflow-specific options are translated into command line arguments in the GitHub Actions workflow. Default options are retained in a default profile config, a practice used in other pathogen workflows.

- [ ] Post-merge: release v19


trvrb:
I'm supportive of this but could you update the top-level readme with how to run Nextstrain analyses (rather than having everything hidden away in dev docs)? With open data this should be runnable by others as well.

trvrb:
This will also require a version bump to the workflow.

victorlin:
Added a quick start section to README in 8e712409f25ebd869fccab315b3e2654ac4e0cb9.

Changelog entry:

https://github.com/nextstrain/ncov/blob/8e712409f25ebd869fccab315b3e2654ac4e0cb9/docs/src/reference/change_log.md#L8

PR #1207, opened 2026-06-27
Color Pango lineages corresponding to clade coloring

trvrb:
_Via Claude_

## Motivation

Since ~2025 the `Nextclade_pango` / `pango_lineage` colorings have been a swath of gray, e.g. [nextstrain.org/ncov/open/global/6m@2026-06-26?c=Nextclade_pango](https://nextstrain.org/ncov/open/global/6m@2026-06-26?c=Nextclade_pango). The lineage ordering in `defaults/color_ordering.tsv` was a hand-maintained **alphabetical list of 4,513 lineages** that stopped being updated, so recent lineages got no color and the 999-color palette was consumed by old `A.*` lineages.

By contrast, the `clade_membership` coloring stays current because `assign-colors.py` colors only clades with circulating members in a recency window. This PR does the analogous thing for lineages, live in the build.

## Approach

Generate the lineage coloring **live during the build**, in **full correspondence** with the clade coloring:

- **Correspondence.** A tip is colored via `Nextclade_pango` iff its clade is colored via `clade_membership`. The colored-clade set is computed with the *exact same* recency logic as the clade coloring (clades with a sequence within `--clade-recency`), so the two always agree — no separate list to maintain, and old/extinct clades stay gray on both.
- **Structure.** Lineages are grouped by clade (in `clade_membership` order) and, within a clade, ordered by the **Nextclade reference tree's pre-order**. This keeps sublineages — including Pango aliases (e.g. `XFG.3.4.1` and its alias `QF.2`) — adjacent, so with the gradient palette each clade becomes a contiguous color band and lineage colors shade by clade.
- **No new dependency.** The reference tree comes from the Nextclade dataset zip the build already downloads (`prepare_nextclade`). The build-time present-lineage filter keeps the colored count well under the palette (e.g. ~467 lineages for a 6m build, vs the old 4,513).

Screenshot 2026-06-26 at 8 10 00 PM

### Changes

- **`scripts/color_utils.py`** (new) — shared recency/scheme helpers so the clade and lineage scripts compute the colored-clade set identically.
- **`scripts/assign-colors.py` → `scripts/assign-clade-colors.py`** — renamed to match the new naming, refactored onto `color_utils` (behavior unchanged).
- **`scripts/assign-lineage-colors.py`** (new) — builds the clade-structured, correspondence-filtered `Nextclade_pango` / `pango_lineage` color rows.
- **`defaults/color_ordering.tsv`** — drop the ~9,000 stale `Nextclade_pango` / `pango_lineage` rows (now generated live).
- **`rule colors` → `rule clade_colors` + `rule lineage_colors`** — split so each runs a single script; final output is still `colors.tsv`.

## Test plan

Trial build: https://github.com/nextstrain/ncov/actions/runs/28276144063
- https://nextstrain.org/staging/ncov/open/trial/lineage-colors/north-america/all-time?c=Nextclade_pango
- https://nextstrain.org/staging/ncov/open/trial/lineage-colors/asia/6m?c=Nextclade_pango
- https://nextstrain.org/staging/ncov/open/trial/lineage-colors/global/2m?c=Nextclade_pango

- [x] On the trial build, `?c=Nextclade_pango` is no longer gray and lineages shade by clade.
- [x] Correspondence holds: only lineages of colored clades are colored (clades among colored lineages == colored `clade_membership` set).
- [x] `clade_recency: all` vs `6M` behave (more vs fewer colored lineages, still corresponding).


corneliusroemer:
For Pango lineages this makes a lot of sense, for clades I'd probably also keep hand-maintaining it, there aren't that many new ones, it's useful to have color ordering so one doesn't need a huge script in e.g. Nextclade workflow.

trvrb:
The color ordering of clades is hand maintained at https://github.com/nextstrain/ncov/blob/master/defaults/color_ordering.tsv#L34010, but the actual coloring (of colors for circulating clades and gray for no longer circulating clades) has been automated through `assign-colors.py` for almost two years now (since https://github.com/nextstrain/ncov/pull/1132). With this PR in place, you'd still need to manually add a new clade to `color_ordering.tsv` just as before. I agree this is fine.

trvrb:
Merging after 10 day comment period

PR #1208, opened 2026-06-27
Remove confirmed-dead scripts, defaults, and config keys

trvrb:
First in a series of small, independently-reviewable PRs cleaning up pandemic-era cruft (see the cleanup plan; scope: dead code + proximity removal + docs/schemes).

## Motivation

The repo accumulated scripts, default data files, and config keys that nothing references anymore. This PR removes the verified-dead ones. The three live contracts — weekly OPEN build, occasional GISAID builds, and external users running their own builds — are unaffected.

## What's removed

- **Orphan scripts** (0 refs in `workflow/ Snakefile nextstrain_profiles/ docs/ tests/`): `scripts/add_labels.py`, `scripts/generate-scientific-credits.py`, and the explicitly-deprecated `scripts/deprecated/` (`calculate_delta_frequency.py`, `parse_mutational_fitness_tsv_into_distance_map.py`).
- **Unused defaults**: `defaults/distance_maps/VoC.json` (only `S1.json` is used by the `distances` rule), `defaults/clade_hierarchy.tsv`, `defaults/clades_who.tsv`.
- **Dead config key** `files.outgroup` in `defaults/parameters.yaml` — pointed at a file that doesn't exist in the repo and was read nowhere; its config-reference entry (already documented "No longer used") is removed too.
- **Deprecated `my_profiles/`** directory (only a deprecation README) and its now-orphan `.gitignore` exception.
- **Unused committed example data**: `data/example_*_worldwide.*`, `data/example_*_aus.*`, `data/example_multiple_inputs.tar.xz`. The CI-used `example_metadata.tsv` / `example_sequences.fasta.gz` are kept.

## Verification

- Every removed path was confirmed to have zero references across `workflow/`, `Snakefile`, `nextstrain_profiles/`, `docs/`, and `tests/` (accounting for config-key indirection, which hides references behind `config["files"][...]`).
- `snakemake --profile nextstrain_profiles/nextstrain-ci -n` builds an unchanged 37-job DAG with no errors.

## Test plan

- [ ] CI green.


defaults/clade_hierarchy.tsv
joverlee521, line 1: This file was originally added a human/machine readable file for reference in https://github.com/nextstrain/ncov/pull/817, but was never intended to be used in the workflow. 

Would it be worth keeping as a reference? It's currently used in [ncov-clades-schema](https://github.com/nextstrain/ncov-clades-schema/blob/a01427056f6492d96783913ecd414fdb259691ea/src/generate_clades_json.py#L10), but that repo also has not been run/updated in a while.

data/example_metadata_aus.tsv.xz
joverlee521, line 1: Noting the example data is still referenced in the [ncov-tutorial](https://github.com/nextstrain/ncov-tutorial/blob/059c072a47adaa1e08b57a5f31f26bb8cc33ee55/examples/example_global_context/builds.yaml#L7-L13), but that folder also has a warning that it is just a [public archive and not guaranteed to work. ](https://github.com/nextstrain/ncov-tutorial/blob/059c072a47adaa1e08b57a5f31f26bb8cc33ee55/examples/README.md?plain=1#L2)

Should be okay to delete 👍

scripts/add_labels.py
joverlee521, line 1: Originally added in https://github.com/nextstrain/ncov/commit/9c4877806cc0ac8f45d99646b668eabeb0dd98df as an orphan script. A copy of it is referenced in a [fork](https://github.com/neherlab/ncov-swiss/blob/57588570c234d193c61dc7038f0efcc2ca196f66/swiss_profile/localrules.smk#L7-L26) (from https://github.com/nextstrain/auspice/issues/1036#issuecomment-700592017), but should be good to delete from this repo.

joverlee521 (approved):
Checked a couple of the files that I've seen referenced elsewhere, but should all be good to remove from this repo.

defaults/clade_hierarchy.tsv
trvrb: Oh! Good catch. I thought this was stale, but we need a source of truth for which clades are parents to which clades. It's this file. I've restored it in a rebased cf490125aa9e7d4713e54a6615ac24aba6aa5c60.

scripts/add_labels.py
trvrb: Thanks for investigating!

data/example_metadata_aus.tsv.xz
trvrb: Great!

PR #1213, opened 2026-06-27
Remove stale emerging_lineage, immune_escape, and ace2_binding colorings

trvrb:
Removes three Auspice colorings that are no longer useful.

## Motivation

- **`emerging_lineage`** — the curated `defaults/emerging_lineages.tsv` list hasn't been maintained and won't be revived.
- **`immune_escape` / `ace2_binding`** — computed from deep-mutational-scanning data against **BA.2**, so the "vs BA.2" colorings are inaccurate against today's circulating strains.

## What's removed

**`emerging_lineage`** had real build machinery, all removed:
- `rule emerging_lineages` (an `augur clades` run) + its node-data wiring in `main_workflow.smk`
- the `files.emerging_lineages` config key in `defaults/parameters.yaml`
- `defaults/emerging_lineages.tsv` (now orphaned)
- the coloring + filter in both `defaults/auspice_config.json` and the dynamic `auspice_config` rule in `export_for_nextstrain.smk`
- the 18 `emerging_lineage` rows in `defaults/color_ordering.tsv`
- the config-reference docs section

**`immune_escape` / `ace2_binding`** are not produced by any rule here (the default `sars-cov-2` Nextclade dataset doesn't emit them; they came from the 21L DMS dataset / upstream metadata), so removal is **display-only**:
- the two coloring blocks in both auspice configs
- their now-inert passthrough rename mappings in `scripts/join-metadata-and-clades.py`

## Verification

- `grep` for `emerging_lineage|immune_escape|ace2_binding` across the repo (excl. historical changelog) returns **zero** hits.
- `defaults/auspice_config.json` remains valid JSON (23 colorings / 9 filters, down from 26 / 10).
- `snakemake --profile nextstrain_profiles/nextstrain-ci -n` resolves with **36 jobs, down from 37** (only the `emerging_lineages` rule), no errors.

## Test plan

- [ ] CI green.
- [ ] Trial build: confirm exported Auspice JSON `colorings`/`filters` no longer contain the three keys.


docs/src/reference/change_log.md
```diff
@@ -9,6 +9,7 @@ We also use this change log to document new features that maintain backward comp
 - 29 July 2025: Improved performance of calls to `augur filter`. This requires a minimum Augur version of 31.3.0. [PR 1178](https://github.com/nextstrain/ncov/pull/1178)
 - 28 January 2026: Large JSON files are now automatically minified. This requires a minimum Augur version of 33.0.0. [PR 1194](https://github.com/nextstrain/ncov/pull/1194)
 - 30 January 2026: Simplified code for reading metadata. This requires a minimum Augur version of 32.1.0. [PR 1195](https://github.com/nextstrain/ncov/pull/1195)
+- 26 June 2026: Remove the stale `emerging_lineage`, `immune_escape`, and `ace2_binding` colorings. The latter two were computed from BA.2 deep-mutational-scanning data and are no longer accurate against current strains. [PR 1213](https://github.com/nextstrain/ncov/pull/1213)
```
joverlee521: I would also flag that the `files.emerging_lineages` config param has been removed. 

This is technically a backwards incompatible change that warrants a new release. Perhaps batch all of the clean up PRs into a singe v18 release.

joverlee521 (approved):
Seems reasonable to remove the outdated colorings, this opens the door to cleaning up ncov-ingest (https://github.com/nextstrain/ncov-ingest/issues/504) as well.

docs/src/reference/change_log.md
trvrb: Yeah, batching as v18 makes sense. We can merge these to `master` right and then tag `master` as `v18`?

trvrb: I added 
>along with the now-unused `files.emerging_lineages` config parameter

to the changelog

joverlee521: yes, merge and then create a new release as per checklist

https://github.com/nextstrain/ncov/blob/f9c1522f6a8db551615bbad70a6fa117fd66a96c/.github/pull_request_template.md?plain=1#L16-L22

PR #1215, opened 2026-07-01
Remove the defunct scripts/curate_metadata toolkit

trvrb:
_Via Claude_

## Summary

Deletes `scripts/curate_metadata/` — a standalone, manually-run toolkit from the 2021–2022 GISAID-download era that is not part of the Snakemake build and has been dormant for ~3–4 years.

## What it did

A human ran these scripts by hand (edit-and-run; no argparse) to:
- **Audit dataset geography** (`curate_metadata.py`) — flag duplicate/misspelled place names, missing `lat_longs.tsv` coordinates (geocoded via OpenStreetMap), and division/location clashes — and emit corrected `geoLocationRules` + `*_annotations.tsv` that were **manually copy-pasted into `ncov-ingest/source-data/`**.
- **Parse GISAID free-text fields** (`parse_additional_info.py`) into geography / `sampling_strategy` annotations for the same manual round-trip.
- **QC new-sequence drops and draft submitter "thank you" tweets** (`parse_new_sequences.py`, `tweet_generation.py`) — these only produced human-facing text, never posted or fed data anywhere.

## Why it can go

- **Not in the build / unreferenced.** Nothing in the repo imports or runs it — no Snakemake rule, no CI, no docs. The only in-repo mention was self-referential `.gitignore` lines (removed here).
- **Not a source of truth.** The live annotation/geo-rule data lives in and is regenerated by `ncov-ingest/source-data/` (`gisaid_annotations.tsv`, `genbank_annotations.tsv`, `gisaid_geoLocationRules.tsv`), read by ncov-ingest's own `transform-*` scripts from ncov-ingest's repo. This toolkit only ever *generated candidates* a human hand-applied there; ncov-ingest does not reference it. The `config_*` dirs here are unique working-set heuristics, mirrored nowhere.
- **Defunct.** No commits since 2021–2022; the clade tables stop at `21H` (pre-Omicron), it assumes the old `downloaded_gisaid.tsv` layout and metadata schema, and its required inputs no longer exist. The manual curation and tweet workflows it supported were discontinued years ago.
- **Recoverable.** Fully preserved in git history if the manual process is ever revived.

Removes ~150 KB / 30k lines of dead code (including the pre-Omicron `clock_deviation` references) and one more "is this still used?" trap.

_Reviewer sanity check: no consumer found in `ncov` or `ncov-ingest`. If any other Nextstrain repo or ops process still references these scripts, flag it and I'll hold._

## Test plan
[...]

PR #1216, opened 2026-07-01
docs: harden changelog links to the remote inputs page (fix flaky Docs CI)

trvrb:
_Via Claude_

## Summary

The **Docs** job intermittently fails with:
```
WARNING: Error, no command xref target from reference/change_log:remote_inputs
```
which — because Sphinx warnings are treated as errors — turns red on PRs that don't touch docs at all (e.g. #1215, whose docs source is byte-identical to a green master build on the same commit).

The cause is two `[…](remote_inputs)` markdown links in `docs/src/reference/change_log.md` (lines 149, 160) that rely on **recommonmark's relative auto-xref resolution**. That resolution is order-dependent and resolves inconsistently between builds, so the same source flakes red or green.

## Fix

Replace both with the full published URL — `https://docs.nextstrain.org/projects/ncov/en/latest/reference/remote_inputs.html` — which is exactly the convention **every other cross-doc link in this changelog already uses** (e.g. the `gisaid-search.html` link on line 147 and the `workflow-config-file.html` link further down). External URLs aren't subject to Sphinx xref resolution, so the intermittent failure can't recur for these links.

Both are historical (2021) entries, so a stable absolute URL is appropriate.

## Test plan
- [x] No remaining bare `](remote_inputs)` auto-xrefs anywhere in `docs/`
- [x] Docs CI green (and no longer flakes on unrelated PRs)

### DIFF
--- a/docs/src/reference/change_log.md
+++ b/docs/src/reference/change_log.md
@@ -146,7 +146,7 @@ We also use this change log to document new features that maintain backward comp
 
  - 11 August 2021: Add support for "Sequences" and "Patient status metadata" downloads from GISAID's search interface including [documentation in the tutorial of how to use these data](https://docs.nextstrain.org/projects/ncov/en/latest/guides/data-prep/gisaid-search.html). ([#701](https://github.com/nextstrain/ncov/pull/701))
  - 6 August 2021: We've replaced the mechanisms that support remote file inputs (e.g. `s3://` URLs) to improve internal workflow structure, extend support to `gs://`, `http://`, and `https://` URLs, and expand support for compressed inputs.
-   Our [remote file inputs documentation](remote_inputs) is updated to reflect the changes.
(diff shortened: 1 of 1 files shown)


victorlin (approved):
Thanks for investigating. I'm also seeing this [elsewhere](https://github.com/nextstrain/augur/actions/runs/28540573016/job/84613037956) today and it hasn't come up in the past, so don't think it's just 'flaky'. Might be due to a new release of some dependency.

trvrb:
Oh, interesting. I don't have a real sense of the underlying cause.

victorlin:
"Error, no command xref target …" is from sphinx-argparse ([src](https://github.com/sphinx-doc/sphinx-argparse/blob/f918d4ebde3cef185ddd362de2967126716842be/sphinxarg/ext.py#L936-L937)) which just released [v0.6.0](https://github.com/sphinx-doc/sphinx-argparse/releases/tag/v0.6.0) yesterday. I haven't looked into specifically what changed, but that's probably the underlying cause.

PR #1217, opened 2026-07-13
linkcheck: ignore cdc.gov

joverlee521:

Docs CI started failing with 403
<https://github.com/nextstrain/ncov/actions/runs/29273949292/job/86898229781>

### DIFF
--- a/docs/src/conf.py
+++ b/docs/src/conf.py
@@ -124,6 +124,7 @@ linkcheck_ignore = [
     # the top level bucket link 404s
     r'^https://data\.nextstrain\.org$',
     # they block the client, probably anti-scraping measure
+    r'^https://www\.cdc\.gov',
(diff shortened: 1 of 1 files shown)


joverlee521:
OH right, recently changed workflow so that [linkcheck only runs on scheduled runs](https://github.com/nextstrain/ncov/blob/master/.github/workflows/docs.yaml#L23).

Verified locally that the CDC links were ignored as expected
```
(reference/change_log: line  234) -ignored- https://www.cdc.gov/covid/php/variants/index.html
...
(tutorial/next-steps: line   76) -ignored- https://www.cdc.gov/advanced-molecular-detection/php/training/?CDC_AAref_Val=https://www.cdc.gov/amd/training/covid-19-gen-epi-toolkit.html
```

PR #1219, opened 2026-07-21
Only apply global include file when combining samples

victorlin:

Previously, applying it to every subsampling set (intermediate augur filter call) guaranteed that there would always be at least one sequence in the output. This was prone to hiding issues with the per-sample filters.

`--empty-output-reporting warn` prevents augur filter from otherwise erroring on empty subsampling sets, which would be a breaking change. Empty subsampling sets now show a warning message in the command output.

## Related issue(s)

Noticed while working on https://github.com/nextstrain/augur/issues/2032

## Testing

Ran locally to confirm that workflow shows the following message after fixing the Augur bug.

> WARNING: All samples have been dropped! Check filter rules and metadata file format.

The empty subsampling set still runs and has no entry in the `combine_samples` log.

```
9428989 strains were dropped during filtering
	9430425 were dropped by `--exclude-all`
	564 were added back because they were in results/africa_6m/sample-region_early.txt
	176 were added back because they were in results/africa_6m/sample-context_early.txt
	695 were added back because they were in results/africa_6m/sample-context_recent.txt
	1 was added back because it was in defaults/include.txt
1436 strains passed all filters
```

## Release checklist

### DIFF
--- a/docs/src/reference/change_log.md
+++ b/docs/src/reference/change_log.md
@@ -5,6 +5,8 @@ We also use this change log to document new features that maintain backward comp
 
 ## New features since last version update
 
+- 22 July 2026: The workflow now warns when a subsampling set is configured with filters that drop all sequences.
+
 ## v18 (6 July 2026)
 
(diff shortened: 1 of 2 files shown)


joverlee521 (approved):
Nice, it will be good to see in the logs which region samples are empty and we can determine if we should just remove them.

